PR #1358, opened 2015-01-07
[Sprint: 41] XD-2416: Handle escaped quotes according to outer quotes

ericbottard:
Had a discussion with @aclement about this. Finally decided not to handle escapes at the Tokenizer level because
1. less regression risks
2. one could even argue if we do that, then the token payload should not contain the outer quotes IMO. So even more regression risks.


dturanski:
LGTM; merged

PR #1359, opened 2015-01-07
[Sprint: 41] XD-2477 Adding stoppable logic to AbstractProcessBuilderTasklet

trisberg:
- improving log management using temp files

- improving exit status handling

- improving step execution status management when job fails

- still can't make Sqoop job stoppable due to the fact that the Sqoop mapreduce job keeps running even if launching process is destroyed

- adding INFO logging for Sqoop tasklet so user can tail Sqoop log for long running jobs

### DIFF
--- a/extensions/spring-xd-extension-batch/src/main/java/org/springframework/batch/step/tasklet/x/AbstractProcessBuilderTasklet.java
+++ b/extensions/spring-xd-extension-batch/src/main/java/org/springframework/batch/step/tasklet/x/AbstractProcessBuilderTasklet.java
@@ -46,6 +46,8 @@ import java.util.Map;
  * Abstract tasklet for running code in a separate process and capturing the log output. The step execution
  * context will be updated with some runtime information as well as with the log output.
  *
+ * Note: This this class is not thread-safe.
+ *
  * @since 1.1
  * @author Thomas Rrisberg
  */
--- a/extensions/spring-xd-extension-batch/src/main/java/org/springframework/batch/step/tasklet/x/AbstractProcessBuilderTasklet.java
+++ b/extensions/spring-xd-extension-batch/src/main/java/org/springframework/batch/step/tasklet/x/AbstractProcessBuilderTasklet.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2014 the original author or authors.
+ * Copyright 2014-2015 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -19,9 +19,11 @@ package org.springframework.batch.step.tasklet.x;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 import org.springframework.batch.core.ExitStatus;
+import org.springframework.batch.core.JobExecution;
 import org.springframework.batch.core.StepContribution;
 import org.springframework.batch.core.StepExecution;
 import org.springframework.batch.core.StepExecutionListener;
+import org.springframework.batch.core.explore.JobExplorer;
 import org.springframework.batch.core.scope.context.ChunkContext;
 import org.springframework.batch.core.step.tasklet.SimpleSystemProcessExitCodeMapper;
 import org.springframework.batch.core.step.tasklet.SystemProcessExitCodeMapper;
@@ -29,8 +31,10 @@ import org.springframework.batch.core.step.tasklet.Tasklet;
 import org.springframework.batch.repeat.RepeatStatus;
 
 import java.io.BufferedReader;
+import java.io.File;
+import java.io.FileInputStream;
+import java.io.FileNotFoundException;
 import java.io.IOException;
-import java.io.InputStream;
 import java.io.InputStreamReader;
 import java.net.URL;
 import java.net.URLClassLoader;
@@ -54,6 +58,18 @@ public abstract class AbstractProcessBuilderTasklet implements Tasklet, StepExec
 	 */
 	protected int exitCode = -1;
 
+	private String exitMessage;
+
+	private boolean complete = false;
+
+	private boolean stopped = false;
+
+	private boolean stoppable = false;
+
+	private long checkInterval = 1000;
+
+	private JobExplorer jobExplorer;
+
 	private SystemProcessExitCodeMapper systemProcessExitCodeMapper = new SimpleSystemProcessExitCodeMapper();
 
 
@@ -61,7 +77,6 @@ public abstract class AbstractProcessBuilderTasklet implements Tasklet, StepExec
 	public RepeatStatus execute(StepContribution contribution, ChunkContext chunkContext) throws Exception {
 
 		StepExecution stepExecution = chunkContext.getStepContext().getStepExecution();
-		ExitStatus exitStatus = stepExecution.getExitStatus();
 
 		String commandDescription = getCommandDescription();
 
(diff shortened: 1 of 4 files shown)


extensions/spring-xd-extension-batch/src/main/java/org/springframework/batch/step/tasklet/x/AbstractProcessBuilderTasklet.java
```diff
@@ -54,6 +58,18 @@
 	 */
```
mminella: We probably should note in the javadoc that this class is **not** threadsafe

extensions/spring-xd-extension-batch/src/main/java/org/springframework/batch/step/tasklet/x/AbstractProcessBuilderTasklet.java
```diff
@@ -73,25 +88,71 @@ public RepeatStatus execute(StepContribution contribution, ChunkContext chunkCon
 
 		String commandClassPath = createClassPath(this.getClass());
 
-		ProcessBuilder pb = new ProcessBuilder(command).redirectErrorStream(true);
+		File out = File.createTempFile(commandName + "-", ".out");
+		stepExecution.getExecutionContext().putString(commandName.toLowerCase() + ".system.out", out.getAbsolutePath());
+		logger.info(commandName + " system.out: " + out.getAbsolutePath());
+		File err = File.createTempFile(commandName + "-", ".err");
```
mminella: While temp files are definitely a step in the right direction, is there a reason we can't inject a location for these instead of using temp files?

extensions/spring-xd-extension-batch/src/main/java/org/springframework/batch/step/tasklet/x/AbstractProcessBuilderTasklet.java
```diff
@@ -73,25 +88,71 @@ public RepeatStatus execute(StepContribution contribution, ChunkContext chunkCon
...
+					if (stoppable) {
+						JobExecution jobExecution =
+								jobExplorer.getJobExecution(stepExecution.getJobExecutionId());
+
+						if (jobExecution.isStopping()) {
+							stopped = true;
+						} else if (chunkContext.getStepContext().getStepExecution().isTerminateOnly()) {
+							stopped = true;
+						}
+					}
```
mminella: In the future, a timeout capability may also be useful (`SystemCommandTasklet` has one)

extensions/spring-xd-extension-batch/src/main/java/org/springframework/batch/step/tasklet/x/AbstractProcessBuilderTasklet.java
```diff
@@ -102,34 +163,37 @@ public RepeatStatus execute(StepContribution contribution, ChunkContext chunkCon
...
-						else {
-							break;
-						}
-					}
+			if (p != null) {
+				p.destroy();
+			}
+			commandOut = getProcessOutput(out);
+			commandErr = getProcessOutput(err);
+			printLog(commandName, commandOut, commandErr);
```
mminella: So above we write the output to temp files.  Then here, we load those files back into memory only to pipe them back out to the real log?  We may want to approach this in a way that doesn't require us to load the entire standard out and standard error from the process to be loaded into memory.

extensions/spring-xd-extension-batch/src/main/java/org/springframework/batch/step/tasklet/x/AbstractProcessBuilderTasklet.java
```diff
@@ -138,13 +202,29 @@ public RepeatStatus execute(StepContribution contribution, ChunkContext chunkCon
 
 	@Override
 	public void beforeStep(StepExecution stepExecution) {
+		exitCode = -1;
```
mminella: Curious if we want to default to -1 (error condition) or 0 (successful).

extensions/spring-xd-extension-batch/src/main/java/org/springframework/batch/step/tasklet/x/AbstractProcessBuilderTasklet.java
```diff
@@ -224,21 +311,41 @@ protected String createClassPath(Class taskletClass) {
 		return lines;
 	}
 
-	protected void printLog(String commandName, List<String> lines, int exitCode) {
-		if (exitCode != 0) {
-			for (String line : lines) {
-				logger.error(commandName + " log: " + line);
+	protected void printLog(String commandName, List<String> out, List<String> err) {
```
mminella: Going back to my previous comment about loading the files into memory, you could change this to take the `File` for out and err instead of reading it into a list first.

extensions/spring-xd-extension-batch/src/main/java/org/springframework/batch/step/tasklet/x/AbstractProcessBuilderTasklet.java
trisberg: Agree

extensions/spring-xd-extension-batch/src/main/java/org/springframework/batch/step/tasklet/x/AbstractProcessBuilderTasklet.java
trisberg: We could consider this as an enhancement down the road, how would this be specified though? We could name them like xd-sqoop-job1-step0.out and put them in XD's log directory.

extensions/spring-xd-extension-batch/src/main/java/org/springframework/batch/step/tasklet/x/AbstractProcessBuilderTasklet.java
trisberg: Yes, for the Soop job that's not useful though, but for other process it might be.

extensions/spring-xd-extension-batch/src/main/java/org/springframework/batch/step/tasklet/x/AbstractProcessBuilderTasklet.java
trisberg: We load 10000 first lines and the main purpose is to store that in the step execution. We also use a selection of these for debug logging and error logging in the XD container log. if there is a better way, I'm fine with that.

extensions/spring-xd-extension-batch/src/main/java/org/springframework/batch/step/tasklet/x/AbstractProcessBuilderTasklet.java
trisberg: I'm pessimistic so -1 seems better until we have success reported from the process

extensions/spring-xd-extension-batch/src/main/java/org/springframework/batch/step/tasklet/x/AbstractProcessBuilderTasklet.java
trisberg: We could, I already have the lists though

extensions/spring-xd-extension-batch/src/main/java/org/springframework/batch/step/tasklet/x/AbstractProcessBuilderTasklet.java
markpollack: might need to have a directory made per container process or put the pid into the log file name.

markpollack: ah, forget that.  I see it is File.createTempFile.  OK to create another JIRA to make this customized, putting into XD log directory seems pretty reasonable.

extensions/spring-xd-extension-batch/src/main/java/org/springframework/batch/step/tasklet/x/AbstractProcessBuilderTasklet.java
markpollack: Is it too crazy to store all the lines in the step execution?  It is the one central place that we do have for storing state.

mminella: @ghillert and I were talking about this in a different capacity.  The `ExecutionContext` is intended for holding state of a job.  It's not intended to be holding entire log files which I'm sure we all agree can grow to be quite large.  This really goes to how logs in general should be aggregated in Spring XD.  Currently this is attempting to address one use case of log management, but there are many others that XD is starting to bump up against.  The use case I was talking with Gunnar about was that users want to see the full stack trace of an error that occurs in a batch job in the UI, however we don't persist it anywhere (we store the first 2500 characters of it in the job repository in most  situations).  I think we need to talk about a more general log aggregation strategy for Spring XD (that may just be punting to PCF) in general.

trisberg: For now I think we need to limit the number of lines not to blow out memory during log processing, or does the Batch API have a way of streaming text to the step context? It does end up in a LOB eventually.

mminella: Spring Batch uses XStream for serialization of the {{ExecutionContext}}.  It's just basic serialization so whatever you want to store in the job repository's EC needs to fit in memory.

mminella:
Merged

PR #1360, opened 2015-01-07
XD-2567: Rabbit Bus - Strip DeliveryMode Header

garyrussell:
JIRA: https://jira.spring.io/browse/XD-2567

Now that the sink is in a different classloader, the
MessagePropertiesConverter cannot cast the deliverymode header.

In any case, the header should not be populated by the bus, because
it would override the `default-delivery-mode` setting in the sink.

Also tested with `time | rabbit` to verify the `WARN` log no longer appears.

Full build done locally.


markfisher:
LGTM; merging

markfisher:
merged as https://github.com/spring-projects/spring-xd/commit/1d935c7df06ef0a066ca2b07107876eaf5344c88

PR #1365, opened 2015-01-08
[Sprint: 41] XD-2562 Fixing dependency issues for running tests in STS

trisberg:


gradle/build-hadoop-distros.gradle
```diff
@@ -28,6 +28,7 @@ project('spring-xd-hadoop:hadoop24') {
 		runtime ("org.apache.hadoop:hadoop-mapreduce-client-jobclient:$hadoop24Version")
 		runtime ("org.apache.hadoop:hadoop-streaming:$hadoop24Version")
 		runtime ("org.apache.hadoop:hadoop-yarn-common:$hadoop24Version")
+		configurations.runtime.exclude(group: 'org.xerial.snappy')
```
dturanski: These will not get picked up by the maven plugin. If they need to be excluded from the pom need to specify the modules explicitly

trisberg: These aren't "real" projects so not too worried about the pom, really would be better to not generate a pom at all.

trisberg: The only reason we need these excludes are for our tests to run in an IDE, so for the pom files we really don't need theese excludes

dturanski: I added the specific modules to excludes to fix spring-xd-dirt and spring-xd-test poms which are critical to spring-xd-module-parent  (also spring-xd-module and spring-xd-module-spi). As long as we can still run SingleNodeApplication from an external maven project. Unfortunately there are no automated tests for this.

build.gradle
```diff
@@ -996,27 +996,23 @@ project('spring-xd-test') {
 		compile "org.springframework.integration:spring-integration-amqp"
 		compile "org.springframework.integration:spring-integration-redis"
 		compile "org.springframework.integration:spring-integration-mqtt"
-		compile("org.springframework.data:spring-data-hadoop") {
-			exclude group: 'org.springframework.batch', module: 'spring-batch-core'
```
dturanski: You changed a lot of stuff here. Are the original exclusions preserved?

dturanski:
I ran ./gradlew install to build the spring-xd-module-parent pom and then ran the build for a couple of module projects. LGTM.

build.gradle
trisberg: I think the old excludes were an outdated list - matched the new excludes with what is in the spring-xd-hadoop module minus a few that we need to get the minicluster running for the tests

PR #1368, opened 2015-01-09
[Sprint 41] XD 2410 & 2412: fix FieldValueCounter names and behavior

ericbottard:


spring-xd-analytics/src/main/java/org/springframework/xd/analytics/metrics/redis/RedisFieldValueCounterRepository.java
```diff
@@ -54,20 +54,16 @@ public RedisFieldValueCounterRepository(RedisConnectionFactory connectionFactory
...
-					increment(fieldValueCounter.getName(), entry.getKey(), entry.getValue());
-				}
-			}
-			else {
-				increment(fieldValueCounter.getName(), MARKER, 0);
-			}
+		delete(fieldValueCounter.getName());
+		for (Map.Entry<String, Double> entry : fieldValueCounter.getFieldValueCount().entrySet()) {
+			increment(fieldValueCounter.getName(), entry.getKey(), entry.getValue());
 		}
```
markpollack: The test in `RedisFieldValueCounterRepositoryTests` fails now.

----
	public void testCrud() {
		FieldValueCounter counter = new FieldValueCounter(tickersFieldValueCounterName);

		// Test with initial save
		FieldValueCounter fvTickersCounter = fieldValueCounterRepository.save(counter);
		assertThat(fieldValueCounterRepository.findOne(fvTickersCounter.getName()), notNullValue());
----
I think the previous code  was trying to fake saving a 0 value entry in redis to preserve 'i saved an empty FVC' semantics.  This seems to still be needed, I updated the save method to.

----
	public <S extends FieldValueCounter> S save(S fieldValueCounter) {
		delete(fieldValueCounter.getName());
		increment(fieldValueCounter.getName(), MARKER, 0);
		for (Map.Entry<String, Double> entry : fieldValueCounter.getFieldValueCount().entrySet()) {
			increment(fieldValueCounter.getName(), entry.getKey(), entry.getValue());
		}
		return fieldValueCounter;
	}
----

and it saves correctly now.  Will add as part of merge.

markpollack:
rebased, squashed and merged as d7563443d6a127c7e0eafe9de71434da1c14c61d

PR #1370, opened 2015-01-09
[Sprint: 41] XD-2547 accept any file name for module configuration and added some tests

dturanski:


spring-xd-module/src/main/java/org/springframework/xd/module/support/ModuleUtils.java
```diff
@@ -75,24 +73,29 @@ public static Resource modulePropertiesFile(SimpleModuleDefinition definition, C
...
 		String ext = extension.startsWith(".") ? extension : "." + extension;
-		for (String fileName : supportedFileNames) {
-			Resource candidate = new ClassPathResource(String.format("/%s/%s%s", "config", fileName, ext),
-					moduleClassLoader);
-			if (candidate.exists() && candidate.isReadable()) {
-				Assert.isNull(result, String.format("duplicate module definitions found: %s and %s",
-						(result == null ? "" : result.getFilename()), candidate.getFilename()));
-				result = candidate;
+		try {
+			Resource[] resources = resolver.getResources("/config/*" + ext);
```
ericbottard: We need to be wary that using the `moduleClassLoader` as it is currently constructed, may return any classpath resource that is inside a `/config/` dir (_e.g._ if log4j.jar has `/config/foo.properties`, this **will** be returned).

Maybe we need to pass only a "top level module classloader" for that purpose

spring-xd-module/src/main/java/org/springframework/xd/module/support/ModuleUtils.java
```diff
@@ -75,24 +73,29 @@ public static Resource modulePropertiesFile(SimpleModuleDefinition definition, C
 	}
 
 	/**
-	 * Return an expected module resource searching conventional file names. Will throw an exception if more than one
-	 * such resource exists
+	 * Return an expected module resource given a file extension. Will throw an exception if more than one such
+	 * resource
```
ericbottard: weird javadoc formatting (+ missing end dot)

spring-xd-module/src/test/java/org/springframework/xd/module/core/ModuleFactoryTests.java
```diff
@@ -58,6 +58,25 @@ public void createXmlModule() {
...
+	@Test(expected = IllegalStateException.class)
+	public void multipleXmlConfigThrowsException() {
+		createResourceConfiguredModule("invalid",ModuleType.processor);
+	}
+
+	@Test(expected = IllegalArgumentException.class)
+	public void missingConfigThrowsException() {
+		ModuleDefinition moduleDefinition = ModuleDefinitions.simple("missing", ModuleType.processor,
+				"classpath:/ModuleFactoryTests/modules/processor/missing");
+		ModuleDescriptor moduleDescriptor = new ModuleDescriptor.Builder()
```
ericbottard: Just browsing through the web interface at the moment, but I don't see `moduleDescriptor` being used in the test

spring-xd-module/src/main/java/org/springframework/xd/module/core/JavaConfiguredModule.java
```diff
@@ -62,8 +57,10 @@ protected void configureModuleApplicationContext(SimpleModuleDefinition moduleDe
 	}
 
 	public static String[] basePackages(SimpleModuleDefinition moduleDefinition, ClassLoader moduleClassLoader) {
-		Resource propertiesFile = ModuleUtils.locateModuleResource(moduleDefinition, moduleClassLoader,".properties");
-		Assert.notNull(propertiesFile, "required module properties not found.");
+		Resource propertiesFile = ModuleUtils.locateModuleResource(moduleDefinition, moduleClassLoader, ".properties");
+		if (propertiesFile == null) {
+			return new String[0];
```
ericbottard: From the implementation of `scan()` downstream, seems like we'd better fail fast:
```java
	/**
	 * Perform a scan within the specified base packages.
	 * <p>Note that {@link #refresh()} must be called in order for the context
	 * to fully process the new classes.
	 * @param basePackages the packages to check for annotated classes
	 * @see #register(Class...)
	 * @see #refresh()
	 */
	public void scan(String... basePackages) {
		Assert.notEmpty(basePackages, "At least one base package must be specified");
		this.scanner.scan(basePackages);
	}
```

dturanski: The reason for this is ModuleFactoryTests.missingConfigThrowsException(). It's hard to tell from the existing test, since it uses @Test(expected=..) and swallows the exception. The current implementation results in a better message:
java.lang.IllegalArgumentException: Required module artifacts are either missing or invalid. Unable to determine module type for module definition: 'processor:missing'.  Vs. "required module properties not found" from JavaConfiguredModule.basePackages(), which is misleading (what MP recently got prior to this change and found confusing).

dturanski:
It works, but I see what you mean. I'll redo it.

ericbottard:
Tested, rebased and merged as 7c8ff6c42b3d8e988cedf45990212413164ce19a

PR #1372, opened 2015-01-09
[Sprint 41] XD 2542 - Create MessageHandler for RxJava based processor modules

markpollack:
Based off 2d59676ccbe20ff2e86cb5e737ea71b25e973e57

* Changed to use PublishSubject
* Remove check for existing observers
* Throw exception if can't dispatch message to a Subject
* Keep track of Subscriptions, unsubscribe on app context destory
* Remove Subject from map when it completes, will be created again on 
  next message for a given dispatcher thread.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -40,6 +40,7 @@ ext {
 		'spring-xd-analytics',
 		'spring-xd-analytics-ml',
 		'spring-xd-reactor',
+		'spring-xd-rxjava',
 		'spring-xd-tuple',
 		'spring-xd-module',
 		'spring-xd-rest-client',
@@ -344,6 +345,14 @@ project('spring-xd-reactor') {
 	description = 'Spring XD Reactor'
 	dependencies {
 		compile "io.projectreactor:reactor-core"
+		compile "org.springframework.integration:spring-integration-core"
+	}
+}
+
+project('spring-xd-rxjava') {
+	description = 'Spring XD RxJava'
+	dependencies {
+		compile "io.reactivex:rxjava"
 		compile "org.springframework.integration:spring-integration-core"
 	}
 }
--- a/dependencies.properties
+++ b/dependencies.properties
@@ -99,6 +99,7 @@ jaxen\:jaxen=1.1.6
 jline\:jline=2.11
 joda-time\:joda-time=2.5
 junit\:junit=4.12
+io.reactivex\:rxjava=1.0.0
 ldapsdk\:ldapsdk=4.1
 log4j\:log4j=1.2.17
 mysql\:mysql-connector-java=5.1.34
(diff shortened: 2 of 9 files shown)


ericbottard:
After some research, seems like http://reactivex.io/RxJava/javadoc/rx/subjects/SerializedSubject.html is exactly what we should be using, rather than trying our own (possibly broken) solution. What am I missing here? cc @smaldini

markpollack:
No processing occurs in parallel afaik with that arrangement.
M


On Jan 12, 2015, at 9:02 AM, Eric Bottard <notifications@github.com> wrote:

After some research, seems like
http://reactivex.io/RxJava/javadoc/rx/subjects/SerializedSubject.html is
exactly what we should be using, rather than trying our own (possibly
broken) solution. What am I missing here? cc @smaldini
<https://github.com/smaldini>

—
Reply to this email directly or view it on GitHub
<https://github.com/spring-projects/spring-xd/pull/1372#issuecomment-69573367>
.

markpollack:
This is sharing a single subject across multiple threads vs assigning a
subject per thread, so it isn't the same.

At the end of the day, we will probably need to support a variety of
dispatching options.

Ironically, ATM the simple sample for RXJava I committed to
spring-xd-samples would benefit from SerializedSubject vs
publisherSubject...but that is for 'hello world'

i don't think we provide enough control for all use-cases ATM so need to
make some pragmatic compromise.

My gut feel is that additional dispatching options, in combination with
message bus/listener container options, will have to wait until a 1.2.1
release.

On Jan 12, 2015, at 9:02 AM, Eric Bottard <notifications@github.com> wrote:

After some research, seems like
http://reactivex.io/RxJava/javadoc/rx/subjects/SerializedSubject.html is
exactly what we should be using, rather than trying our own (possibly
broken) solution. What am I missing here? cc @smaldini
<https://github.com/smaldini>

—
Reply to this email directly or view it on GitHub
<https://github.com/spring-projects/spring-xd/pull/1372#issuecomment-69573367>
.

markpollack:
Changed implementation to use SerialzedSubject - it is sitting in a seperate branch at 

https://github.com/markpollack/spring-xd/commit/ccf299795761088c1ec77d51fa239b5d14d06fd8

It works nicely with the example in https://github.com/spring-projects/spring-xd-samples/tree/master/rxjava-moving-average

Will experiment with introducing a SubjectFactory with two implementations.  One uses SerializedSubject and the other creates a Subject based on the result of evaluating a SpEL expression on the message.  This way one could have a spel expression that creates a Subject per Kafka Partition, or a subject per threadId, etc.

markpollack:
closing, will be issuing a new PR.

PR #1373, opened 2015-01-12
XD-2564 Enhance XD on YARN to use SHDP container clustering

jvalkeal:
I have done some updates to wiki in my clone: https://github.com/jvalkeal/spring-xd/wiki/Running-on-YARN

- Use container grouping functionality from shdp which allows
  to run different yarn container configurations within
  a same yarn app.
- Remove splitted functionality running admins and containers via
  a separate yarn apps.
- Tune xd-yarn script to work better.
- Change `-p` and `--path` parameters in xd-yarn script to
  `-d` and `--dir` because `-p` is already reserved in some of
  the grouping control commands.
- Admin and container logging which previously got broken is fixed.
  This turned out to be a missing `-Dlogging.config` arguments for both
  admins and containers in application.yml. It looks like boot previously
  enabled some default logging which then got broken when boot was
  updated to version 1.2.x.


trisberg:
LGTM now, getting ready to merge

trisberg:
Squashed and Merged

PR #1374, opened 2015-01-12
[Sprint-41] XD-2411: Use optimistic locking in Redis RichGauge repo

ericbottard:


spring-xd-analytics/src/main/java/org/springframework/xd/analytics/metrics/redis/RedisRichGaugeRepository.java
```diff
@@ -26,14 +35,17 @@
...
+ * @author Eric Bottard
  */
-public final class RedisRichGaugeRepository extends
+public class RedisRichGaugeRepository extends
 		AbstractRedisMetricRepository<RichGauge, String> implements RichGaugeRepository {
 
-	private static final String ZERO = serialize(new RichGauge("zero"));
+	private static final String ZERO = serialize(new RichGauge("ZERO"));
+
+	private RetryTemplate retryTemplate = new RetryTemplate();
```
markpollack: I think we should expose the options for `ExponentialBackOffPolicy` and max number of retries instead as in 

https://github.com/spring-projects/spring-xd/blob/master/spring-xd-messagebus-spi/src/main/java/org/springframework/xd/dirt/integration/bus/MessageBusSupport.java#L751

The defaults in `ExponentialBackOffPolicy` may need to be tweeked.  It probably doesn't make sense to try and write if there is contention for the key more than 5 seconds or so.  The current retry as fast as possible for 3x is probably too aggressive.

spring-xd-analytics/src/main/java/org/springframework/xd/analytics/metrics/redis/RedisRichGaugeRepository.java
```diff
@@ -67,14 +79,39 @@ String value(RichGauge metric) {
...
 	@Override
-	public void setValue(String name, double value, double alpha) {
-		String key = getMetricKey(name);
-		RichGauge g = findOne(name);
-		if (g == null) {
-			g = new RichGauge(name);
-		}
-		MetricUtils.setRichGaugeValue(g, value, alpha);
-		getValueOperations().set(key, serialize(g));
+	public void setValue(final String name, final double value, final double alpha) {
```
markpollack: Can you also rebase to incorporate new method name for setValue

ericbottard:
Used a simple policy as IMO:
- the gauge needs to reflect the value in a timely manner, it does not make sense to have a 5secs outdated value
- there is either a) little chance that a conflicting write will last more than a couple ms for the same key or else b) there is a lot of contention for that particular key anyway, and in all cases we're screwed, however complicated the policy is

Would you agree?

markpollack:
I agree that 5 seconds is too long but a tight spin loop is too short.  An
policy that tries after 50 ms, then 100 ms, then 200 ms or something like
that up to a max of one second is what I am suggesting.

On Jan 13, 2015, at 9:35 AM, Eric Bottard <notifications@github.com> wrote:

Used a simple policy as IMO:

   - the gauge needs to reflect the value in a timely manner, it does not
   make sense to have a 5secs outdated value
   - there is either a) little chance that a conflicting write will last
   more than a couple ms for the same key or else b) there is a lot of
   contention for that particular key anyway, and in all cases we're screwed,
   however complicated the policy is

Would you agree?

—
Reply to this email directly or view it on GitHub
<https://github.com/spring-projects/spring-xd/pull/1374#issuecomment-69752452>
.

markpollack:
Merged as c442c5102c745ee25e59132542a7783489397251

PR #1367, opened 2015-01-09
http source now accepts binary content larger that a few KB

prismec:
The problem was the assumption that the ChannelBuffer is always backed
by an array - which is not the case if you send e.g. binary data that
is larger than around 10KB. The solution is to check the availability
of a backing byte[] with the hasArray() method.

Note that after this fix the source is still not able to accept
messages > ~1MB in size. This affects all messages, binary and others.
I guess that this must be somehow fixed in the netty setup ? I’ve just
not investigated this since this fix solves my current problem.

PR #1377, opened 2015-01-13
[Sprint: 41] XD-2590: Handle Large Messages in HTTP Source

garyrussell:
JIRA: https://jira.spring.io/browse/XD-2590

Based on original work by Peter Rietzler:

http source now accepts binary content larger that a few KB

The problem was the assumption that the ChannerBuffer is always backed
by an array - which is not the case if you send e.g. binary data that
is larger than around 10KB. The solution is to check the availability
of a backing byte[] with the hasArray() method.

- move the decoding work to a `MessageConverter`
- allow the message converter to be configured
- add a setter for the max content length; add error handling when the content length is exceeded
  - previously, the client hung waiting for a reply.
  - close the channel in the catch all `exceptionCaught` method and log the error.
- add properties to the http source (maxContentLength, messageConverterClass)

__replaces #1367 __

### DIFF
--- a/modules/source/http/config/http.xml
+++ b/modules/source/http/config/http.xml
@@ -13,8 +13,16 @@
 		<beans:property name="autoStartup" value="false"/>
 		<beans:property name="outputChannel" ref="output"/>
 		<beans:property name="sslPropertiesLocation" value="${sslPropertiesLocation}"/>
+		<beans:property name="maxContentLength" value="${maxContentLength}"/>
+		<beans:property name="messageConverter" ref="converter"/>
 	</beans:bean>
 
+	<beans:bean id="converter" class="org.springframework.integration.x.http.NettyInboundMessageConverter"/>
+
 	<channel id="output"/>
+
+	<beans:beans profile="customConverter">
+		<beans:bean id="converter" class="${messageConverterClass}"/>
+	</beans:beans>
 
 </beans:beans>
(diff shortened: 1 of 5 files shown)


modules/source/http/config/http.xml
```diff
@@ -13,8 +13,16 @@
 		<beans:property name="autoStartup" value="false"/>
 		<beans:property name="outputChannel" ref="output"/>
 		<beans:property name="sslPropertiesLocation" value="${sslPropertiesLocation}"/>
+		<beans:property name="maxContentLength" value="${maxContentLength}"/>
+		<beans:property name="messageConverter" ref="converter"/>
 	</beans:bean>
 
+	<beans:bean id="converter" class="org.springframework.integration.x.http.NettyInboundMessageConverter"/>
```
ericbottard: The default handling for the converter is cumbersome in my opinion:
1. Why not have the `messageConverterClass` default to `NettyInboundMessageConverter`? No profiles needed then, and the user knows about it (it is a long String displayed when doing `module:info` though, but we have it in other modules IIRC)
2. Moreover, there is some some defaulting behavior in `doStart()` of the Adapter, that takes an extra argument in the constructor.

ericbottard:
Other than the minor converter thing, LGTM.

modules/source/http/config/http.xml
garyrussell: Makes sense; will rework.

garyrussell:
Rebased, pushed; reworked; removed profile; added the `constructor-arg` - I decided to leave the safety net in `doStart()`.

ericbottard:
Squashed and merged as 4ebf7fdcdbda3e16f96be3ca7acfd6ff58002e8e

PR #1380, opened 2015-01-14
[DO NOT MERGE: FOR REVIEW ONLY] Spark Streaming Integration

ilayaperumalg:
- Spark streaming module changes
	- Spark streaming `Processor`
    	- This interface provides the method `process` that end user would implement
        - The process method has the transformation operation to be applied on input DStream
and returns the output DStream. In XD's context this represents a `processor` module
        - The same process method can be implemented for a `sink` module with the return value set to null
        - The spark configuration properties can be specified via @SparkConfig annotation
    - SparkStreamingDriverModule
        - This is the spark `driver` that creates the SparkStreamingContext.
        - This is defined as a Spring XD module (processor or sink)
        - Computes spark streaming application and XD dependent jars before setting `spark.jars`
        - Uses Spark's streaming listener to verify the streaming receiver start status
- Spark streaming module deployment changes
    - SparkStreamingPlugin
    	- This plugin adds necessary beans required for the spark streaming module
    		- Sets up MessageBusReceiver
    			- This is the actual spark streaming receiver that connects to the underlying XD
    			message bus to receive messages
    			- Adds the message bus properties from XD container environment to the spark streaming
    			module so that all those properties are reused at the Spark executors when the receiver
    			is setup and running.
    		- Sets up MessageBusSender (in case of processor module)
    			- This sends the processed messages to the message bus so any other XD module could pick up
    - ContainerFilter
    	- This interface defines the method to filter out matched containers by applying any specific runtime
    	criteria
    - SparkStreamingContainerFilter
    	- filters out the containers that have a spark streaming module running.This is necessary as
    	multiple spark contexts can not be created on the same JVM-SPARK-2243.
    - Add ModuleExecutionFramework
    	- Introduce module execution framework `spark` via hidden module option
    	- This will be the key to identify if the underlying module is of spark streaming module.

- Add tests for spark streaming processor and log modules
  	- Add SparkStreamingTests class that can be run across different transports.
  	- Add test modules: SparkWordCount and SparkLog
- Fixes and polishing

    - Reduce the number of jars for spark application
    - Fix message bus dependency
    - Fix jetty server dependencies

   - Polishing


dependencies.properties
```diff
@@ -226,10 +226,10 @@ org.crashub\:crash.embed.spring=1.3.0
 org.crashub\:crash.plugins.cron=1.3.0
 org.crashub\:crash.plugins.mail=1.3.0
 org.crashub\:crash.shell=1.3.0
-org.eclipse.jetty\:jetty-annotations=9.2.4.v20141103
-org.eclipse.jetty\:jetty-jsp=9.2.4.v20141103
-org.eclipse.jetty\:jetty-util=9.2.4.v20141103
-org.eclipse.jetty\:jetty-webapp=9.2.4.v20141103
+org.eclipse.jetty\:jetty-annotations=8.1.14.v20131031
```
ilayaperumalg: May be we can add a comment here mentioning jetty version needs to `8.1.14.v20131031` for spark streaming. Also, we need to investigate if this has any impact on other XD components.

build.gradle
```diff
@@ -69,6 +69,7 @@ allprojects {
 
 	repositories {
 		mavenCentral()
+		mavenLocal()
```
ericbottard: This should be removed

build.gradle
```diff
@@ -821,6 +824,7 @@ project('spring-xd-module') {
 		compile "org.springframework.boot:spring-boot-loader"
 		compile "com.fasterxml.jackson.core:jackson-databind"
 		compile project(':spring-xd-module-spi')
+		compile project(':spring-xd-spark-streaming')
```
ericbottard: This feels wrong to me: why would we depend on spark stuff for the core module contract?

build.gradle
```diff
@@ -833,6 +837,32 @@ project('spring-xd-module-spi') {
...
+	dependencies {
+		compile("org.apache.spark:spark-streaming_$sparkScalaVersion:$sparkVersion") {
+			exclude group: 'hsqldb'
+		}
+		compile "org.springframework.integration:spring-integration-core"
+		compile project(":spring-xd-module-spi")
+	}
+}
+
+project('spring-xd-spark-streaming-tests') {
```
ericbottard: Not saying that this is bad, but definitely not the approach we have for testing other modules, right?

dependencies.properties
ericbottard: Also, can you explain "why"? _i.e._ what is different in 8.x, in a nutshell

gradle/build-dist.gradle
```diff
@@ -50,10 +50,19 @@ task copyZookeeperInstall(type: Copy) {
...
 
 task copyXDInstall(type: Copy, dependsOn: [":spring-xd-dirt:build", ":spring-xd-dirt:installApp"]) {
+	def versions = [:]
+	def file = rootProject.file('dependencies.properties')
+		file.withReader { reader ->
+			def properties = new Properties()
+			properties.load(reader)
+			versions << properties
+		}
+	def version = versions["com.google.protobuf:protobuf-java"]
```
ericbottard: This logic is starting to get crazy.
In any case, if we go through with this, need to rename `version` to something more descriptive (_e.g_ `protobufVersion`)

spring-xd-dirt/src/main/java/org/springframework/xd/dirt/cluster/ContainerFilter.java
```diff
@@ -0,0 +1,33 @@
...
+
+import org.springframework.xd.module.ModuleDescriptor;
+
+/**
+ * A filter interface that applies filter for the given containers and returns
+ * eligible containers for deployment.
+ *
+ * @author Ilayaperumal Gopinathan
+ */
+public interface ContainerFilter {
```
ericbottard: Why the need for yet another interface. This seems like a specialization of `ContainerMatcher`

spring-xd-dirt/src/main/java/org/springframework/xd/dirt/integration/bus/KafkaConnectionPropertyNames.java
```diff
@@ -0,0 +1,36 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.springframework.xd.dirt.integration.bus;
+
+/**
```
ericbottard: Javadoc

spring-xd-dirt/src/main/java/org/springframework/xd/dirt/integration/bus/RabbitConnectionPropertyNames.java
```diff
@@ -0,0 +1,41 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.springframework.xd.dirt.integration.bus;
+
+/**
```
ericbottard: Javadoc

spring-xd-dirt/src/main/java/org/springframework/xd/dirt/integration/bus/RedisConnectionPropertyNames.java
```diff
@@ -0,0 +1,41 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.springframework.xd.dirt.integration.bus;
+
+/**
```
ericbottard: Javadoc

spring-xd-dirt/src/main/java/org/springframework/xd/dirt/plugins/spark/MessageBusConfiguration.java
```diff
@@ -0,0 +1,67 @@
...
+ *
+ * @author Ilayaperumal Gopinathan
+ */
+@Configuration
+@Import(PropertyPlaceholderAutoConfiguration.class)
+@ImportResource({"classpath*:" + ConfigLocations.XD_CONFIG_ROOT + "bus/${XD_TRANSPORT}-bus.xml",
+		ConfigLocations.XD_CONFIG_ROOT + "bus/codec.xml"})
+class MessageBusConfiguration {
+
+	static ConfigurableApplicationContext createApplicationContext(final Properties properties) {
```
ericbottard: Javadoc should explain who is calling this method

spring-xd-dirt/src/main/java/org/springframework/xd/dirt/plugins/spark/MessageBusReceiver.java
```diff
@@ -0,0 +1,115 @@
...
+import org.springframework.messaging.Message;
+import org.springframework.xd.dirt.integration.bus.MessageBus;
+
+/**
+ * Spark {@link Receiver} implementation that binds to the MessageBus as a consumer.
+ * 
+ * @author Mark Fisher
+ * @author Ilayaperumal Gopinathan
+ */
+class MessageBusReceiver extends Receiver {
```
ericbottard: Do we want "Spark" in the classname?

spring-xd-dirt/src/main/java/org/springframework/xd/dirt/spark/SparkStreamingContainerFilter.java
```diff
@@ -0,0 +1,92 @@
...
+import org.springframework.xd.module.ModuleDescriptor;
+import org.springframework.xd.module.core.ModuleFactory;
+import org.springframework.xd.module.options.ModuleOptions;
+import org.springframework.xd.module.options.ModuleOptionsMetadata;
+import org.springframework.xd.module.options.ModuleOptionsMetadataResolver;
+import org.springframework.xd.spark.streaming.Processor;
+
+/**
+ * A {@link org.springframework.xd.dirt.cluster.ContainerFilter} which filters out the containers that have a spark
+ * streaming module running.This is necessary as multiple spark contexts can not be created on the same JVM-SPARK-2243.
```
ericbottard: Space after dot

ericbottard: End of sentence is weird. Should be "on the same JVM. See SPARK-2243" I assume

build.gradle
markfisher: This was to avoid a dependency cycle, but related to your comment immediately above, it would be ideal to avoid dependencies on the spark code from dirt. However, that will require significant changes to ModuleFactory and others that are beyond the scope of this PR (and unfortunately most likely beyond the scope of 1.1).

spring-xd-dirt/src/main/java/org/springframework/xd/dirt/plugins/spark/MessageBusReceiver.java
markfisher: Package name is a sufficient qualifier here, although we might want to add a "streaming" subpackage. For the record, those classes that *do* contain Spark in their name are for the extension points of XD infrastructure itself, typically subclasses or interface impls where the name itself should include qualification (e.g. the Plugin impl and the Module impl), whereas the classes that are merely supporting those should have simple names - and live in appropriate packages, especially these non-public classes (internal support classes).

spring-xd-spark-streaming-tests/src/test/java/org/springframework/xd/spark/streaming/SparkStreamingTests.java
```diff
@@ -0,0 +1,81 @@
...
+	public void testSparkLog() throws Exception {
+		String fileName = getClass().getSimpleName() + new Random().nextInt() + ".txt";
+		File fileToDelete = new File(fileName);
+		try {
+			final HttpSource source = newHttpSource();
+			final String stream = String.format("%s | spark-log --filePath=%s", source, fileName);
+			stream().create(generateStreamName(), stream);
+			source.ensureReady().postData(messageToPost);
+			//explicit sleep to wait for the message to get processed.
+			Thread.sleep(2000);
```
pperalta: The `sleep` and file reading/matching should be replaced with `EventuallyMatcher`.

spring-xd-module/src/main/java/org/springframework/xd/module/core/ModuleFactory.java
```diff
@@ -57,6 +60,8 @@
 
 	private ResourcePatternResolver resourceLoader = new PathMatchingResourcePatternResolver();
 
+	public static final String MODULE_EXECUTION_FRAMEWORK = "moduleExecutionFramework";
```
ericbottard: Add javaodc

spring-xd-module/src/main/java/org/springframework/xd/module/core/ModuleFactory.java
```diff
@@ -94,6 +99,10 @@ public Module createModule(ModuleDescriptor moduleDescriptor, ModuleDeploymentPr
 	 */
 	private Module createAndConfigureModuleInstance(ModuleDescriptor moduleDescriptor, ModuleOptions moduleOptions,
 			ModuleDeploymentProperties deploymentProperties) {
+		String name = (String) moduleOptions.asPropertySource().getProperty(MODULE_EXECUTION_FRAMEWORK);
+		if (Processor.MODULE_EXECUTION_FRAMEWORK.equals(name)) {
```
ericbottard: Having two constants named the same makes things hard to follow here

spring-xd-module/src/main/java/org/springframework/xd/module/core/ModuleFactory.java
```diff
@@ -94,6 +99,10 @@ public Module createModule(ModuleDescriptor moduleDescriptor, ModuleDeploymentPr
...
+			return createSparkModule(moduleDescriptor, moduleOptions, deploymentProperties);
```
ericbottard: Should see what Rx/Reactor brings, but the main flow of decisions seems weird here. Also, you're not leveraging the module ClassLoader, which seems wrong.

ericbottard: At this point in the code, there should only be 2 "main" cases here: Simple vs Composed.

spring-xd-module/src/main/java/org/springframework/xd/module/core/ModuleFactory.java
```diff
@@ -135,6 +144,18 @@ else if (JavaConfiguredModule.basePackages(moduleDefinition, moduleClassLoader).
 		return null;
 	}
 
+	private Module createSparkModule(ModuleDescriptor moduleDescriptor, ModuleOptions moduleOptions,
```
ericbottard: OK, this seems to confirm what I was thinking: I believe this should move to a case in SimpleModuleCreator, not here

spring-xd-module/src/main/java/org/springframework/xd/module/spark/SparkStreamingDriverModule.java
```diff
@@ -0,0 +1,329 @@
...
+			throw new RuntimeException(ioe);
+		}
+		// Get necessary dependencies from XD DIRT.
+		URLClassLoader parentClassLoader = (URLClassLoader) this.getClassLoader().getParent();
+		URL[] urls = parentClassLoader.getURLs();
+		for (URL url : urls) {
+			String file = url.getFile().split("\\!", 2)[0];
+			//if (file.endsWith(".jar")) {
+			if (file.endsWith(".jar") && (// Add spark jars
+					file.contains("spark") ||
```
ericbottard: Man, this is ugly :(

spring-xd-module/src/main/java/org/springframework/xd/module/spark/package-info.java
```diff
@@ -0,0 +1,21 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/**
+ * Package for Spark module implementations.
+ */
+
+package org.springframework.xd.module.spark;
```
ericbottard: Should we clearly separate spark from spark-streaming?

spring-xd-module/src/main/java/org/springframework/xd/module/core/ModuleFactory.java
markfisher: Agreed (for now), but eventually we should refactor the ModuleFactory to support a contribution model (also relevant for the comments above about dependencies).

spring-xd-module/src/main/java/org/springframework/xd/module/spark/package-info.java
markfisher: yes

spring-xd-shell/src/test/java/org/springframework/xd/shell/AbstractShellIntegrationTest.java
```diff
@@ -89,8 +90,11 @@
 	@BeforeClass
 	public static synchronized void startUp() throws InterruptedException, IOException {
 		RandomConfigurationSupport randomConfigSupport = new RandomConfigurationSupport();
+		// todo: Ability to verify if the transport is valid and the required message broker is setup.
+		String transport = System.getProperty("XD_TRANSPORT");
```
ericbottard: This feels wrong:
1. either we set things in stone (like it was before)
2. or we allow selecting a transport to be selected with a prop, but then the defaults are coded at two places: here and in SingleNodeApplication itself

spring-xd-shell/src/test/java/org/springframework/xd/shell/command/ModuleListTests.java
```diff
@@ -16,18 +16,16 @@
...
 package org.springframework.xd.shell.command;
 
-import static com.google.common.collect.Lists.newArrayList;
-import static com.google.common.collect.Maps.transformValues;
-import static org.hamcrest.Matchers.allOf;
-import static org.hamcrest.Matchers.equalTo;
-import static org.hamcrest.Matchers.hasEntry;
-import static org.hamcrest.Matchers.hasSize;
-import static org.junit.Assert.assertThat;
+import static com.google.common.collect.Lists.*;
```
ericbottard: Seems like your rules for imports are off (or is this what we should be using?)

Also noticed in another class above, but this has no other change apart from imports

spring-xd-shell/src/test/resources/log4j.properties
```diff
@@ -0,0 +1,8 @@
+log4j.rootLogger=INFO, stdout
```
ericbottard: I believe this should not be put in VCS

spring-xd-spark-streaming-tests/src/test/java/org/springframework/xd/spark/streaming/SparkStreamingTests.java
```diff
@@ -0,0 +1,81 @@
...
+ * @author Ilayaperumal Gopinathan
+ */
+public class SparkStreamingTests extends AbstractStreamIntegrationTest {
+
+	static {
+		// Change the transport here to run the test against other message bus transports.
+		System.setProperty("XD_TRANSPORT", "local");
+	}
+
+	private static final String messageToPost = "foo foo foo";
```
ericbottard: Not sure using a constant really helps readability, but if we do, should be UPPERCASE

spring-xd-spark-streaming-tests/src/test/java/org/springframework/xd/spark/streaming/SparkStreamingTests.java
```diff
@@ -0,0 +1,81 @@
...
+		final String stream = String.format("%s | spark-word-count | %s", source, sink);
+		stream().create(generateStreamName(), stream);
+		source.ensureReady().postData(messageToPost);
+		assertThat(sink, XDMatchers.eventually(XDMatchers.hasContentsThat(equalTo("(foo,3)"))));
+	}
+
+	@Test
+	public void testSparkLog() throws Exception {
+		String fileName = getClass().getSimpleName() + new Random().nextInt() + ".txt";
+		File fileToDelete = new File(fileName);
```
ericbottard: This creates the file in the current working dir, not sure this is a good idea. There is support for temp files _e.g._ as JUnit Rules (and others)

spring-xd-spark-streaming/src/main/java/org/springframework/xd/spark/streaming/Processor.java
```diff
@@ -0,0 +1,43 @@
...
+import org.apache.spark.streaming.api.java.JavaDStreamLike;
+
+/**
+ * Interface for modules using the Spark Streaming API to process messages from the Message Bus.
+ *
+ * @author Mark Fisher
+ */
+public interface Processor extends Serializable {
+
+	public static final String SPARK_MASTER_URL_PROP = "spark.master";
```
ericbottard: The meaning of those constants should be documented, esp. MODULE_EXECUTION_FRAMEWORK

Are we trying to have an all purpose Processor interface?

spring-xd-spark-streaming/src/main/java/org/springframework/xd/spark/streaming/examples/SparkLog.java
```diff
@@ -0,0 +1,96 @@
...
+
+import org.springframework.xd.spark.streaming.Processor;
+import org.springframework.xd.spark.streaming.SparkConfig;
+
+/**
+ * @author Mark Fisher
+ * @author Ilayaperumal Gopinathan
+ */
+@SuppressWarnings({ "unchecked", "rawtypes", "serial" })
+public class SparkLog implements Processor {
```
ericbottard: Should this make it in the codebase in the end? In `src/main/java` ?

spring-xd-spark-streaming/src/main/java/org/springframework/xd/spark/streaming/examples/SparkLog.java
```diff
@@ -0,0 +1,96 @@
...
+	public JavaDStreamLike process(JavaDStreamLike input) {
+		input.foreachRDD(new Function<JavaRDD, Void>() {
+
+			@Override
+			public Void call(JavaRDD rdd) {
+				rdd.foreachPartition(new VoidFunction<Iterator<?>>() {
+
+					@Override
+					public void call(Iterator<?> items) throws Exception {
+						final StringBuffer buffer = new StringBuffer(sdf.format(new Date()));
```
ericbottard: Not sure what the threading guarantees are here, but for the sake of argument, that call to `sdf.format()` is not thread safe.

spring-xd-spark-streaming/src/main/java/org/springframework/xd/spark/streaming/examples/SparkLog.java
```diff
@@ -0,0 +1,96 @@
...
+						while (items.hasNext()) {
+							buffer.append(items.next());
+						}
+						try {
+							FileWriter fw = new FileWriter(file.getAbsoluteFile());
```
ericbottard: So is that file access... So, how is the threading model taken care of here?

spring-xd-spark-streaming/src/main/java/org/springframework/xd/spark/streaming/examples/SparkLogModuleOptionsMetadata.java
```diff
@@ -0,0 +1,37 @@
...
+
+package org.springframework.xd.spark.streaming.examples;
+
+import org.springframework.xd.module.options.spi.ModuleOption;
+import org.springframework.xd.spark.streaming.DefaultSparkStreamingModuleOptionsMetadata;
+
+/**
+ * @author Ilayaperumal Gopinathan
+ */
+public class SparkLogModuleOptionsMetadata extends DefaultSparkStreamingModuleOptionsMetadata {
```
ericbottard: extends or MixIn ?

spring-xd-dirt/src/main/java/org/springframework/xd/dirt/plugins/spark/MessageBusSender.java
```diff
@@ -0,0 +1,90 @@
...
+	}
+
+	public MessageBusSender(LocalMessageBusHolder messageBusHolder, String outputChannelName, Properties messageBusProperties, Properties moduleProducerProperties) {
+		this.messageBusHolder = messageBusHolder;
+		this.outputChannelName = outputChannelName;
+		this.messageBusProperties = messageBusProperties;
+		this.moduleProducerProperties = moduleProducerProperties;
+	}
+
+	public void start() {
```
mminella: Should have `@Override`

spring-xd-dirt/src/main/java/org/springframework/xd/dirt/plugins/spark/MessageBusSender.java
```diff
@@ -0,0 +1,90 @@
...
+			}
+			else {
+				applicationContext = MessageBusConfiguration.createApplicationContext(messageBusProperties);
+				messageBus = applicationContext.getBean(MessageBus.class);
+			}
+			messageBus.bindProducer(outputChannelName, this, moduleProducerProperties);
+		}
+	}
+
+	public void stop() {
```
mminella: Should have `@Override`

ericbottard:
Overall, I don't understand the whole fiddling with the MessageBus, can you elaborate?

markfisher:
Can you ask a more direct question?

spring-xd-spark-streaming/src/main/java/org/springframework/xd/spark/streaming/examples/SparkLogModuleOptionsMetadata.java
markfisher: I probably don't fully understand what you have in mind by a Mixin alternative to this, but I can explain the idea. Any Spark Streaming module should have the DefaultSparkStreamingModuleOptionsMetadata since that provides common general settings like "batchInterval" as well as the indicator that it's actually a Spark Streaming module. However, if a developer is creating a custom implementation of Processor and that will accept some option values (like filePath here), they need the simplest possible way to add those. If they can do something that does not require an explicit extends, that would be even better, but can we then handle the mixin of the defaults within the framework so that is also hidden from the user?

spring-xd-shell/src/test/java/org/springframework/xd/shell/command/ModuleListTests.java
pperalta: Yeah the default in IDEA is to squash 10 or so imports into a wild card. This can be changed in the settings.

spring-xd-spark-streaming/src/main/java/org/springframework/xd/spark/streaming/examples/SparkLog.java
pperalta: +1 on `sdf`, and the questions Eric raised on thread safety. 

Also, `StringBuilder` can be used here instead of `StringBuffer`.

spring-xd-dirt/src/main/java/org/springframework/xd/dirt/plugins/spark/MessageBusSender.java
```diff
@@ -0,0 +1,90 @@
...
+
+	private final String outputChannelName;
+
+	private final LocalMessageBusHolder messageBusHolder;
+
+	private final Properties messageBusProperties;
+
+	private final Properties moduleProducerProperties;
+
+	private MessageBus messageBus;
```
pperalta: If `messageBus` can be accessed by multiple threads, it should be marked `volatile` or access to it should be synchronized.

spring-xd-dirt/src/main/java/org/springframework/xd/dirt/plugins/spark/MessageBusSender.java
pperalta: Should start/stop be synchronized, or do we assume this will always be invoked by a single thread?

spring-xd-dirt/src/main/java/org/springframework/xd/dirt/cluster/ContainerFilter.java
pperalta: +1. The existing `ContainerMatcher` interface already has a method defined that closely matches the one specified by this interface. Implementations of `ContainerMatcher` can be composed; see `RedeploymentContainerMatcher` as an example.

ericbottard:
I'm talking about the stuff happening in MessageBusConfiguration in particular. Thinking about it some more, seems weird that a module needs to be "bus-aware". Moreover, the code in MessageBusConfiguration seems to duplicate bus creation logic that we have elsewhere, or am I missing something? What triggered my interest is the fact that the bus classloader logic is not present for example.
Maybe I need to have a bigger picture view of the PR, but comparing with the Rx stuff I reviewed for example, seems that this PR is more "intrusive". Granted, Mark's Rx stuff does not cover the Plugin aspect (only dealt with the SI Handler), but still seems that we know too much about the bus here.

spring-xd-spark-streaming/src/main/java/org/springframework/xd/spark/streaming/examples/SparkLogModuleOptionsMetadata.java
ericbottard: Our `org.springframework.xd.module.options.spi.Mixin` is the way we usually use to do that (adding "multiple inheritance" if needed).
Not sure I understand your last sentence though? All options are "user-visible", even here with `extends`

markfisher:
The significant difference between this and the Rx stuff, which also explains the need for the  MessageBusConfiguration at a high level, is that the MessageBusReceiver (an impl of Spark Streaming's Receiver) is shipped to the Spark cluster to be run there. So, think of it as the minimal XD shim that is necessary to interact with XD modules. It is a process running *on that cluster* that binds to the MessageBus similar to any XD module (including the Rx ones) and then processes the Messages via Spark Streaming operations as provided by an impl of our spark.streaming.Processor. Likewise, the results of processing can be returned via the output DStream, and those are then sent to the XD MessageBus.... in that case within the MessageBusSender instances, each one running in the tasks on worker nodes on the cluster, so again a shim is needed to bind these "producers" to the bus so that a downstream XD module receives those results.

spring-xd-spark-streaming/src/main/java/org/springframework/xd/spark/streaming/examples/SparkLogModuleOptionsMetadata.java
markfisher: Sorry by "hidden from the user", I actually meant "not a concern for the developer". In other words, if a developer is providing a Processor impl (thus providing a custom XD module that uses Spark Streaming), we want them to have the simplest possible experience. In the case where you only need the default options, you would simply add DefaultSparkStreamingModuleOptionsMetadata to the properties file. If you need other options in addition to those, you would provide SomeModuleOptionsMetadata impl of your own. It seems that using "extends" there is pretty simple, but if you think there's a simpler option *from that module developer's perspective*, we should consider it.

ericbottard:
It all comes back to me now, but not entirely: 
One subsequent question then: isn't our processor acting as a "Driver" and gathering the results?

Or are you saying that the remote processes (running on the Spark cluster) are accessing the bus themselves?

markfisher:
The driver role is in SparkStreamingDriverModule. It actually delegates to SparkStreamingModuleExecutor, and that is where the code is actually a mix of stuff that runs on the driver side and then nested functions that run on the worker node. Where you see the actual *send* call, that is running on tasks on worker nodes (i.e. we do not "gather" results on the driver side).

spring-xd-spark-streaming/src/main/java/org/springframework/xd/spark/streaming/examples/SparkLogModuleOptionsMetadata.java
ericbottard: In this case, I think both extension and mixin amount to the same. Extension maybe even is more "natural".

As a side note, should we come up with a way that `DefaultSparkStreamingModuleOptionsMetadata` are included (maybe in the plugin?), in which case introspection via extension is easier (while mixing in should still be legal IMO)

markfisher: I think it might be a bit of a chicken and egg problem, since we only know it's a Spark Streaming module based on the "spark" value for moduleExecutionFramework which is provided via the options. But if there's a cleaner alternative, we should do it... would involve changing the way we detect the Spark Streaming type of module.

spring-xd-spark-streaming/src/main/java/org/springframework/xd/spark/streaming/examples/SparkLog.java
markfisher: Right now this code is only used by the test, and only invoked ONCE :) ... however, good point on the thread safety for SDF and for using StringBuilder.

spring-xd-shell/src/test/resources/log4j.properties
ilayaperumalg: This file was added to avoid a conflict from kafka tests jar's log4j.properties. As you can notice the upstream master test reports would use kafka tests jar's log4j and I believe it is set to WARN level.

spring-xd-dirt/src/main/java/org/springframework/xd/dirt/cluster/ContainerFilter.java
markfisher: We are going to try to refactor this to rely on composition. The main requirement was to contribute the logic to ensure that no more than one Spark Streaming module (driver) is being deployed to an XD Container. The idea was to have a more general contribution model here (we find any bean of type ContainerFilter), and that approach is not yet available for ContainerMatcher (find any bean of type ContainerMatcher and add). Will explore.

build.gradle
ilayaperumalg: The reason we do that here is because of the ModuleFactory requires to know about `SparkStreamingDriverModule`. Once we address that we can eliminate this dependency.

spring-xd-dirt/src/main/java/org/springframework/xd/dirt/plugins/spark/MessageBusSender.java
ilayaperumalg: Actually the `messageBus` is accessed by a single thread only once (when binding happens).

spring-xd-module/src/main/java/org/springframework/xd/module/spark/SparkStreamingDriverModule.java
ilayaperumalg: yeah, agreed ;) but the fact that this significantly reduced the number of jars sent across to spark executor running on the outside cluster. This code still can be optimized once we have a simple way to setup message bus configuration (for MessageBusReceiver/Sender) on the spark executor side.

spring-xd-spark-streaming-tests/src/test/java/org/springframework/xd/spark/streaming/SparkStreamingTests.java
markfisher: https://github.com/ilayaperumalg/spring-xd/pull/4

spring-xd-spark-streaming/src/main/java/org/springframework/xd/spark/streaming/examples/SparkLog.java
ilayaperumalg: The spark module implementation has to be inside the jar file as only jar files are included (sent across to spark cluster) as spark.jars when creating the spark streaming context.

spring-xd-dirt/src/main/java/org/springframework/xd/dirt/cluster/ContainerFilter.java
markfisher: I started down this path of refactoring the ContainerFilter to just be another ContainerMatcher (and I have a CompositeContainerMatcher that delegates), but I'm actually questioning that approach now. The main concern is that DefaultContainerMatcher is very particular about the way results are ordered and finalized. It's important for the round-robin distribution of module deployments. If we are providing an extension point, we do not want to lose that control. Therefore, the approach in this PR does seem correct in the sense that the *filtering* of containers happens within the scope of the DefaultContainerMatcher itself, after the initial criteria check but *before* the final distributeForRequestedCount(). So rather than having a delegating ContainerMatcher that iterates a list, we really do want this filtering to happen in that well-defined scope of the existing algorithm. Given that, I'm actually much less opposed to the idea of defining an explicit (even if similar) interface that is intended as a "safe" extension point within those controlled bounds. Thoughts?

markfisher: After discussion with @pperalta this is the proposed plan:

1. merge this PR without changing anything related to ContainerMatcher or ContainerFilter yet
2. refactor RedploymentContainerMatcher so that it implements ContainerFilter
3. remove the ContainerMatcher interface and rename DefaultContainerMatcher to just ContainerMatcher
4. move ContainerMatcher to the .server package *and* make it non-public

markfisher: Actually, the above has been applied on this PR.

spring-xd-dirt/src/main/java/org/springframework/xd/dirt/plugins/spark/streaming/MessageBusReceiver.java
```diff
@@ -0,0 +1,116 @@
...
+	/**
+	 * The {@link DirectChannel} that stores the received messages into Spark's memory.
+	 */
+	private class MessageStoringChannel extends DirectChannel implements Serializable {
+
+		private static final long serialVersionUID = 1L;
+
+		@Override
+		protected boolean doSend(Message<?> message, long timeout) {
+			//todo: probably just make this take the payload as is (especially if the inputType is applied as usual)...
```
markfisher: we need to remove this todo since we are now just sending the payload as is

we might want to create a JIRA issue for the configurable MessageConverter

spring-xd-dirt/src/main/resources/META-INF/spring-xd/internal/admin-server.xml
```diff
@@ -6,11 +6,19 @@
 
 	<import resource="deployers.xml" />
 
+	<import resource="../sparkstreaming/container-filter.xml" />
```
markfisher: I think we can remove this line and the container-filter.xml file now that the SparkStreamingContainerFilter bean is provided directly on the ContainerMatcher.

Alternatively, if we want it to be more extensible, we could have ContainerMatcher find any ContainerFilter beans from the ApplicationContext similar to how this was implemented for the initial version of this PR (although I think that can wait until post 1.1)

spring-xd-dirt/src/main/resources/META-INF/spring-xd/plugins/spark.xml
```diff
@@ -0,0 +1,7 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
markfisher: see comment above - this file can be removed now

spring-xd-dirt/src/main/resources/META-INF/spring-xd/sparkstreaming/container-filter.xml
```diff
@@ -0,0 +1,9 @@
+<?xml version="1.0" encoding="UTF-8"?>
```
markfisher: see comment above - this file can be removed now

spring-xd-spark-streaming/src/main/java/org/springframework/xd/spark/streaming/Processor.java
markfisher: That was the original idea behind XD-2465, but now I think the plan is to have the Reactor and RxJava module support follow a similar plugin-based approach (with a module execution framework indicator) to what we're doing here with Spark Streaming. That will be on the roadmap after the 1.1 release.

markfisher:
rebased, polished, squashed, and merged

also created this follow up issue: https://jira.spring.io/browse/XD-2644

PR #1382, opened 2015-01-15
[Sprint 41] XD-2570 - Create a new Broadcast Stream per thread

markpollack:
Also XD-2586 Update to Reactor 2.0 build snapshots


build.gradle
```diff
@@ -345,6 +345,7 @@ project('spring-xd-reactor') {
 	dependencies {
 		compile "io.projectreactor:reactor-core"
 		compile "org.springframework.integration:spring-integration-core"
+		compile "org.springframework.integration:spring-integration-expression"
```
ericbottard: Fool me once, shame on you. Fool me twice, shame on me: I'm not falling in your trap again; The groupId/artifactId is wrong here

spring-xd-reactor/src/main/java/org/springframework/xd/reactor/MultipleBroadcasterMessageHandler.java
```diff
@@ -0,0 +1,196 @@
...
+import reactor.rx.stream.Broadcaster;
+
+import java.lang.reflect.Method;
+import java.util.Hashtable;
+import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ConcurrentMap;
+
+/**
+ * Adapts the item at a time delivery of a {@link org.springframework.messaging.MessageHandler}
```
ericbottard: Struck me too when reading the Rx version, maybe this should read "item-at-a-time" (with dashes). At least for non-english natives, eases reading

spring-xd-reactor/src/main/java/org/springframework/xd/reactor/MultipleBroadcasterMessageHandler.java
```diff
@@ -0,0 +1,196 @@
...
+ * by delegating processing to an Stream based on a partitionExpression.
+ * <p/>
+ * The specific Stream that the message is delegated is determined by the partitionExpression value.
```
ericbottard: missing "to" in "the message is delegated to"

spring-xd-reactor/src/main/java/org/springframework/xd/reactor/MultipleBroadcasterMessageHandler.java
```diff
@@ -0,0 +1,196 @@
...
+            broadcasterToUse.onNext(message);
+        } else if (ClassUtils.isAssignable(inputType.getRawClass(), message.getPayload().getClass())) {
+            broadcasterToUse.onNext(message.getPayload());
+        } else {
+            throw new MessageHandlingException(message, "Processor signature does not match [" + message.getClass()
+                    + "] or [" + message.getPayload().getClass() + "]");
+        }
+
+        if (logger.isDebugEnabled()) {
+            Controls controls = this.controlsMap.get(Thread.currentThread().getId());
```
ericbottard: Hardcoded expression here, which may not be the one we want

build.gradle
markpollack: Doh!

spring-xd-reactor/src/main/java/org/springframework/xd/reactor/MultipleBroadcasterMessageHandler.java
markpollack: sure

spring-xd-reactor/src/main/java/org/springframework/xd/reactor/MultipleBroadcasterMessageHandler.java
markpollack: I think we have this approach in a bunch of other places...e.g. hardcoded....

ericbottard: What I'm saying is that this map is populated by using the SpEL partition expression. There is no reason that it maps to either a String, let alone a Thread id

markpollack: Yes, this is indeed wrong.  Should be fixed.  Thanks for catching.  The SpEL expression should be used instead.

markpollack:
Moving into spring-xd-modules

PR #1384, opened 2015-01-16
XD-2482: Add `initialDelay` and `TimeUnit`

garyrussell:
JIRA: https://jira.spring.io/browse/XD-2482

Support `initialDelay` and `timeUnit` properties for the `time` and `trigger` sources.

Create a new `Mixin` but the `fixedDelay` property has to remain in the existing
options metadata because of different defaults.

### DIFF
--- a/modules/source/time/config/time.xml
+++ b/modules/source/time/config/time.xml
@@ -10,7 +10,13 @@
 	<inbound-channel-adapter channel="output"
 			auto-startup="false"
 			expression="new java.text.SimpleDateFormat('${format}').format(new java.util.Date())">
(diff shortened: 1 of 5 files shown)


markfisher:
LGTM; merging

markfisher:
actually one idea: we could convert to uppercase in the setter for time unit

that way we can provide "milliseconds" instead of "MILLISECONDS"

markfisher:
I went ahead and added the toUpperCase and merged as https://github.com/spring-projects/spring-xd/commit/44cc46ce81d6424ebcf381f225905ebfacfda9fd

PR #1388, opened 2015-01-19
XD-2612 Upgrade to kryo 3.0 and refactor to use object pooling

dturanski:


spring-xd-dirt/src/main/java/org/springframework/xd/dirt/integration/bus/serializer/kryo/FileCodec.java
```diff
@@ -21,32 +21,20 @@
...
-	protected File doDeserialize(Kryo kryo, Input input) {
+	protected File doDeserialize(Kryo kryo, final Input input) {
 		File file = kryo.readObject(input, File.class);
 		return new File(file.getPath());
 	}
-
-	@Override
-	protected Kryo getKryoInstance() {
-		Kryo kryo = new Kryo();
-		kryo.register(File.class);
```
markfisher: At first I was expecting to see an implementation of the configureKryoInstance() method here, but I assume the type is no longer registered since it is inferred? Also, the StdInstantiatorStrategy no longer needs to be set?

spring-xd-dirt/src/main/java/org/springframework/xd/dirt/integration/bus/serializer/kryo/PojoCodec.java
```diff
@@ -41,12 +40,4 @@ protected Object doDeserialize(Kryo kryo, Input input, Class<? extends Object> t
 		kryo.register(type);
```
markfisher: Now that the Kryo instances are pooled and shared across multiple invocations, does it matter that we register types on-demand, in arbitrary order? The reason I ask is this comment in the Kryo#register(..) JavaDoc:

````
* Because the ID assigned is affected by the IDs registered before it, the order classes are registered is important when
* using this method. The order must be the same at deserialization as it was for serialization. */
````

spring-xd-dirt/src/main/java/org/springframework/xd/dirt/integration/bus/serializer/kryo/FileCodec.java
dturanski: Std is no longer needed. Dropping the register call wasn't intentional but perhaps not necessary.

spring-xd-dirt/src/main/java/org/springframework/xd/dirt/integration/bus/serializer/kryo/PojoCodec.java
dturanski: Good catch

markfisher:
LGTM; rebased, squashed, and merged as https://github.com/spring-projects/spring-xd/commit/f288ea9216cd7d9f97ce8fcc9354906f2425cc01

PR #1389, opened 2015-01-20
[Sprint 41] XD-2616 Read topic metadata after creation/modification

mbogoevici:
Ensures that topic metadata is created/updated correctly when a topic is created. As the call to update them will return before the creation is complete, a race condition between topic creation and producer/consumer start may occur.

### DIFF
--- a/spring-xd-messagebus-kafka/src/main/java/org/springframework/xd/dirt/integration/kafka/KafkaMessageBus.java
+++ b/spring-xd-messagebus-kafka/src/main/java/org/springframework/xd/dirt/integration/kafka/KafkaMessageBus.java
@@ -30,6 +30,8 @@ import java.util.concurrent.Executors;
 import java.util.concurrent.atomic.AtomicInteger;
 
 import kafka.admin.AdminUtils;
+import kafka.api.TopicMetadata;
+import kafka.common.ErrorMapping;
 import kafka.consumer.Consumer;
 import kafka.consumer.ConsumerConfig;
 import kafka.consumer.ConsumerIterator;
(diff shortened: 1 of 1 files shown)


spring-xd-messagebus-kafka/src/main/java/org/springframework/xd/dirt/integration/kafka/KafkaMessageBus.java
```diff
@@ -377,20 +384,50 @@ public void bindReplier(String name, MessageChannel requests, MessageChannel rep
...
+			timeoutRetryPolicy.setTimeout(METADATA_VERIFICATION_TIMEOUT);
+			SimpleRetryPolicy simpleRetryPolicy = new SimpleRetryPolicy();
+			simpleRetryPolicy.setMaxAttempts(METADATA_VERIFICATION_RETRY_ATTEMPTS);
+			policy.setPolicies(new RetryPolicy[] {timeoutRetryPolicy, simpleRetryPolicy});
 
+			try {
+				retryTemplate.execute(new RetryCallback<Boolean, Exception>() {
+					@Override
+					public Boolean doWithRetry(RetryContext context) throws Exception {
+						TopicMetadata topicMetadata = AdminUtils.fetchTopicMetadataFromZk(topicName, zkClient);
```
artembilan: I think for such an low-level resource we need some `BackOffPolicy` between each retry. By default it is `NoBackOffPolicy`. 
The `TimeoutRetryPolicy` just says how long our `retryTemplate.execute()` can live, but it doesn't help to have some `Thread.sleep` between retries.
Am I missing something here otherwise ?

markpollack: Adding in a exponential backoff would prob be ok.  I think we can expose the same config as in the RabbitMessageBus.  Looked into that during the course of the retry logic for richgauge recently.

mbogoevici: Ok, I'll make that change.

ericbottard:
Rebased, squashed and merged as 407ad8e95206e2832735e95aedd6e574b5bf09cc

PR #1390, opened 2015-01-20
[Sprint 41] XD-2358-2359-2360-2549 Kafka source using the Simple Consumer API

mbogoevici:
[Do not merge - dependency in staging]

- Upgrade to Spring Integration Kafka 1.0.0.M3
- Replace the Kafka Source with one based on the KafkaMessageDrivenChannelAdapter
- Add support for listening to a subset of the original set of partitions ('partitions' module parameter)
- Evenly split the listened partitions between module instance and coordinate via Zookeeper
- Add support for starting reading from a series of initial offsets ('initialOffsets' parameter)
- Pluggable strategy for storing offsets during consumption (supported: in-memory/transient and Redis)


spring-xd-dirt/src/main/java/org/springframework/xd/dirt/plugins/ZookeeperConnectionPlugin.java
```diff
@@ -0,0 +1,53 @@
...
+	}
+
+	@Override
+	public void preProcessModule(Module module) {
+		module.addListener(new ApplicationListener<ApplicationPreparedEvent>() {
+			@Override
+			public void onApplicationEvent(ApplicationPreparedEvent event) {
+				ZooKeeperConnection bean = getApplicationContext().getBean(ZooKeeperConnection.class);
+				ConfigurableListableBeanFactory moduleBeanFactory = event.getApplicationContext().getBeanFactory();
+				moduleBeanFactory.registerSingleton(XD_ZOOKEEPER_CONNECTION, bean.getClient());
```
artembilan: You call it like `xd.zookeeper.connection`, but ask as `bean.getClient()`.
So, maybe `xd.zookeeper.client`?

extensions/spring-xd-extension-kafka/src/main/java/org/springframework/integration/x/kafka/KafkaPartitionAllocator.java
```diff
@@ -0,0 +1,356 @@
...
+ *
+ * As a {@link FactoryBean} it will return the partitions that the current module instance should listen to.
+ * The list of partitions is stored in ZooKeeper, and is created when the module is started for the first time.
+ * If multiple module instances are started at the same time, they synchronize via ZooKeeper, and only the first
+ * will end up creating the module list, while the other instances will read it.
+ *
+ * Restarted modules will read the list of partitions that corresponds to their own sequence. Zero-count Kafka source
+ * modules are not supported.
+ *
+ * As an {@link ApplicationListener}
```
pperalta: Is the JavaDoc complete?

extensions/spring-xd-extension-kafka/src/main/java/org/springframework/integration/x/kafka/KafkaPartitionAllocator.java
```diff
@@ -0,0 +1,356 @@
...
+							client.delete().deletingChildrenIfNeeded().forPath(getDataPath());
+						}
+						catch (KeeperException.NoNodeException e) {
+							if (KeeperException.Code.NONODE.equals(e.code())) {
+								// ignore, most likely someone else has deleted the path already
+							}
+							else {
+								// this actually was an exception : we cannot recover, but cannot stop either
+								log.error(e);
+								e.printStackTrace();
```
pperalta: Why log the error and print a stacktrace? Won't this cause the stack trace to appear twice?

extensions/spring-xd-extension-kafka/src/main/java/org/springframework/integration/x/kafka/KafkaPartitionAllocator.java
```diff
@@ -0,0 +1,356 @@
...
+							}
+						}
+					}
+				}
+				else {
+					System.out.println("We cannot clean, because: " + status);
```
pperalta: Should be log messaage

build.gradle
```diff
@@ -70,7 +70,9 @@ allprojects {
 
 	repositories {
 		mavenCentral()
+		mavenLocal()
```
artembilan: What is the reason of this ?

extensions/spring-xd-extension-kafka/src/main/java/org/springframework/integration/x/kafka/KafkaPartitionAllocator.java
```diff
@@ -0,0 +1,356 @@
...
+		List<Partition> partitions = new ArrayList<Partition>();
+		for (Integer partitionId : partitionIds) {
+			partitions.add(new Partition(topic, partitionId));
+		}
+		return partitions.toArray(new Partition[partitions.size()]);
+	}
+
+	@Override
+	public Class<?> getObjectType() {
+		return Array.class;
```
artembilan: `Partition[].class` ?

extensions/spring-xd-extension-kafka/src/main/java/org/springframework/integration/x/kafka/KafkaPartitionAllocator.java
```diff
@@ -0,0 +1,356 @@
...
+						}
+					}
+				}
+				else {
+					System.out.println("We cannot clean, because: " + status);
+				}
+			}
+			catch (Exception e) {
+				log.error(e);
+				e.printStackTrace();
```
artembilan: The same `printStackTrace()` issue

extensions/spring-xd-extension-kafka/src/main/java/org/springframework/integration/x/kafka/KafkaPartitionAllocator.java
```diff
@@ -0,0 +1,356 @@
...
+	 * @return the list of integers
+	 * @throws IllegalArgumentException if the format of the list is incorrect
+	 */
+	public static Iterable<Integer> parseNumberList(String numberList) throws IllegalArgumentException {
+		Assert.hasText(numberList, "must contain a list of values");
+		// TODO: regex validate
+		Set<Integer> numbers = new TreeSet<Integer>();
+		String[] numbersOrRanges = numberList.split(",");
+		for (String numberOrRange : numbersOrRanges) {
+			if (numberOrRange.contains("-")) {
```
artembilan: Can't we use here Scala `Range` ?

extensions/spring-xd-extension-kafka/src/main/java/org/springframework/integration/x/kafka/SpringXdOffsetManager.java
```diff
@@ -0,0 +1,97 @@
...
+	public SpringXdOffsetManager(ConnectionFactory connectionFactory, String topic, String initialOffsets) {
+		super(connectionFactory,
+				StringUtils.hasText(initialOffsets) ? parseOffsetList(topic, initialOffsets) : Collections.EMPTY_MAP);
+	}
+
+	public void deleteOffset(Partition partition) {
+		getMetadataStore().remove(partition.getTopic() + ":" + partition.getId() + ":" + getConsumerId());
+	}
+
+	public void setAutoOffsetReset(AutoOffsetResetStrategy autoOffsetResetStrategy) {
```
artembilan: Out of use. `KafkaSourceModuleOptionsMetadata.setAutoOffset()` ?

modules/source/kafka/config/kafka.xml
```diff
@@ -1,59 +1,80 @@
...
-	   xmlns:int="http://www.springframework.org/schema/integration"
-	   xsi:schemaLocation="http://www.springframework.org/schema/integration http://www.springframework.org/schema/integration/spring-integration.xsd
+			 xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+			 xmlns:int-kafka="http://www.springframework.org/schema/integration/kafka"
+			 xmlns:int="http://www.springframework.org/schema/integration"
+			 xsi:schemaLocation="http://www.springframework.org/schema/integration http://www.springframework.org/schema/integration/spring-integration.xsd
 		http://www.springframework.org/schema/beans http://www.springframework.org/schema/beans/spring-beans.xsd
 		http://www.springframework.org/schema/integration/kafka http://www.springframework.org/schema/integration/kafka/spring-integration-kafka.xsd">
 
+	<import resource="${offsetStorage:inmemory}-metadata-store.xml"/>
```
artembilan: Why not `profile`?

modules/source/kafka/config/inmemory-metadata-store.xml
```diff
@@ -0,0 +1,6 @@
+<beans xmlns="http://www.springframework.org/schema/beans"
+			 xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+			 xsi:schemaLocation="http://www.springframework.org/schema/beans http://www.springframework.org/schema/beans/spring-beans.xsd">
+
+	<bean id="metadataStore" class="org.springframework.integration.metadata.SimpleMetadataStore"/>
+</beans>
```
artembilan: New line in the end of file

modules/source/kafka/config/kafka.xml
```diff
@@ -1,59 +1,80 @@
...
-					<prop key="rebalance.backoff.ms">${rebalanceBackoff}</prop>
-					<prop key="refresh.leader.backoff.ms">${refreshLeaderBackOff}</prop>
-					<prop key="auto.offset.reset">${autoOffsetReset}</prop>
-					<prop key="consumer.timeout.ms">${consumerTimeout}</prop>
-				</props>
-		</property>
+	<int-kafka:zookeeper-connect id="kafkaSourceZookeeperConnect" zk-connect="${zkconnect}"
+															 zk-connection-timeout="${zkconnectionTimeout}"
+															 zk-session-timeout="${zksessionTimeout}"
+															 zk-sync-time="${zksyncTime}"/>
```
artembilan: Please, format this better to make readable on GH

artembilan:
No tests on matter. 
Or do existing tests for Kafka source cover this well?

build.gradle
mbogoevici: Omission.

extensions/spring-xd-extension-kafka/src/main/java/org/springframework/integration/x/kafka/KafkaPartitionAllocator.java
mbogoevici: Yeah, I'll clean these up.

extensions/spring-xd-extension-kafka/src/main/java/org/springframework/integration/x/kafka/KafkaPartitionAllocator.java
mbogoevici: I'd seriously prefer avoiding mixing any Scala code anymore.

modules/source/kafka/config/kafka.xml
mbogoevici: No particular reason. I guess either could work. We do something similar for the message bus.

extensions/spring-xd-extension-kafka/src/main/java/org/springframework/integration/x/kafka/KafkaPartitionAllocator.java
mbogoevici: No :). Omission

extensions/spring-xd-extension-kafka/src/main/java/org/springframework/integration/x/kafka/KafkaPartitionAllocator.java
mbogoevici: Omission :)

modules/source/kafka/config/kafka.xml
ericbottard: Profiles are preferred for modules.

extensions/spring-xd-extension-kafka/src/main/java/org/springframework/integration/x/kafka/KafkaPartitionAllocator.java
```diff
@@ -0,0 +1,356 @@
...
+
+	private final CuratorFramework client;
+
+	private final SpringXdOffsetManager offsetManager;
+
+	private String moduleName;
+
+	private String streamName;
+
+	private ObjectMapper objectMapper = new ObjectMapper();
```
pperalta: I think this can be final too.

extensions/spring-xd-extension-kafka/src/main/java/org/springframework/integration/x/kafka/KafkaPartitionAllocator.java
```diff
@@ -0,0 +1,356 @@
...
+
+	private InterProcessMutex partitionDataMutex;
+
+	private String partitionDataPath;
+
+	private Partition[] partitions;
+
+	public KafkaPartitionAllocator(CuratorFramework client, ConnectionFactory connectionFactory,
```
pperalta: This bean combines constructor injection with setter injection - can all of the properties be supplied via constructor injection? If so, this means you can
* make `moduleName` and `streamName` final
* eagerly instantiate `partitionDataPath` and make that final

artembilan: We can't do low-level resource (IO operation) from ctor

extensions/spring-xd-extension-kafka/src/main/java/org/springframework/integration/x/kafka/KafkaPartitionAllocator.java
```diff
@@ -0,0 +1,356 @@
...
+
+	private String streamName;
+
+	private ObjectMapper objectMapper = new ObjectMapper();
+
+	private InterProcessMutex partitionDataMutex;
+
+	private String partitionDataPath;
+
+	private Partition[] partitions;
```
pperalta: `partitions` is initialized in synchronized method `getObject()` but accessed without a memory barrier in `onApplicationEvent`. In practice these methods are probably executed on the same thread; in theory there's nothing preventing different threads from executing these methods. The quickest fix would be to make `partitions` volatile.

extensions/spring-xd-extension-kafka/src/main/java/org/springframework/integration/x/kafka/KafkaPartitionAllocator.java
pperalta: Right; data cannot be read from ZK in the ctor. However I don't think that `moduleName` and `streamName` are read from ZK - or am I mistaken?

artembilan: Doh. Yes, you are right, @pperalta . I mix your comment with the `partitions` retrival stuff. 
Yes, evrything that together with `afterPropertiesSet()` can be achieved from ctor

mbogoevici: Yes, it doesn't make sense to have them as setters. Good catch.

I wouldn't create the data path in ZK and initialize the mutex in the constructor, so IMO `afterPropertiesSet` should stay as it is. Maybe use a final field for `partitionDataPath`.

mbogoevici: Hm, `partitionDataPath` is a variable and can be initialized in constructor too.

extensions/spring-xd-extension-kafka/src/main/java/org/springframework/integration/x/kafka/KafkaPartitionAllocator.java
mbogoevici: Thanks, will fix.

modules/source/kafka/config/kafka.xml
mbogoevici: Hm, not sure what IntelliJ did here, I'll fix.

extensions/spring-xd-extension-kafka/src/main/java/org/springframework/integration/x/kafka/KafkaPartitionAllocator.java
mbogoevici: Good catch, Patrick.

modules/source/kafka/config/kafka.xml
mbogoevici: well, ok. that's a minor change :)

spring-xd-dirt/src/main/java/org/springframework/xd/dirt/plugins/ZookeeperConnectionPlugin.java
mbogoevici: Okay - the original intent of this was to actually share the ZookeeperConnection, but that can't be done.

mbogoevici:
Re: testing - the core components have been tested as part of Spring Integration Kafka. There currently are is `KafkaSinkSourceTest` that exercises the source. Added https://jira.spring.io/browse/XD-2624 to cover the more comprehensive testing before RELEASE.

extensions/spring-xd-extension-kafka/src/main/java/org/springframework/integration/x/kafka/SpringXdOffsetManager.java
```diff
@@ -0,0 +1,101 @@
...
+
+package org.springframework.integration.x.kafka;
+
+import java.util.Collections;
+import java.util.HashMap;
+import java.util.Map;
+import java.util.regex.Pattern;
+
+import kafka.api.OffsetRequest;
+import org.thymeleaf.spring4.util.SpringVersionUtils;
```
artembilan: Redundant

extensions/spring-xd-extension-kafka/src/main/java/org/springframework/integration/x/kafka/SpringXdOffsetManager.java
```diff
@@ -0,0 +1,101 @@
...
+	 *
+	 * @param connectionFactory
+	 * @param initialOffsets
+	 */
+	@SuppressWarnings("unchecked")
+	public SpringXdOffsetManager(ConnectionFactory connectionFactory, String topic, String initialOffsets) {
+		super(connectionFactory,
+				StringUtils.hasText(initialOffsets) ? parseOffsetList(topic, initialOffsets) : Collections.EMPTY_MAP);
+		for (Map.Entry<Partition, Long> integerLongEntry : parseOffsetList(topic, initialOffsets).entrySet()) {
+			System.out.println(topic + ":" + integerLongEntry.getKey().getTopic() + ":" + integerLongEntry.getKey().getId() + ":" + integerLongEntry.getValue());
```
artembilan: SOUT ?

artembilan:
That's all.
@mbogoevici , be sure to comment in the PR in the future after adding commits.
Unfortunately GH doesn't notify us, when there is just only commit added...

mbogoevici:
Ok. Fixed.

markpollack:
squashed, rebased, merged as a3e40c2979a882a017c1394abcd057a73325c0b2

PR #1391, opened 2015-01-20
[Sprint: 41] [Backport 1.0.4] XD-2609 Error when listing Streams in admin-ui

ghillert:
* Fix comparison in `ZooKeeperStreamRepository#findAllInRange()`
* Add `toString()` to `BaseInstance`
* Add more clarity to Exception thrown in `XDController#enhanceWithDeployments()`

Jira: https://jira.spring.io/browse/XD-2609

### DIFF
--- a/spring-xd-dirt/src/main/java/org/springframework/xd/dirt/rest/XDController.java
+++ b/spring-xd-dirt/src/main/java/org/springframework/xd/dirt/rest/XDController.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2013 the original author or authors.
+ * Copyright 2013-2015 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -16,8 +16,10 @@
 
 package org.springframework.xd.dirt.rest;
 
+import java.util.ArrayList;
 import java.util.Collections;
 import java.util.Iterator;
+import java.util.List;
 
 import org.springframework.data.domain.Page;
 import org.springframework.data.domain.Pageable;
@@ -26,7 +28,7 @@ import org.springframework.hateoas.PagedResources;
 import org.springframework.hateoas.ResourceSupport;
 import org.springframework.hateoas.mvc.ResourceAssemblerSupport;
 import org.springframework.http.HttpStatus;
-import org.springframework.util.Assert;
+import org.springframework.util.StringUtils;
 import org.springframework.web.bind.annotation.PathVariable;
 import org.springframework.web.bind.annotation.RequestMapping;
 import org.springframework.web.bind.annotation.RequestMethod;
@@ -54,6 +56,7 @@ import org.springframework.xd.rest.domain.NamedResource;
  * @author Glenn Renfro
  * @author Ilayaperumal Gopinathan
  * @author David Turanski
+ * @author Gunnar Hillert
  */
 
 public abstract class XDController<D extends BaseDefinition, A extends ResourceAssemblerSupport<D, R>, R extends NamedResource> {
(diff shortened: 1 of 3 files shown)


spring-xd-dirt/src/main/java/org/springframework/xd/dirt/stream/zookeeper/ZooKeeperStreamRepository.java
```diff
@@ -308,7 +308,7 @@ public DeploymentUnitStatus getDeploymentStatus(String id) {
 
 		return (statusBytes == null)
 				? new DeploymentUnitStatus(DeploymentUnitStatus.State.undeployed)
-				: new DeploymentUnitStatus(ZooKeeperUtils.bytesToMap(statusBytes));
+		: new DeploymentUnitStatus(ZooKeeperUtils.bytesToMap(statusBytes));
```
pperalta: Looks like the indentation was removed?

spring-xd-dirt/src/main/java/org/springframework/xd/dirt/rest/XDController.java
```diff
@@ -204,7 +207,16 @@ private void enhanceWithDeployments(Page<D> page, PagedResources<R> result) {
...
 			}
-			Assert.state(!deployedInstances.hasNext(), "Not all instances were looked at");
+			if (deployedInstances.hasNext()) {
+				final List<String> uninspectedInstanceNames = new ArrayList<String>();
+
+				while (deployedInstances.hasNext()) {
+					final BaseInstance<D> uninspectedInstance = deployedInstances.next();
+					uninspectedInstanceNames.add(uninspectedInstance.getDefinition().getName());
+				}
+				throw new IllegalStateException("Not all instances were looked at: "
```
markpollack: isn't this going to throw whenever there are deployed instances?

ericbottard: This is really an "assertion" in the pure sense of the term, should never happen: would mean there is an "instance" without a corresponding definition

spring-xd-dirt/src/main/java/org/springframework/xd/dirt/stream/zookeeper/ZooKeeperStreamRepository.java
```diff
@@ -282,7 +282,7 @@ public void deleteAll() {
 
 		List<Stream> results = new ArrayList<Stream>();
 		for (Stream stream : all) {
-			if (stream.getDefinition().getName().compareTo(to) > 1) {
+			if (stream.getDefinition().getName().compareTo(to) > 0) {
```
ericbottard: So this is the only relevant changed line, and indeed the previous code was broken.
One point still, although it is not the subject of this PR: The `fromInclusive` and `toInclusive` parameters are not honored

spring-xd-dirt/src/main/java/org/springframework/xd/dirt/stream/zookeeper/ZooKeeperStreamRepository.java
ghillert: Stupid STS auto-formatting

spring-xd-dirt/src/main/java/org/springframework/xd/dirt/rest/XDController.java
ghillert: In general, the way the lookups are done using iterator feels odd. I chose to remove the pure Assert - in case of an error I wanted to provide some more information `Not all instances were looked at` is not that helpful.

spring-xd-dirt/src/main/java/org/springframework/xd/dirt/stream/zookeeper/ZooKeeperStreamRepository.java
ghillert: That whole class needs to be revisited. That's why I created: https://jira.spring.io/browse/XD-2620
For instance the class is untested, but it looks like adding test at this point seems tricky.

ghillert: Created https://jira.spring.io/browse/XD-2632

ghillert:
Rebased + polished

ericbottard:
Rebased, squashed and merged as
* 2000c308e9881f3115a3ad99f4bd47dc0377db26 (master)
* e8f842dd04d33c9dec8da544d91d6b3615aee097 (1.0.x)

PR #1393, opened 2015-01-21
[Sprint: 41] XD-2631 Checkin UI build artifacts

ghillert:
* Run UI build locally
* Run e2e tests
* Update `.gitignore`


ilayaperumalg:
If we are going to checkin the build artifacts then shouldn't the lib js files also need to be checked in? I don't see them here. I guess the idea here is to eliminate the explicit run of npm/bower install. If that is the case, then the lib files are missing here.
Should I expect to get all the UI resources available in spring-xd-ui jar even without running setupUI gradle task right? Let me know if I am missing something here.

ghillert:
Rebased and added lib folder

ilayaperumalg:
LGTM, merging. 
As a side note, if someone does `cleanUI` the added resources will be marked as deleted from git repo and `setupUI` will re-add them (I noticed grunt `rev` task assigns the same revision all the time - which is good for us in this case).

ilayaperumalg:
Rebased, squashed and merged as bd0d95aa450030f6bad530a1b532f790b581b666

PR #1394, opened 2015-01-21
[Sprint: 41] XD-2532 Modifying logic to access the classloader loading the Job

trisberg:


cppwfs:
LGTM.  But to verify I will test it against local, Redis, Rabbit & Kafka transports, to make sure we don't have any missing jars.

trisberg:
Looking at it again I don't think we need the last change for the AbstractProcessBuilderTasklet

extensions/spring-xd-extension-batch/src/main/java/org/springframework/batch/step/tasklet/x/AbstractProcessBuilderTasklet.java
```diff
@@ -236,23 +236,23 @@ public ExitStatus afterStep(StepExecution stepExecution) {
 	protected abstract String getCommandDescription();
 
 	protected String createClassPath(Class taskletClass) {
-		URLClassLoader thisClassLoader;
```
cppwfs: Now I'm being picky... But wouldn't this affect only the SqoopTasklet and not the Spark Tasklet?

trisberg:
Nope, it affects both, although the logic they use to obtain the classloaders are slightly different. We might have them use the same logic going forward, I'll see if I can make that happen as part of this.

cppwfs:
Gotcha.   The problem with SparkTasklet will also show up in SqoopTasklet!

PR #1396, opened 2015-01-22
[Sprint: 41] XD-2637 Added synchronized code blocks in shell processor

dturanski:
https://jira.spring.io/browse/XD-2637

### DIFF
--- a/extensions/spring-xd-extension-process/src/main/java/org/springframework/xd/extension/process/ShellCommandProcessor.java
+++ b/extensions/spring-xd-extension-process/src/main/java/org/springframework/xd/extension/process/ShellCommandProcessor.java
@@ -127,7 +127,7 @@ public class ShellCommandProcessor implements Lifecycle, InitializingBean {
 	 * Receive data from the process.
 	 * @return any available data from stdout
 	 */
-	public synchronized String receive() {
+	public String receive() {
 		Assert.isTrue(isRunning(), "Shell process is not started.");
 		synchronized (process) {
 			String data;
@@ -148,6 +148,7 @@ public class ShellCommandProcessor implements Lifecycle, InitializingBean {
 	 * @param data the data
 	 */
 	public void send(String data) {
+		Assert.isTrue(isRunning(), "Shell process is not started.");
 		synchronized (process) {
 			try {
 				serializer.serialize(data.getBytes(charset), stdin);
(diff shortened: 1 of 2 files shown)


extensions/spring-xd-extension-process/src/main/java/org/springframework/xd/extension/process/ShellCommandProcessor.java
```diff
@@ -129,15 +129,18 @@ public void start() {
...
 	public String receive() {
 		Assert.isTrue(isRunning(), "Shell process is not started.");
-		String data;
-		try {
-			byte[] buffer = serializer.deserialize(stdout);
-			data = new String(buffer, charset);
-		}
-		catch (IOException e) {
-			throw new RuntimeException(e.getMessage(), e);
+		synchronized (process) {
```
pperalta: Since each `ShellCommandProcessor` has one `process` and the `receive` & `send` are wrapped in a synchronized block (except for the `Assert`), does it make sense to simply mark the methods themselves as synchronized?

pperalta:
The `start` & `stop` methods in this class are not thread safe; if the `receive` & `send` methods are marked as synchronized then `start` & `stop` can be too.

As an alternative - if possible you could also indicate that instances of this object are not thread safe and thus should be externally protected by synchronization (similar to `java.util.HashMap`).

dturanski:
@pperalta - Good comments. WDYT about making the Process and related state ThreadLocal instead?

markpollack:
I don't see David's comment here, but did in the github email notification, `WDYT about making the Process and related state ThreadLocal instead? `

This type of approach overlaps a bit with what was done with Reactor/RxJava.  The easiest out of the box experience is for all incoming messages on the message bus to go through the same processor, so that any aggregate state is accumulated for all the data.   To do a mapping of incoming message to specific process, instead of hard-coding threadId, I used a SpEL expression so it is more flexible, this way for example you could have a python process per Kafka partition, and get some more concurrency going.  It is probably a good option to have, but a post 1.1 feature at this point.

mminella:
@markpollack I chatted with @dturanski this morning about this.  We agreed that using the `ThreadLocal` concept wouldn't work because it prevents a user from creating a stateful processor (which the connected car python processor is).  I agree that partitioning/multiple instances via regular module management means is a better option.  I'm going to be applying the updates that @pperalta mentioned and merging it this morning.

mminella:
Rebased, squashed, and merged.  Thanks!

PR #1397, opened 2015-01-22
[Sprint: 41] XD-2597 Adding new 'admininfo' command to list admins host:port info

trisberg:
- using ZooKeeper APIs directly since we only do reads

- we don't monitor for live changes since if data is not there we just show empty list

- user would have to run command again to pick up any changes

### DIFF
--- a/spring-xd-yarn/site/scripts/xd-yarn
+++ b/spring-xd-yarn/site/scripts/xd-yarn
@@ -16,7 +16,7 @@ COMMANDS:
     xd-yarn submit [OPTIONS]                 Submit a new application instance
     xd-yarn submitted [OPTIONS]              List status of submitted instances
     xd-yarn kill [OPTIONS]                   Kill an running instance
-    xd-yarn admininfo                        List http host and port info for the admin servers
+    xd-yarn admininfo [OPTIONS]              List http host and port info for the admin servers
     xd-yarn clustersinfo [OPTIONS]           List instance container clusters info
     xd-yarn clustercreate [OPTIONS]          Create a new container cluster
     xd-yarn clusterdestroy [OPTIONS]         Destroy a existing container cluster
--- a/spring-xd-yarn/site/scripts/xd-yarn
+++ b/spring-xd-yarn/site/scripts/xd-yarn
@@ -16,6 +16,7 @@ COMMANDS:
     xd-yarn submit [OPTIONS]                 Submit a new application instance
     xd-yarn submitted [OPTIONS]              List status of submitted instances
     xd-yarn kill [OPTIONS]                   Kill an running instance
+    xd-yarn admininfo                        List http host and port info for the admin servers
     xd-yarn clustersinfo [OPTIONS]           List instance container clusters info
     xd-yarn clustercreate [OPTIONS]          Create a new container cluster
     xd-yarn clusterdestroy [OPTIONS]         Destroy a existing container cluster
@@ -120,7 +121,7 @@ do
         else
             die "No application directory location specified for -d option"
         fi
-    elif [ "$1" = "push" ] || [ "$1" = "shell" ] || [ "$1" = "kill" ] || [ "$1" = "pushed" ] || [ "$1" = "submit" ] || [ "$1" = "submitted" ] || [ "$1" = "clustercreate" ] || [ "$1" = "clusterdestroy" ] || [ "$1" = "clusterinfo" ] || [ "$1" = "clustermodify" ] || [ "$1" = "clustersinfo" ] || [ "$1" = "clusterstart" ] || [ "$1" = "clusterstop" ]; then
+    elif [ "$1" = "push" ] || [ "$1" = "shell" ] || [ "$1" = "kill" ] || [ "$1" = "pushed" ] || [ "$1" = "submit" ] || [ "$1" = "submitted" ] || [ "$1" = "clustercreate" ] || [ "$1" = "clusterdestroy" ] || [ "$1" = "clusterinfo" ] || [ "$1" = "clustermodify" ] || [ "$1" = "clustersinfo" ] || [ "$1" = "clusterstart" ] || [ "$1" = "clusterstop" ] || [ "$1" = "admininfo" ]; then
         PARAMS=("${PARAMS[@]}" "$1")
         if [ "$ACTION" = "" ]; then
             ACTION="$1"
--- a/spring-xd-yarn/spring-xd-yarn-client/src/main/java/org/springframework/xd/yarn/ClientApplication.java
+++ b/spring-xd-yarn/spring-xd-yarn-client/src/main/java/org/springframework/xd/yarn/ClientApplication.java
@@ -53,6 +53,7 @@ public class ClientApplication extends AbstractCli {
 		cmdCommands.add(new YarnSubmitCommand());
 		cmdCommands.add(new YarnSubmittedCommand(new SubmittedOptionHandler("XD")));
 		cmdCommands.add(new YarnKillCommand());
+		cmdCommands.add(new XdYarnAdminInfoCommand());
 		cmdCommands.add(new YarnClustersInfoCommand());
 		cmdCommands.add(new YarnClusterInfoCommand());
 		cmdCommands.add(new XdYarnClusterCreateCommand());
@@ -62,7 +63,7 @@ public class ClientApplication extends AbstractCli {
 		cmdCommands.add(new YarnClusterDestroyCommand());		
 		app.registerCommands(cmdCommands);
 
-		// shell commands, push not supported in shell
+		// shell commands, push and admininfo not supported in shell
 		List<Command> shellCommands = new ArrayList<Command>();
 		shellCommands.add(new YarnPushedCommand());
 		shellCommands.add(new YarnSubmitCommand());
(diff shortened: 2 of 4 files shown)


spring-xd-yarn/site/scripts/xd-yarn
```diff
@@ -16,6 +16,7 @@ COMMANDS:
     xd-yarn submit [OPTIONS]                 Submit a new application instance
     xd-yarn submitted [OPTIONS]              List status of submitted instances
     xd-yarn kill [OPTIONS]                   Kill an running instance
+    xd-yarn admininfo                        List http host and port info for the admin servers
```
jvalkeal: We have options because we always have dir option for manifest directory.

spring-xd-yarn/spring-xd-yarn-client/src/main/java/org/springframework/xd/yarn/ClientApplication.java
```diff
@@ -62,7 +63,7 @@ public static void main(String... args) {
 		cmdCommands.add(new YarnClusterDestroyCommand());		
 		app.registerCommands(cmdCommands);
 
-		// shell commands, push not supported in shell
+		// shell commands, push and admininfo not supported in shell
```
jvalkeal: This didn't work with shell? I had to disable push command because xd-yarn script creates the manifest. Do we have same trouble with this?

spring-xd-yarn/spring-xd-yarn-client/src/main/java/org/springframework/xd/yarn/XdYarnAdminInfoCommand.java
```diff
@@ -0,0 +1,126 @@
...
+	}
+
+	private void close() throws InterruptedException {
+		if (zk != null) {
+			zk.close();
+		}
+	}
+
+	private List<String> getAdmins() throws KeeperException {
+		String adminsPath = "/" + zkNamespace + "/" + XD_ADMINS;
```
jvalkeal: Worth to check but don't we support slashes in namespace? If so this would probably break with double slash if zk barks of it.

spring-xd-yarn/spring-xd-yarn-client/src/main/java/org/springframework/xd/yarn/XdYarnAdminInfoCommand.java
```diff
@@ -0,0 +1,126 @@
...
+					Stat dataStat = new Stat();
+					byte[] data = zk.getData(adminsPath + "/" + admin, false, dataStat);
+					try {
+						HashMap<String,Object> dataValues =
+								new ObjectMapper().readValue(new String(data), HashMap.class);
+						results.add("http://" + dataValues.get("host") + ":" + dataValues.get("port"));
+					} catch (IOException e) {}
+				}
+			}
+		} catch (InterruptedException ignore) {}
```
pperalta: The method calling `getAdmins` throws `InterruptedException` - so why bother catching it here? I would just declare that this method throws it and let it through if it happens.

spring-xd-yarn/spring-xd-yarn-client/src/main/java/org/springframework/xd/yarn/XdYarnAdminInfoCommand.java
pperalta: AFAIK we haven't tested with slashes in the namespace. I'm not inclined to support it, but we haven't explicitly given it much thought.

spring-xd-yarn/spring-xd-yarn-client/src/main/java/org/springframework/xd/yarn/XdYarnAdminInfoCommand.java
```diff
@@ -0,0 +1,126 @@
...
+	private List<String> getAdmins() throws KeeperException {
+		String adminsPath = "/" + zkNamespace + "/" + XD_ADMINS;
+		List<String> results = new ArrayList<String>();
+		try {
+			Stat stat = zk.exists(adminsPath, false);
+			if (stat != null) {
+				List<String> admins = zk.getChildren(adminsPath, false);
+				for (String admin : admins) {
+					Stat dataStat = new Stat();
+					byte[] data = zk.getData(adminsPath + "/" + admin, false, dataStat);
```
pperalta: If you're not going to use `dataStat` I think it is OK to pass `null` here.

spring-xd-yarn/spring-xd-yarn-client/src/main/java/org/springframework/xd/yarn/XdYarnAdminInfoCommand.java
```diff
@@ -0,0 +1,126 @@
...
+					try {
+						HashMap<String,Object> dataValues =
+								new ObjectMapper().readValue(new String(data), HashMap.class);
+						results.add("http://" + dataValues.get("host") + ":" + dataValues.get("port"));
+					} catch (IOException e) {}
```
pperalta: Should we at least log something at debug level?

spring-xd-yarn/spring-xd-yarn-client/src/main/java/org/springframework/xd/yarn/XdYarnAdminInfoCommand.java
```diff
@@ -0,0 +1,126 @@
...
+		ypfb.setResources(new FileSystemResource(System.getProperty("spring.config.location")));
+		Properties conf = ypfb.getObject();
+		if (conf.containsKey(ZK_NAMESPACE)) {
+			zkNamespace = (String) conf.get(ZK_NAMESPACE);
+		}
+		if (conf.containsKey(ZK_CLIENT_CONNECT)) {
+			zkClientConnect = (String) conf.get(ZK_CLIENT_CONNECT);
+		}
+		zk = new ZooKeeper(zkClientConnect, 5000,
+				new Watcher() {
```
pperalta: I don't think you have to provide a `Watcher` if you don't need one.

jvalkeal:
Beyond the comments, just tried it and works great.

spring-xd-yarn/spring-xd-yarn-client/src/main/java/org/springframework/xd/yarn/XdYarnAdminInfoCommand.java
trisberg: I think we can expect namespace to not have slashes, don't see a reason to allow that. If there are slashes we just get an error like: 
Invalid path string "//spring-xd/admins" caused by empty node name specified @1

spring-xd-yarn/spring-xd-yarn-client/src/main/java/org/springframework/xd/yarn/XdYarnAdminInfoCommand.java
trisberg: yep, removing that catch

spring-xd-yarn/spring-xd-yarn-client/src/main/java/org/springframework/xd/yarn/XdYarnAdminInfoCommand.java
trisberg: yes, worked passing in null

spring-xd-yarn/spring-xd-yarn-client/src/main/java/org/springframework/xd/yarn/XdYarnAdminInfoCommand.java
trisberg: yes, added log statement

spring-xd-yarn/spring-xd-yarn-client/src/main/java/org/springframework/xd/yarn/XdYarnAdminInfoCommand.java
trisberg: Hmm, seems I do, get an NPE when passing in null.

spring-xd-yarn/spring-xd-yarn-client/src/main/java/org/springframework/xd/yarn/ClientApplication.java
trisberg: Didn't even try adding it. Just did though and it worked, so adding it.

spring-xd-yarn/site/scripts/xd-yarn
trisberg: ok, didn't think of that

trisberg:
pushed some changes

jvalkeal:
ok, I'll give a quick spin.

jvalkeal:
Merged

spring-xd-yarn/spring-xd-yarn-client/src/main/java/org/springframework/xd/yarn/XdYarnAdminInfoCommand.java
pperalta: Ahh OK, I see the reason why. According to the JavaDoc:

> Session establishment is asynchronous. This constructor will initiate
> connection to the server and return immediately - potentially (usually)
> before the session is fully established. The watcher argument specifies
> the watcher that will be notified of any changes in state. This
> notification can come at any point before or after the constructor call
> has returned.

So that `Watcher` can be used to determine when a session is established. I don't think you'll need it in this case; your read request should simply queue up if it is issued before the session is established.

PR #1401, opened 2015-01-26
[Sprint 41]XD-2523 implement spring-xd-module gradle plugin

dturanski:
See https://github.com/spring-projects/spring-xd/wiki/Modules#creating-a-module-project and also 
https://github.com/dturanski/spring-xd-samples/tree/XD-2523/si-dsl-module for descriptions and examples of the plugin


gradle/build-gradle-plugins.gradle
```diff
@@ -0,0 +1,44 @@
...
+		jcenter()
+	}
+	dependencies {
+		classpath("org.springframework.boot:spring-boot-gradle-plugin:1.2.1.RELEASE")
+		classpath("org.springframework.build.gradle:propdeps-plugin:0.0.7")
+	}
+}
+project('spring-xd-module-plugin') {
+	apply plugin: 'maven'
+    description = 'Gradle plugin to build a Spring XD Module'
```
ericbottard: Odd indentation, will fix on merge if that's the only issue

gradle/build-gradle-plugins.gradle
```diff
@@ -0,0 +1,44 @@
...
+	dependencies {
+ 		compile gradleApi()
+ 		compile localGroovy()
+ 		runtime "org.springframework.boot:spring-boot-gradle-plugin:1.2.1.RELEASE"
```
ericbottard: Weird that these two appear both here and on line 26.
Also, can we inherit the versions somehow?

spring-xd-gradle-plugins/spring-xd-module-plugin/src/main/groovy/org/springframework/xd/gradle/plugin/ModulePlugin.groovy
```diff
@@ -0,0 +1,106 @@
...
+	def excludeTransitiveDependencies(Project project, dependency) {
+		project.configurations.exported.exclude group: dependency.moduleGroup, module: dependency.moduleName
+
+		dependency.children.each {
+			excludeTransitiveDependencies(project, it)
+		}
+	}
+
+	/**
+	 * *
```
ericbottard: Bogus formatting

spring-xd-gradle-plugins/spring-xd-module-plugin/src/main/groovy/org/springframework/xd/gradle/plugin/ModulePlugin.groovy
```diff
@@ -0,0 +1,106 @@
...
+
+/**
+ * Gradle plugin to support module projects
+ * @author David Turanski
+ * @since 1.1
+ */
+class ModulePlugin implements Plugin<Project> {
+
+	/**
+	 * *
```
ericbottard: Bogus formatting, empty javadoc

spring-xd-gradle-plugins/spring-xd-module-plugin/src/main/groovy/org/springframework/xd/gradle/plugin/ModulePlugin.groovy
```diff
@@ -0,0 +1,106 @@
...
+ */
+
+package org.springframework.xd.gradle.plugin
+
+import org.gradle.api.Plugin
+import org.gradle.api.Project
+import org.gradle.api.tasks.Copy
+
+/**
+ * Gradle plugin to support module projects
```
ericbottard: Javadoc should/could explain what the plugin does.

spring-xd-gradle-plugins/spring-xd-module-plugin/src/main/groovy/org/springframework/xd/gradle/plugin/ModulePlugin.groovy
```diff
@@ -0,0 +1,106 @@
...
+				excludeTransitiveDependencies(project, it)
+			}
+		}
+
+		project.task('copyLocalMessageBus', type: Copy) {
+			from project.configurations.messageBus
+			into project.file('lib/messagebus/local')
+		}
+
+		project.processTestResources.dependsOn 'copyLocalMessageBus'
```
ericbottard: Could we have the IDE tasks depend on that as well (c.f. the section in the wikidoc about the required manual step)

gradle/build-gradle-plugins.gradle
dturanski: I'll look at into inheriting the versions. I think the the buildscript deps aren't technically needed, but we commonly use these in the xd build (e.g., to use provided dependencies if needed). The project deps are needed to package these plugins with the spring-xd-module plugin (i.e., they don't need to be declared in the module project.

spring-xd-gradle-plugins/spring-xd-module-plugin/src/main/groovy/org/springframework/xd/gradle/plugin/ModulePlugin.groovy
dturanski: I'm not sure how to do that

ericbottard: Same thing as above I guess: `project.idea.dependsOn 'copyLocalMessageBus'`, ditto for eclipse

dturanski: Couldn't hurt but this only works for ./gradlew idea.  It's not necessary to run this when importing the project into Intellij. Not sure if that's true for eclipse.

build.gradle
```diff
@@ -128,16 +128,28 @@ apply from: 'gradle/build-common.gradle'
 configure(javaProjects) { subproject ->
 
 	apply plugin: 'groovy'
-	if (!subproject.projectDir.absolutePath.contains('spring-xd-starters')) {
+	apply plugin: 'spring-boot'
+	apply plugin: 'eclipse'
```
ericbottard: At the very least, there is duplication here as some of those  also appear inside the if() block.
At worst, you did move those unintentionally from the if() block

markpollack:
squashed and merged as 6f50422a8c34de276f9753cac76e5ec6d5e1cfb9

PR #1405, opened 2015-01-26
XD-2643 Kafka test creates Topic prior to tests

cppwfs:
Kafka source and sink require a topic to be present before execution.  This code creates the topic at the beginning of the test.

### DIFF
--- a/spring-xd-integration-test/src/main/java/org/springframework/xd/integration/fixtures/Sources.java
+++ b/spring-xd-integration-test/src/main/java/org/springframework/xd/integration/fixtures/Sources.java
@@ -16,6 +16,8 @@
 
 package org.springframework.xd.integration.fixtures;
 
+import kafka.utils.ZKStringSerializer$;
+import org.I0Itec.zkclient.ZkClient;
 import org.springframework.util.Assert;
 import org.springframework.xd.integration.util.XdEnvironment;
 import org.springframework.xd.test.fixtures.GemFireCQSource;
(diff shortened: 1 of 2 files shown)


spring-xd-integration-test/src/main/java/org/springframework/xd/integration/fixtures/Sources.java
```diff
@@ -343,4 +345,14 @@ public TcpClientSource tcpClientSource() {
 		return new TcpClientSource(xdEnvironment.getTcpClientHost(), xdEnvironment.getTcpClientPort());
 	}
 
+	/**
+	 * Creates a ZkClient connected to the Zookeeper that the kafka is associated.
+	 * @return zkClient instance that is connected to the kafka's zookeeper
+	 */
+	public ZkClient getKafkaZkClient(){
```
markpollack: This doesn't correspond to a real source, it should be moved somewhere else... thinking of where..

spring-xd-integration-test/src/test/java/org/springframework/xd/integration/test/KafkaTests.java
```diff
@@ -35,10 +38,13 @@
...
 		String data = UUID.randomUUID().toString();
-		String topic = UUID.randomUUID().toString();
+		String topicToUse = UUID.randomUUID().toString();
+		ZkClient zkClient = sources.getKafkaZkClient();
+		AdminUtils.createTopic(zkClient, topicToUse, 1, 1, new Properties());
+
 		String sinkStreamName = "ks" + UUID.randomUUID().toString();
-		KafkaSource source = sources.kafkaSource().topic(topic).ensureReady();
-		KafkaSink sink = sinks.kafkaSink().topic(topic).ensureReady();
+		KafkaSource source = sources.kafkaSource().topic(topicToUse).ensureReady();
```
markpollack: ensureReady can make sure that the topic is created.  There is duplicated code in KafkaSource and KafkaSink in the ensureReady method, this should be moved up to a new common abstract base class.

markpollack: tried it out, because of the api chaining, prob best to refactor common code into a util class can call it from within KafkaSource/Sink, playing around some more

markpollack:
I closed this and created https://github.com/spring-projects/spring-xd/pull/1407 that you can review.

PR #1406, opened 2015-01-26
[Sprint: 41] XD-2638 Update spring-data-hadoop version to 2.1.0.RC1

trisberg:
- missed some distros for Spring Shell zip distribution in XD-2594

- adding hadoop26 and hdp22 to distShellZip task


markpollack:
LTGM, merged

PR #1414, opened 2015-01-27
Fix compiler warnings

ilayaperumalg:


spring-xd-spark-streaming/src/main/java/org/springframework/xd/spark/streaming/SparkStreamingModuleExecutor.java
```diff
@@ -40,7 +40,7 @@
  * @author Ilayaperumal Gopinathan
  * @author Mark Fisher
  */
-@SuppressWarnings("unchecked")
+@SuppressWarnings({"unchecked", "rawtypes", "serial"})
 public class SparkStreamingModuleExecutor implements Serializable {
```
markfisher: should we generate a serial ID here?

ilayaperumalg: Yeah, that makes sense. let me try adding the serialVersionUID here.

markfisher:
LGTM; merged

PR #1416, opened 2015-01-28
XD-2655 Configure deployment timeout

ilayaperumalg:
- Add `xd.admin.deploymentTimeout` property which can be
configured via servers.yml
 - Make single instance of ModuleDeploymentWriter for the admin JVM
 - Refactor `ModuleDeploymentWriter`
    - remove container matcher dependency as it is no longer needed
in its constructor
    - always set deployment timeout

### DIFF
--- a/config/servers.yml
+++ b/config/servers.yml
@@ -290,9 +290,16 @@ endpoints:
 #
 # The default value for this property is 15000 ms.
 #
-# xd:
+#xd:
 #  admin:
 #    quietPeriod: 15000
+---
+# Module deployment timeout
+# The module deployment timeout is the amount of time the Module deployment writer
+# waits to get the deployment status after writing the ZK deployment requests' paths
+#xd:
+#  admin:
+#    deploymentTimeout: 30000
 ---
 # User Extensions: Where XD scans the classpath to discover extended container configuration to add beans to the Plugins context.
 # Each property may be a comma delimited string. 'basepackages' refers to package names used for
--- a/config/servers.yml
+++ b/config/servers.yml
@@ -295,8 +295,8 @@ endpoints:
 #    quietPeriod: 15000
 ---
 # Module deployment timeout
-# The module deployment timeout is the amount of time the Module deployment writer
-# waits to get the deployment status after writing the ZK deployment requests' paths
+# The module deployment timeout is the number of milliseconds the admin server
+# will wait for a response to a module deployment request to a container.
 #xd:
 #  admin:
 #    deploymentTimeout: 30000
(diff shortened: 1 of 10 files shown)


config/servers.yml
```diff
@@ -290,10 +290,17 @@ endpoints:
 #
 # The default value for this property is 15000 ms.
 #
-# xd:
+#xd:
 #  admin:
 #    quietPeriod: 15000
 ---
+# Module deployment timeout
+# The module deployment timeout is the amount of time the Module deployment writer
```
pperalta: What do you think about this wording?

> The module deployment timeout is the number of milliseconds the admin server will wait for a response to a module deployment request to a container. 

I think we want to avoid going too deep into the implementation details - for instance, the end user doesn't know/care about ModuleDeploymentWriter, ZK, or paths.

ilayaperumalg: +1

spring-xd-dirt/src/main/java/org/springframework/xd/dirt/server/ModuleDeploymentWriter.java
```diff
@@ -504,7 +482,7 @@ public synchronized void addResult(ModuleDeploymentStatus deploymentStatus) {
 								key.moduleDescriptorKey,
 								ModuleDeploymentStatus.State.failed,
 								String.format("Deployment of module '%s' to container '%s' timed out after %d ms",
-										key.moduleDescriptorKey, key.container, timeout)));
+										key.moduleDescriptorKey, key.container, deploymentTimeout)));
```
pperalta: Should be `deploymentTimeout.get()` because the string format is expecting a number.

spring-xd-dirt/src/main/java/org/springframework/xd/dirt/server/ModuleDeploymentWriter.java
```diff
@@ -86,32 +77,19 @@
...
-	 * Construct a {@code ModuleDeploymentWriter} with the default timeout
-	 * value indicated by {@link #DEFAULT_TIMEOUT}.
 	 *
-	 * @param zkConnection         ZooKeeper connection
-	 * @param containerMatcher     matcher for modules to containers
 	 */
-	public ModuleDeploymentWriter(ZooKeeperConnection zkConnection, ContainerMatcher containerMatcher) {
-		this(zkConnection, containerMatcher, DEFAULT_TIMEOUT);
-	}
+	private final AtomicLong deploymentTimeout;
```
pperalta: Do we expect this value to change during runtime? If not, I think using `final long` instead of `AtomicLong` should suffice here.

pperalta:
Merged as 61347f2975d5680546eba2ef00e8fbd95c956a20

PR #1422, opened 2015-01-30
[Sprint 42] XD-2672 Scala support for spark streaming module

ilayaperumalg:
- Add `ScalaProcessor` trait that defines the `process` method
  - Handle the scala module that extends this trait
     - Update `SparkStreamingPlugin` and `SparkStreamingDriverModule`
to handle this appropriately
- Add examples that extend this trait
- Add tests

### DIFF
--- a/spring-xd-dirt/src/main/java/org/springframework/xd/dirt/plugins/spark/streaming/SparkStreamingPlugin.java
+++ b/spring-xd-dirt/src/main/java/org/springframework/xd/dirt/plugins/spark/streaming/SparkStreamingPlugin.java
@@ -24,6 +24,7 @@ import java.util.Properties;
 
 import org.apache.spark.storage.StorageLevel;
 
+import org.springframework.beans.factory.NoSuchBeanDefinitionException;
 import org.springframework.beans.factory.annotation.Autowired;
 import org.springframework.beans.factory.config.ConfigurableBeanFactory;
 import org.springframework.context.ConfigurableApplicationContext;
@@ -40,7 +41,7 @@ import org.springframework.xd.module.ModuleType;
 import org.springframework.xd.module.core.Module;
 import org.springframework.xd.module.core.ModuleFactory;
 import org.springframework.xd.module.spark.streaming.SparkStreamingDriverModule;
-import org.springframework.xd.spark.streaming.Processor;
+import org.springframework.xd.spark.streaming.SparkStreamingSupport;
 
 /**
  * Plugin for Spark Streaming support. This plugin sets up the necessary beans for the spark streaming module
@@ -62,7 +63,7 @@ public class SparkStreamingPlugin extends AbstractStreamPlugin {
 	@Override
 	public boolean supports(Module module) {
 		String moduleExecutionFramework = module.getProperties().getProperty(ModuleFactory.MODULE_EXECUTION_FRAMEWORK_KEY);
-		return (Processor.MODULE_EXECUTION_FRAMEWORK.equals(moduleExecutionFramework));
+		return (SparkStreamingSupport.MODULE_EXECUTION_FRAMEWORK.equals(moduleExecutionFramework));
 	}
 
 	@Override
@@ -73,19 +74,28 @@ public class SparkStreamingPlugin extends AbstractStreamPlugin {
 		Properties messageBusProperties = getMessageBusProperties(module, transport);
 		Properties inboundModuleProperties = this.extractConsumerProducerProperties(module)[0];
 		Properties outboundModuleProperties = this.extractConsumerProducerProperties(module)[1];
-		String defaultStorageLevel = env.getProperty(Processor.SPARK_STORAGE_LEVEL_PROP);
+		String defaultStorageLevel = env.getProperty(SparkStreamingSupport.SPARK_STORAGE_LEVEL_PROP);
 		StorageLevel configuredStorageLevel = StorageLevel.fromString(StringUtils.hasText(defaultStorageLevel) ?
 				defaultStorageLevel : DEFAULT_STORAGE_LEVEL);
-		String storageLevelFromModule = module.getProperties().getProperty(Processor.SPARK_STORAGE_LEVEL_MODULE_OPTION);
+		String storageLevelFromModule = module.getProperties().getProperty(SparkStreamingSupport.SPARK_STORAGE_LEVEL_MODULE_OPTION);
 		StorageLevel storageLevel = StringUtils.hasText(storageLevelFromModule) ?
 				StorageLevel.fromString(storageLevelFromModule) : configuredStorageLevel;
 		MessageBusReceiver receiver = null;
 		if (transport.equals("local")) {
-			final Processor sparkModule = module.getComponent(Processor.class);
-			Properties sparkConfigs = SparkStreamingDriverModule.getSparkModuleProperties(sparkModule);
-			String sparkMasterUrl = env.getProperty(Processor.SPARK_MASTER_URL_PROP);
-			if (sparkConfigs != null && StringUtils.hasText(sparkConfigs.getProperty(Processor.SPARK_MASTER_URL_PROP))) {
-				sparkMasterUrl = sparkConfigs.getProperty(Processor.SPARK_MASTER_URL_PROP);
+			SparkStreamingSupport processor;
+			Properties sparkConfigs = null;
+			try {
+				processor = module.getComponent(SparkStreamingSupport.class);
+				if (processor != null) {
+					sparkConfigs = SparkStreamingDriverModule.getSparkModuleProperties(processor);
+				}
+			}
+			catch (NoSuchBeanDefinitionException e) {
+				throw new RuntimeException("Either java or scala module should be present.");
+			}
+			String sparkMasterUrl = env.getProperty(SparkStreamingSupport.SPARK_MASTER_URL_PROP);
+			if (sparkConfigs != null && StringUtils.hasText(sparkConfigs.getProperty(SparkStreamingSupport.SPARK_MASTER_URL_PROP))) {
+				sparkMasterUrl = sparkConfigs.getProperty(SparkStreamingSupport.SPARK_MASTER_URL_PROP);
 			}
 			Assert.notNull(sparkMasterUrl, "Spark Master URL must be set.");
 			if (!sparkMasterUrl.startsWith("local")) {
--- a/spring-xd-dirt/src/main/java/org/springframework/xd/dirt/plugins/spark/streaming/SparkStreamingPlugin.java
+++ b/spring-xd-dirt/src/main/java/org/springframework/xd/dirt/plugins/spark/streaming/SparkStreamingPlugin.java
@@ -53,8 +53,6 @@ import org.springframework.xd.spark.streaming.SparkStreamingSupport;
 @SuppressWarnings("rawtypes")
 public class SparkStreamingPlugin extends AbstractStreamPlugin {
 
-	private static final String DEFAULT_STORAGE_LEVEL = "MEMORY_ONLY";
-
 	@Autowired
 	public SparkStreamingPlugin(MessageBus messageBus, ZooKeeperConnection zkConnection) {
 		super(messageBus, zkConnection);
@@ -76,7 +74,7 @@ public class SparkStreamingPlugin extends AbstractStreamPlugin {
 		Properties outboundModuleProperties = this.extractConsumerProducerProperties(module)[1];
 		String defaultStorageLevel = env.getProperty(SparkStreamingSupport.SPARK_STORAGE_LEVEL_PROP);
 		StorageLevel configuredStorageLevel = StorageLevel.fromString(StringUtils.hasText(defaultStorageLevel) ?
-				defaultStorageLevel : DEFAULT_STORAGE_LEVEL);
+				defaultStorageLevel : SparkStreamingSupport.SPARK_DEFAULT_STORAGE_LEVEL);
 		String storageLevelFromModule = module.getProperties().getProperty(SparkStreamingSupport.SPARK_STORAGE_LEVEL_MODULE_OPTION);
 		StorageLevel storageLevel = StringUtils.hasText(storageLevelFromModule) ?
 				StorageLevel.fromString(storageLevelFromModule) : configuredStorageLevel;
@@ -90,7 +88,7 @@ public class SparkStreamingPlugin extends AbstractStreamPlugin {
 				sparkConfigs = SparkStreamingDriverModule.getSparkModuleProperties(processor);
 			}
 			catch (NoSuchBeanDefinitionException e) {
-				throw new RuntimeException("Either java or scala module should be present.");
+				throw new IllegalStateException("Either java or scala module should be present.");
 			}
 			String sparkMasterUrl = env.getProperty(SparkStreamingSupport.SPARK_MASTER_URL_PROP);
 			if (sparkConfigs != null && StringUtils.hasText(sparkConfigs.getProperty(SparkStreamingSupport.SPARK_MASTER_URL_PROP))) {
--- a/spring-xd-dirt/src/main/java/org/springframework/xd/dirt/plugins/spark/streaming/SparkStreamingPlugin.java
+++ b/spring-xd-dirt/src/main/java/org/springframework/xd/dirt/plugins/spark/streaming/SparkStreamingPlugin.java
@@ -86,12 +86,8 @@ public class SparkStreamingPlugin extends AbstractStreamPlugin {
 			Properties sparkConfigs = null;
 			try {
 				processor = module.getComponent(SparkStreamingSupport.class);
-				if (processor != null) {
-					sparkConfigs = SparkStreamingDriverModule.getSparkModuleProperties(processor);
-				}
-				else {
-					throw new RuntimeException("Either java or scala module should be present.");
-				}
+				Assert.notNull(processor, "Problem getting the spark streaming module. Is the module context active?");
+				sparkConfigs = SparkStreamingDriverModule.getSparkModuleProperties(processor);
 			}
 			catch (NoSuchBeanDefinitionException e) {
 				throw new RuntimeException("Either java or scala module should be present.");
--- a/spring-xd-dirt/src/main/java/org/springframework/xd/dirt/plugins/spark/streaming/SparkStreamingPlugin.java
+++ b/spring-xd-dirt/src/main/java/org/springframework/xd/dirt/plugins/spark/streaming/SparkStreamingPlugin.java
@@ -89,6 +89,9 @@ public class SparkStreamingPlugin extends AbstractStreamPlugin {
 				if (processor != null) {
 					sparkConfigs = SparkStreamingDriverModule.getSparkModuleProperties(processor);
 				}
+				else {
+					throw new RuntimeException("Either java or scala module should be present.");
+				}
 			}
 			catch (NoSuchBeanDefinitionException e) {
 				throw new RuntimeException("Either java or scala module should be present.");
--- a/spring-xd-module/src/main/java/org/springframework/xd/module/spark/streaming/SparkStreamingDriverModule.java
+++ b/spring-xd-module/src/main/java/org/springframework/xd/module/spark/streaming/SparkStreamingDriverModule.java
@@ -291,25 +291,25 @@ public class SparkStreamingDriverModule extends ResourceConfiguredModule {
 			//if (file.endsWith(".jar")) {
 			if (file.endsWith(".jar") && (// Add spark jars
 					file.contains("spark") ||
-							// Add SpringXD dependencies
-							file.contains("spring-xd-") ||
-							// Add Spring dependencies
-							file.contains("spring-core") ||
-							file.contains("spring-integration-core") ||
-							file.contains("spring-beans") ||
-							file.contains("spring-context") ||
-							file.contains("spring-boot") ||
-							file.contains("spring-aop") ||
-							file.contains("spring-expression") ||
-							file.contains("spring-messaging") ||
-							file.contains("spring-retry") ||
-							file.contains("spring-tx") ||
-							file.contains("spring-data-commons") ||
-							file.contains("spring-data-redis") ||
-							file.contains("commons-pool") ||
-							file.contains("jedis") ||
-							// Add codec dependency
-							file.contains("kryo"))) {
+					// Add SpringXD dependencies
+					file.contains("spring-xd-") ||
+					// Add Spring dependencies
+					file.contains("spring-core") ||
+					file.contains("spring-integration-core") ||
+					file.contains("spring-beans") ||
+					file.contains("spring-context") ||
+					file.contains("spring-boot") ||
+					file.contains("spring-aop") ||
+					file.contains("spring-expression") ||
+					file.contains("spring-messaging") ||
+					file.contains("spring-retry") ||
+					file.contains("spring-tx") ||
+					file.contains("spring-data-commons") ||
+					file.contains("spring-data-redis") ||
+					file.contains("commons-pool") ||
+					file.contains("jedis") ||
+					// Add codec dependency
+					file.contains("kryo"))) {
 				jars.add(file);
 			}
 		}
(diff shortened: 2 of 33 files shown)


spring-xd-dirt/src/main/java/org/springframework/xd/dirt/plugins/spark/streaming/SparkStreamingPlugin.java
```diff
@@ -73,19 +74,28 @@ public void postProcessModule(Module module) {
...
-			final Processor sparkModule = module.getComponent(Processor.class);
-			Properties sparkConfigs = SparkStreamingDriverModule.getSparkModuleProperties(sparkModule);
-			String sparkMasterUrl = env.getProperty(Processor.SPARK_MASTER_URL_PROP);
-			if (sparkConfigs != null && StringUtils.hasText(sparkConfigs.getProperty(Processor.SPARK_MASTER_URL_PROP))) {
-				sparkMasterUrl = sparkConfigs.getProperty(Processor.SPARK_MASTER_URL_PROP);
+			SparkStreamingSupport processor;
+			Properties sparkConfigs = null;
+			try {
+				processor = module.getComponent(SparkStreamingSupport.class);
+				if (processor != null) {
```
ericbottard: Not sure I understand the role of this null check vs. the catch below

spring-xd-module/src/main/java/org/springframework/xd/module/spark/streaming/SparkStreamingDriverModule.java
```diff
@@ -243,25 +291,25 @@ public static Properties getSparkModuleProperties(Processor processor) {
...
-					file.contains("spring-messaging") ||
-					file.contains("spring-retry") ||
-					file.contains("spring-tx") ||
-					file.contains("spring-data-commons") ||
-					file.contains("spring-data-redis") ||
-					file.contains("commons-pool") ||
-					file.contains("jedis") ||
-					// Add codec dependency
-					file.contains("kryo"))) {
+							// Add SpringXD dependencies
```
ericbottard: weird format change

spring-xd-spark-streaming/src/main/java/org/springframework/xd/spark/streaming/java/Processor.java
```diff
@@ -0,0 +1,38 @@
...
+ */
+
+package org.springframework.xd.spark.streaming.java;
+
+import org.apache.spark.streaming.api.java.JavaDStreamLike;
+
+import org.springframework.xd.spark.streaming.SparkStreamingSupport;
+
+/**
+ * Interface for modules using the Spark Streaming API to process messages from the Message Bus.
```
ericbottard: I think the word "Java" (as opposed to "Scala") should appear here

spring-xd-spark-streaming/src/main/scala/org/springframework/xd/spark/streaming/examples/SparkScalaLog.scala
```diff
@@ -0,0 +1,75 @@
...
+  }
+
+  @SparkConfig def getSparkConfigProperties: Properties = {
+    val props: Properties = new Properties
+    props.setProperty("spark.master", "local[4]")
+    return props
+  }
+
+  def process(input: ReceiverInputDStream[String]): DStream[String] = {
+    if (input != null) {
```
ericbottard: Should the user really need to protect against null here?

spring-xd-spark-streaming/src/main/scala/org/springframework/xd/spark/streaming/examples/SparkScalaLog.scala
```diff
@@ -0,0 +1,75 @@
...
+      input.foreachRDD(rdd => {
+        rdd.foreachPartition(partition => {
+          try {
+            val fw: FileWriter = new FileWriter(file.getAbsoluteFile)
+            val bw: BufferedWriter = new BufferedWriter(fw)
+            while (partition.hasNext) {
+              bw.append(partition.next.toString + System.lineSeparator)
+            }
+            bw.close
```
ericbottard: Shall we bother moving to finally block?

spring-xd-spark-streaming/src/main/scala/org/springframework/xd/spark/streaming/scala/Processor.scala
```diff
@@ -0,0 +1,36 @@
...
+ * limitations under the License.
+ */
+package org.springframework.xd.spark.streaming.scala
+
+import org.apache.spark.streaming.dstream.{DStream, ReceiverInputDStream}
+import org.springframework.xd.spark.streaming.SparkStreamingSupport
+
+/**
+ * Trait for scala modules using the Spark Streaming API to process messages from the Message Bus.
+ *
```
ericbottard: A and B should be documented with `@tparam`. Arguably, I and O would be better names

spring-xd-spark-streaming/src/main/scala/org/springframework/xd/spark/streaming/scala/Processor.scala
```diff
@@ -0,0 +1,36 @@
...
+ * @author Ilayaperumal Gopinathan
+ */
+trait Processor[A, B] extends SparkStreamingSupport {
```
ericbottard: Variance indicators should be set here, namely -A +B if I'm not mistaken

ericbottard:
I've not gone into the details of this PR at all, but overall, seems like it's calling for the Strategy pattern, what do you think?

spring-xd-dirt/src/main/java/org/springframework/xd/dirt/plugins/spark/streaming/SparkStreamingPlugin.java
ilayaperumalg: That was an extra check to protect the returned value (from SimpleModule's getComponent) against null. I will add additional else block to throw an exception if this is null. does that make sense?

spring-xd-spark-streaming/src/main/scala/org/springframework/xd/spark/streaming/scala/Processor.scala
ilayaperumalg: sorry, as a scala primer I don't understand what's needed here. Could you help understand? Thanks!

ericbottard: Variance indicators tell the compiler how sub/super types of another type relate wrt their generics.
If you have a A = Processor[X, Animal] which return what is deemed to be an Animal, then B = Processor[X, Goose] is indeed a subtype of A (because it can be used anywhere an A was used, as any Goose is guaranteed to be an Animal). Goes the other way around wrt the input type. http://en.wikipedia.org/wiki/Covariance_and_contravariance_%28computer_science%29 has a good explanation

spring-xd-dirt/src/main/java/org/springframework/xd/dirt/plugins/spark/streaming/SparkStreamingPlugin.java
ericbottard: I was referring to the `catch (NoSuchBeanDefinitionException e) {` block.
Not sure how `getComponent()` behaves now that I think of it, but in any case, there should be only one of [null check, catch NoSuchBeanDefinitionException] IMO

spring-xd-spark-streaming/src/main/scala/org/springframework/xd/spark/streaming/scala/Processor.scala
ilayaperumalg: ok, thanks Eric! I don't think variance indicators matter here as A and B can be of any type based on the computation logic defined by the developer while implementing the process method. what do you think?

ericbottard: They don't "matter" from a practical standpoint I believe, yes. But if you want to be 100% correct, I believe they can go in

spring-xd-dirt/src/main/java/org/springframework/xd/dirt/plugins/spark/streaming/SparkStreamingPlugin.java
ilayaperumalg: The extra null check I added because `getComponent` could return null while it could throw `NoSuchBeanDefinitionException`

markfisher: This is the method in question:

````
public <T> T getComponent(Class<T> requiredType) {
   return (this.context.isActive()) ? this.context.getBean(requiredType) : null;
}
````

So as @ilayaperumalg mentions, it can return *null* (whenever the context is not active), or if the context is active but no bean of that type exists, it can throw `NoSuchBeanDefinitionException`.

My question would be if that underlying context would ever be inactive at this point? And if so (thus *null* returned), should we handle that differently, minimally with a different error message?

**Comment by markfisher** [comment]
I agree with @ericbottard that the parameterized types for the Scala Processor would be clearer if I and O instead of A and B, respectively.

File: spring-xd-module/src/main/java/org/springframework/xd/module/spark/streaming/SparkStreamingDriverModule.java [hunk 8]
```diff
@@ -174,14 +186,50 @@ public void run() {
...
+	 *
+	 * @param masterURL the spark cluster master URL
+	 * @param sparkConfigs the spark configuration properties
+	 * @return SparkConf for this spark streaming module
+	 */
+	private SparkConf setupSparkConf(String masterURL, Properties sparkConfigs) {
+		SparkConf sparkConf = new SparkConf()
+				// Set spark UI port to random available port to support multiple spark modules on the same host.
+				.set("spark.ui.port", String.valueOf(SocketUtils.findAvailableTcpPort()))
+						// Set the cores max so that multiple (at least a few) spark modules can be deployed on the same host.
```
markfisher: excessive indent on this comment

spring-xd-module/src/main/java/org/springframework/xd/module/spark/streaming/SparkStreamingDriverModule.java
```diff
@@ -174,14 +186,50 @@ public void run() {
...
+	/**
+	 * Retrieve spark configuration properties from the {@link org.springframework.xd.spark.streaming.java.Processor} implementation.
 	 * This method uses {@link SparkConfig} annotation to derive the {@link Properties} returned
 	 * from the annotated methods.
 	 *
 	 * @param processor the spark streaming processor
 	 * @return the spark configuration properties (if defined) or empty properties
 	 */
-	public static Properties getSparkModuleProperties(Processor processor) {
+	public static Properties getSparkModuleProperties(Object processor) {
```
markfisher: the param type could be SparkStreamingSupport?

spring-xd-spark-streaming/src/main/java/org/springframework/xd/spark/streaming/DefaultSparkStreamingModuleOptionsMetadata.java
```diff
@@ -23,13 +23,13 @@
 import org.springframework.xd.module.options.spi.ModuleOption;
 
 /**
- * Default module options for {@link Processor}.
+ * Default module options for {@link org.springframework.xd.spark.streaming.java.Processor}.
```
markfisher: should probably link to the base type here (and mention that is the common parent for both java and scala Processor instances)

spring-xd-dirt/src/main/java/org/springframework/xd/dirt/plugins/spark/streaming/SparkStreamingPlugin.java
```diff
@@ -73,19 +74,27 @@ public void postProcessModule(Module module) {
...
-			if (sparkConfigs != null && StringUtils.hasText(sparkConfigs.getProperty(Processor.SPARK_MASTER_URL_PROP))) {
-				sparkMasterUrl = sparkConfigs.getProperty(Processor.SPARK_MASTER_URL_PROP);
+			SparkStreamingSupport processor;
+			Properties sparkConfigs = null;
+			try {
+				processor = module.getComponent(SparkStreamingSupport.class);
+				Assert.notNull(processor, "Problem getting the spark streaming module. Is the module context active?");
+				sparkConfigs = SparkStreamingDriverModule.getSparkModuleProperties(processor);
+			}
+			catch (NoSuchBeanDefinitionException e) {
```
markfisher: This should probably throw something more specific than RuntimeException, perhaps IllegalStateException.

spring-xd-module/src/main/java/org/springframework/xd/module/spark/streaming/SparkStreamingDriverModule.java
```diff
@@ -106,51 +113,56 @@ public SparkStreamingDriverModule(ModuleDescriptor descriptor, ModuleDeploymentP
...
-		List<String> sparkJars = new ArrayList<String>();
-		// Add jars from spark.jars (if any) set from spark module.
+		SparkStreamingSupport processor;
+		Properties sparkConfigs = null;
 		try {
-			String jarsFromConf = sparkConf.get("spark.jars");
-			if (StringUtils.hasText(jarsFromConf)) {
-				sparkJars.addAll(Arrays.asList(jarsFromConf.split("\\s*,\\s*")));
+			processor = getComponent(SparkStreamingSupport.class);
+			if (processor != null) {
```
markfisher: Do we need to Assert.notNull here as well?

spring-xd-module/src/main/java/org/springframework/xd/module/spark/streaming/SparkStreamingDriverModule.java
```diff
@@ -56,13 +59,14 @@
...
+import org.springframework.xd.spark.streaming.java.Processor;
 import org.springframework.xd.spark.streaming.SparkConfig;
 import org.springframework.xd.spark.streaming.SparkMessageSender;
-import org.springframework.xd.spark.streaming.SparkStreamingModuleExecutor;
+import org.springframework.xd.spark.streaming.java.SparkStreamingModuleExecutor;
+import org.springframework.xd.spark.streaming.SparkStreamingSupport;
 
 /**
- * The driver that adapts an implementation of {@link org.springframework.xd.spark.streaming.Processor} to be executed as an XD module.
+ * The driver that adapts an implementation of {@link org.springframework.xd.spark.streaming.java.Processor} to be executed as an XD module.
```
markfisher: should mention that this adapts either the java or scala Processor implementation (not just java)

spring-xd-module/src/main/java/org/springframework/xd/module/spark/streaming/SparkStreamingDriverModule.java
```diff
@@ -106,51 +113,56 @@ public SparkStreamingDriverModule(ModuleDescriptor descriptor, ModuleDeploymentP
...
+	 */
+	private void startSparkStreamingContext(Properties sparkConfigs,
+			final SparkStreamingSupport sparkStreamingSupport) {
+		final Receiver receiver = getComponent(Receiver.class);
+		Environment env = this.getApplicationContext().getEnvironment();
+		String masterURL = env.getProperty(SparkStreamingSupport.SPARK_MASTER_URL_PROP,
+				SparkStreamingSupport.SPARK_DEFAULT_MASTER_URL);
+		final SparkConf sparkConf = setupSparkConf(masterURL, sparkConfigs);
+		final String batchInterval = env.getProperty("batchInterval",
+				env.getProperty("spark.streaming.batchInterval", SPARK_STREAMING_BATCH_INTERVAL));
```
markfisher: should we a constant (to SparkStreamingSupport) for the key: "spark.streaming.batchInterval"?

spring-xd-spark-streaming/src/main/java/org/springframework/xd/spark/streaming/java/SparkStreamingModuleExecutor.java
```diff
@@ -45,13 +39,8 @@
...
 	private static SparkMessageSender messageSender;
 
-	private static final String HEADER_PREFIX = "spark.";
-
-	// todo: allow this to be overridden for any Spark Streaming Processor
-	private final MessageConverter converter = new DefaultSparkStreamingMessageConverter();
-
 	@SuppressWarnings("rawtypes")
-	public void execute(JavaDStream input, Processor processor, final SparkMessageSender sender) {
+	public void execute(JavaDStreamLike input, Processor processor, final SparkMessageSender sender) {
```
markfisher: Can this and the scala version implement a common interface? If so, it should be parameterized for *&lt;P extends Processor&gt;*.

markfisher:
we need to add this to .gitignore (it appears after running a full 'clean build'):

````
spring-xd-spark-streaming/.cache
````

**Comment by markfisher** [comment]
Queues created in Rabbit are not cleaned up properly, e.g.:

````
xdbus.SparkLogModuleTest958666173.0	0
xdbus.SparkProcessorModuleTest-1773612481.0	0
xdbus.SparkProcessorModuleTest-1773612481.1	0
xdbus.SparkScalaLogModuleTest-672629880.0	0
xdbus.SparkScalaProcessorModuleTest56289912.0	0
xdbus.SparkScalaProcessorModuleTest56289912.1	0
````

File: spring-xd-spark-streaming/src/main/java/org/springframework/xd/spark/streaming/java/SparkStreamingModuleExecutor.java (hunk 15 above)
Comment (ilayaperumalg) [inline spring-xd-spark-streaming/src/main/java/org/springframework/xd/spark/streaming/java/SparkStreamingModuleExecutor.java]: It is tricky to have a common interface as both this and scala versions have different input arg. what do you think?

Comment (markfisher) [inline spring-xd-spark-streaming/src/main/java/org/springframework/xd/spark/streaming/java/SparkStreamingModuleExecutor.java]: I was thinking of something like this (but it may not be applicable):

````
public interface SparkStreamingModuleExecutor<I, P extends SparkStreamingSupport> {
    void execute(I input, P processor, SparkMessageSender sender);
}
````

Comment (markfisher) [inline spring-xd-spark-streaming/src/main/java/org/springframework/xd/spark/streaming/java/SparkStreamingModuleExecutor.java]: It would still require casting where it's used, so it's probably not worth the effort anyways. So, nevermind.

File: spring-xd-module/src/main/java/org/springframework/xd/module/spark/streaming/SparkStreamingDriverModule.java [hunk 16]
```diff
@@ -174,14 +186,50 @@ public void run() {
...
+	 * @return SparkConf for this spark streaming module
+	 */
+	private SparkConf setupSparkConf(String masterURL, Properties sparkConfigs) {
+		SparkConf sparkConf = new SparkConf()
+				// Set spark UI port to random available port to support multiple spark modules on the same host.
+				.set("spark.ui.port", String.valueOf(SocketUtils.findAvailableTcpPort()))
+				// Set the cores max so that multiple (at least a few) spark modules can be deployed on the same host.
+				.set("spark.cores.max", "3")
+				.setMaster(masterURL)
+				.setAppName(getDescriptor().getGroup() + "-" + getDescriptor().getModuleLabel());
```
markfisher: how does the appname work when we have module count > 1? (they have the same app name, but IIRC they also have a unique app ID?)

spring-xd-spark-streaming/src/main/java/org/springframework/xd/spark/streaming/java/SparkStreamingModuleExecutor.java
ilayaperumalg: No, you are right. It works this way.

spring-xd-module/src/main/java/org/springframework/xd/module/spark/streaming/SparkStreamingDriverModule.java
```diff
@@ -273,8 +321,14 @@ public void stop() {
...
 			// todo: when possible (spark 1.3.0), change to streamingContext.stop(false, true) without the cancel
-			streamingContext.ssc().sc().cancelAllJobs();
-			streamingContext.close();
+			if (javaStreamingContext != null) {
+				javaStreamingContext.ssc().sc().cancelAllJobs();
+				javaStreamingContext.close();
+			}
+			if (streamingContext != null) {
+				streamingContext.sc().cancelAllJobs();
+				streamingContext.stop(true, false);
```
markfisher: is this what we want here? The first boolean indicates stopping the actual SparkContext and the 2nd boolean is FALSE for "graceful"
So, two things jump out:
1) those boolean values are the reverse of what we mention in the TODO above this
2) this is different than the Java version (where we just call close), and that was done that way - along with the cancel call - as a temporary workaround until the proper stop (with false, true) would work

spring-xd-spark-streaming/src/main/java/org/springframework/xd/spark/streaming/java/Processor.java
```diff
@@ -0,0 +1,38 @@
...
+
+import org.springframework.xd.spark.streaming.SparkStreamingSupport;
+
+/**
+ * Interface for the Java based modules using the Spark Streaming API to process messages from the Message Bus.
+ *
+ * @author Mark Fisher
+ */
+@SuppressWarnings("rawtypes")
+public interface Processor extends SparkStreamingSupport {
```
markfisher: should we parameterize this interface? (I/O extends JavaDStreamLike)

ilayaperumalg: +1

spring-xd-module/src/main/java/org/springframework/xd/module/spark/streaming/SparkStreamingDriverModule.java
ilayaperumalg: For some reason, the graceful shutdown wasn't working properly hence I switched the boolean.

markfisher: but do we want to stop the underlying SparkContext itself? I'm mainly concerned that we are not consistent here (for the Java version, we just call close())

spring-xd-module/src/main/java/org/springframework/xd/module/spark/streaming/SparkStreamingDriverModule.java
ilayaperumalg: Yeah, it creates a unique App ID for them and I have just verified that as well.

spring-xd-module/src/main/java/org/springframework/xd/module/spark/streaming/SparkStreamingDriverModule.java
ilayaperumalg: yeah, we want to stop the sparkContext as we are not sharing the same sparkContext across modules. Looking at it again I realized we refactored the JavaStreamingContext to use the underlying streamingContext. I should have used the same streaming context to stop. fixing it.

spring-xd-spark-streaming/src/main/java/org/springframework/xd/spark/streaming/SparkStreamingSupport.java
```diff
@@ -13,40 +13,37 @@
...
 
 	public static final String SPARK_STORAGE_LEVEL_MODULE_OPTION = "storageLevel";
 
+	public static final String SPARK_DEFAULT_STORAGE_LEVEL = "MEMORY_ONLY";
+
+	public static final String SPARK_STREAMING_BATCH_INTERVAL_PROP = "spark.streaming.batchInterval";
+
+	public static final String SPARK_STREAMING_BATCH_INTERVAL_MODULE_OPTION = "batchInterval";
+
+	public static final String SPARK_STREAMING_BATCH_INTERVAL = "2000";
```
markfisher: should this one include "DEFAULT" like the others above?

spring-xd-spark-streaming/src/main/scala/org/springframework/xd/spark/streaming/examples/SparkScalaWordCount.scala
```diff
@@ -0,0 +1,45 @@
...
+
+import org.apache.spark.streaming.StreamingContext._
+import org.apache.spark.streaming.dstream.{DStream, ReceiverInputDStream}
+import org.springframework.xd.spark.streaming.SparkConfig
+import org.springframework.xd.spark.streaming.scala.Processor
+
+/**
+ * @author Ilayaperumal Gopinathan
+ */
+class SparkScalaWordCount extends Processor[String, (String, Int)] {
```
markfisher: Do you think we should put the examples into parallel "java" and "scala" sub-packages under "examples". Then we don't need to include Scala in the class names. In fact, we might want to keep Spark out of the class names also, just "WordCount" and "Logger".

markfisher:
rebased, squashed and merged as https://github.com/spring-projects/spring-xd/commit/77bb4cb81b4ee6184696b8c68536e1a22759620c

PR #1431, opened 2015-02-04
[Sprint: 42] XD-2674: Add authentication to mongo modules

ericbottard:
Also, I believe we should consider https://github.com/thiloplanz/jmockmongo or https://github.com/flapdoodle-oss/de.flapdoodle.embed.mongo for automated tests

### DIFF
--- a/modules/job/hdfsmongodb/config/hdfsmongodb.xml
+++ b/modules/job/hdfsmongodb/config/hdfsmongodb.xml
@@ -56,7 +56,7 @@
 		fs.defaultFS=${fsUri}
 	</hdp:configuration>
 	<hdp:resource-loader id="hadoopResourceLoader"/>
-	<mongo:db-factory dbname="${databaseName}" host="${host}" port="${port}" username="${username}" password="${password}"/>
+	<mongo:db-factory dbname="${databaseName}" host="${host}" port="${port}" username="${username}" password="${password}" write-concern="${writeConcern}"/>
 
 	<mongo:mapping-converter>
 		<mongo:custom-converters>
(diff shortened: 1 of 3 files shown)


artembilan:
`write-concern` would be good option too.
Re. Embedded Mongo: I use the last one for the SI-Java-DSL: https://github.com/artembilan/spring-integration-java-dsl/blob/INTEXT-130/src/test/java/org/springframework/integration/dsl/test/mongodb/MongoDbTests.java.
And pay attention to this: https://github.com/spring-projects/spring-boot/issues/1794

ericbottard:
@artembilan yeah why not, even though the original request is about authentication.
I never know where to put the write-concern attribute though: on the factory, or the mongo element?

artembilan:
According to the `<mongo:mongo>`'s `write-concern` description, it will be fine for us specify it there, because we have only a single `<mongo:db-factory>`. But, if we don't have the first one we can use it on the `<mongo:db-factory>`. For us it's just a PP holder, so, it will work fine.

Although I'm not sure what he asks about SSL http://stackoverflow.com/questions/28280206/how-can-i-use-authentication-in-mongo-sink

markpollack:
Mongo is in xd/lib - due to same issue we had with oxm... Not happy about it but there you go.  As such I think we should be calling back out of the box to support all the properties boot supports (I think write concern is one of them) I'm going through this now for Redis sink.  I think the rabbit support is one to model ourselves after....

artembilan:
Well. Than this is what we need: https://github.com/spring-projects/spring-boot/blob/master/spring-boot-autoconfigure/src/main/java/org/springframework/boot/autoconfigure/mongo/MongoProperties.java

markpollack:
Yea, no write concern but mongodb.authentication-database
Maybe useful

ericbottard:
Rebased and added write concern

markpollack:
Should add support for authentication-database

mbogoevici:
Added support for authenticationDatabase.

Merged w/https://github.com/spring-projects/spring-xd/commit/225eb19ec79a850ac93d4536401e604fb774818e

PR #1434, opened 2015-02-04
[Sprint 42] XD-2666 Upgrade to Spring Integration Kafka 1.0.0.RC1

mbogoevici:
- library upgrade
- adjustments to the API
- removed local implementations


modules/source/kafka/config/kafka.xml
```diff
@@ -68,23 +58,42 @@
 	</bean>
 
 	<bean id="kafkaInboundChannelAdapter"
```
artembilan: How about to change to the namespace component - `<int-kafka:message-driven-channel-adapter>` ?

extensions/spring-xd-extension-kafka/src/main/java/org/springframework/integration/x/kafka/AutoOffsetResetStrategy.java
```diff
@@ -14,26 +14,27 @@
...
- * To be refactored once https://jira.spring.io/browse/INTEXT-131 is fixed
+ * The possible values for automatically resetting the offset of Kafka client
  *
  * @author Marius Bogoevici
  */
-public interface DeletableOffsetManager extends OffsetManager {
+public enum AutoOffsetResetStrategy {
+
+	smallest(OffsetRequest.EarliestTime()),
+	largest(OffsetRequest.LatestTime());
```
artembilan: Can't they be in upper case?

mbogoevici: No, they parse automatically from the module options.

modules/source/kafka/config/kafka.xml
mbogoevici: I'd rather make that a later enhancement (seeing that all this is internal) and keep current changes to a minimum,

artembilan:
So, nothing to complain about, if all existing tests pass :smile:

markpollack:
squashed, rebased and merged as 044a242d172cb664a87172bee7f306a049bf4c13

PR #1435, opened 2015-02-04
[Sprint: 42] Xd 2684 - Update to java 1.7

dturanski:
Update to jdk 1.7

### DIFF
--- a/spring-xd-starters/spring-xd-module-parent/publish-maven.gradle
+++ b/spring-xd-starters/spring-xd-module-parent/publish-maven.gradle
@@ -55,13 +55,14 @@ def customizePom(pom, gradleProject) {
 
 			properties {
 				'project.build.sourceEncoding' 'UTF-8'
-				'java.version' '1.6'
+				'java.source.version' '1.7'
+				'java.target.version' '1.6'
 				'resource.delimiter' '@'
 				'project.reporting.outputEncoding' 'UTF-8'
-				'maven.compiler.source' '${java.version}'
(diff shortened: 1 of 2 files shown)


spring-xd-starters/spring-xd-module-parent/publish-maven.gradle
```diff
@@ -162,7 +163,7 @@ def customizePom(pom, gradleProject) {
 				dependency {
 					groupId 'org.apache.hadoop'
 					artifactId 'hadoop-common'
-					version '2.3.0'
+					version '2.6.0'
```
mbogoevici: Ok, got it (scrap my earlier comment). Both the boot dependency and this were out of sync with `dependencies.properties`.

ericbottard: so, shouldn't we be using the value from the .properties file then?

mbogoevici: We probably could, but I would suggest creating a separate issue for it, as it's already stretching the scope of the original PR.

mbogoevici: LGTM; captured @ericbottard 's suggestion in https://jira.spring.io/browse/XD-2694

mbogoevici:
Merged w/https://github.com/spring-projects/spring-xd/commit/a26badf5e1c4e1db21f2a72ef54159cff72fc055

PR #1437, opened 2015-02-04
[Sprint: 42] A Bunch O' Sonar Fixes

garyrussell:
Various Fixes ("Critical" Sonar Violations).

In some cases, I added comments that the issue is ignored so we don't have to keep evaluating.

We don't want to disable the rule because it's an important one, but in these cases, does not apply.

### DIFF
--- a/extensions/spring-xd-extension-kafka/src/main/java/org/springframework/integration/x/kafka/KafkaPartitionAllocator.java
+++ b/extensions/spring-xd-extension-kafka/src/main/java/org/springframework/integration/x/kafka/KafkaPartitionAllocator.java
@@ -132,6 +132,7 @@ public class KafkaPartitionAllocator implements InitializingBean, FactoryBean<Pa
 		if (STARTED.equals(this.client.getState())) {
 			EnsurePath ensurePath = new EnsurePath(partitionDataPath);
 			ensurePath.ensure(client.getZookeeperClient());
+			// ignore sonar's complaint about this not being synchronized
 			this.partitionDataMutex = new InterProcessMutex(client, partitionDataPath);
 		}
 		else {
--- a/extensions/spring-xd-extension-kafka/src/main/java/org/springframework/integration/x/kafka/KafkaPartitionAllocator.java
+++ b/extensions/spring-xd-extension-kafka/src/main/java/org/springframework/integration/x/kafka/KafkaPartitionAllocator.java
@@ -132,8 +132,7 @@ public class KafkaPartitionAllocator implements InitializingBean, FactoryBean<Pa
 		if (STARTED.equals(this.client.getState())) {
 			EnsurePath ensurePath = new EnsurePath(partitionDataPath);
 			ensurePath.ensure(client.getZookeeperClient());
-			// ignore sonar's complaint about this not being synchronized
-			this.partitionDataMutex = new InterProcessMutex(client, partitionDataPath);
+			this.partitionDataMutex = new InterProcessMutex(client, partitionDataPath);//NOSONAR
 		}
 		else {
 			throw new BeanInitializationException("Cannot connect to ZooKeeper, client state is " + client.getState());
--- a/spring-xd-dirt/src/main/java/org/springframework/xd/dirt/stream/completion/StacktraceFingerprintingCompletionRecoveryStrategy.java
+++ b/spring-xd-dirt/src/main/java/org/springframework/xd/dirt/stream/completion/StacktraceFingerprintingCompletionRecoveryStrategy.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2013 the original author or authors.
+ * Copyright 2013-2015 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -32,7 +32,7 @@ import org.springframework.xd.rest.domain.CompletionKind;
  * encounters the ill formed input. Multiple fingerprints are supported, as the control flow in the parser code may be
  * different depending on the form of the expression. For example, for the rule {@code stream = module (| module)* },
  * the pseudo code for the parser may look like
- * 
+ *
  * <pre>
  * <code>
  * stream() {
@@ -44,11 +44,12 @@ import org.springframework.xd.rest.domain.CompletionKind;
  * }
  * </code>
  * </pre>
- * 
+ *
  * In that setup, whether we're dealing with the first module, or a subsequent module, stack frames would be different
  * (see (1) and (2)).
- * 
+ *
  * @author Eric Bottard
+ * @author Gary Russell
  */
 public abstract class StacktraceFingerprintingCompletionRecoveryStrategy<E extends Exception> implements
 		CompletionRecoveryStrategy<E> {
@@ -64,7 +65,7 @@ public abstract class StacktraceFingerprintingCompletionRecoveryStrategy<E exten
 	/**
 	 * Construct a new StacktraceFingerprintingCompletionRecoveryStrategy given the parser, and the expected exception
 	 * class to be thrown for sample fragments of a stream definition that is to be parsed.
-	 * 
+	 *
 	 * @param parser the parser used to parse the sample fragment stream definitions.
 	 * @param exceptionClass the expected exception that results from parsing the sample fragment stream definitions.
 	 *        Stack frames from the thrown exception are used to store the fingerprint of this exception thrown by the
@@ -90,15 +91,12 @@ public abstract class StacktraceFingerprintingCompletionRecoveryStrategy<E exten
 					// not influenced by the kind of parse. Use stream for now
 					this.parser.parse("__dummy", sample, ParsingContext.partial_stream);
 				}
-				catch (Exception exception) {
+				catch (RuntimeException exception) {
 					if (this.exceptionClass.isAssignableFrom(exception.getClass())) {
 						computeFingerprint((E) exception);
 					}
-					else if (exception instanceof RuntimeException) {
-						throw (RuntimeException) exception;
-					}
 					else {
-						throw new RuntimeException(exception);
+						throw exception;
 					}
 				}
 			}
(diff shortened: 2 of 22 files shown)


extensions/spring-xd-extension-kafka/src/main/java/org/springframework/integration/x/kafka/KafkaPartitionAllocator.java
```diff
@@ -132,6 +132,7 @@ public void afterPropertiesSet() throws Exception {
 		if (STARTED.equals(this.client.getState())) {
 			EnsurePath ensurePath = new EnsurePath(partitionDataPath);
 			ensurePath.ensure(client.getZookeeperClient());
+			// ignore sonar's complaint about this not being synchronized
```
ericbottard: Are you saying that this sole comment stops Sonar from complaining? Does it have to follow a given pattern?

spring-xd-dirt/src/main/java/org/springframework/xd/dirt/stream/completion/StacktraceFingerprintingCompletionRecoveryStrategy.java
```diff
@@ -90,15 +91,12 @@ private void initFingerprints() {
 					// not influenced by the kind of parse. Use stream for now
 					this.parser.parse("__dummy", sample, ParsingContext.partial_stream);
 				}
-				catch (Exception exception) {
```
ericbottard: Not sure about this one. It used to be able to catch (and compute stacktrace fingerprint) of checked exceptions, and now doesn't

spring-xd-messagebus-spi/src/main/java/org/springframework/xd/dirt/integration/bus/MessageBusSupport.java
```diff
@@ -898,7 +898,11 @@ private void revertDirectBindingIfNecessary(Binding binding) {
 
 		@Override
 		public int selectPartition(Object key, int partitionCount) {
-			return Math.abs(key.hashCode()) % partitionCount;
+			int hashCode = key.hashCode();
+			if (hashCode == Integer.MIN_VALUE) {
+				hashCode = 0;
+			}
+			return Math.abs(hashCode) % partitionCount;
```
ericbottard: Actually, I think the modulo should not be part of the strategy as it is always applied.

Even, on closer inspection, I think there is an issue with 
```
if (partition >= meta.partitionCount) {
			partition = partition % meta.partitionCount;
		}
		if (partition < 0) {
			partition = Math.abs(partition);
		}
```

=> What if partition is less than `-meta.partitionCount` ? The modulo should be always applied, then abs() IMHO

spring-xd-module/src/main/java/org/springframework/xd/module/ModuleDefinition.java
```diff
@@ -76,21 +73,37 @@ public int compareTo(ModuleDefinition other) {
...
 	@Override
-	public final boolean equals(Object o) {
-		if (this == o) return true;
-
-		ModuleDefinition that = (ModuleDefinition) o;
-
-		if (!name.equals(that.name)) return false;
-		if (type != that.type) return false;
-
+	public boolean equals(Object obj) {
```
ericbottard: Should be careful about this one. IMO the previous implementation better reflects that e.g. can't create/upload module `source:foo` if there is already a `source:foo` (be it composed, or not, etc)

spring-xd-module/src/main/java/org/springframework/xd/module/ModuleDefinition.java
```diff
@@ -76,21 +73,37 @@ public int compareTo(ModuleDefinition other) {
...
+		if (this == obj) {
+			return true;
+		}
+		if (obj == null) {
+			return false;
+		}
+		if (getClass() != obj.getClass()) {
```
ericbottard: So to clarify, **this** line bothers me

spring-xd-module/src/main/java/org/springframework/xd/module/ModuleDefinition.java
```diff
@@ -76,21 +73,37 @@ public int compareTo(ModuleDefinition other) {
...
+			return false;
+		}
+		ModuleDefinition other = (ModuleDefinition) obj;
+		if (name == null) {
```
ericbottard: Moreover, name and type can't be null

spring-xd-test-fixtures/src/main/java/org/springframework/xd/test/fixtures/JdbcSink.java
```diff
@@ -71,12 +71,23 @@ public JdbcTemplate getJdbcTemplate() {
 	@Override
 	protected String toDSL() {
 		StringBuilder dsl = new StringBuilder();
+		Connection connection = null;
 		try {
-			dsl.append("jdbc --initializeDatabase="+initializeDB+" --url=" + dataSource.getConnection().getMetaData().getURL());
+			connection = dataSource.getConnection();
+			dsl.append("jdbc --initializeDatabase="+initializeDB+" --url=" + connection.getMetaData().getURL());
```
ericbottard: Depending on the actual datasource we're using (this is used in tests, for the record) this, albeit "cleaner", may be problematic: what if we use a SingleConnectionDataSource. Should we use SmartDataSource or is it overkill?

extensions/spring-xd-extension-kafka/src/main/java/org/springframework/integration/x/kafka/KafkaPartitionAllocator.java
garyrussell: No; it's just so the next person that looks at resolving the "issue" will see that it's already been looked at.

I think the rule is too important to disable, but it raises these false positives.

spring-xd-dirt/src/main/java/org/springframework/xd/dirt/stream/completion/StacktraceFingerprintingCompletionRecoveryStrategy.java
garyrussell: Except `parser.parse()` cannot throw any checked exceptions :smile:

spring-xd-messagebus-spi/src/main/java/org/springframework/xd/dirt/integration/bus/MessageBusSupport.java
garyrussell: This fix was simply because `Math.abs(Integer.MIN_VALUE) == Integer.MIN_VALUE`.

I see your point but we can't change a strategy interface after a release candidate.

> What if partition is less than -meta.partitionCount ? The modulo should be always applied, then abs() IMHO

If it's already `0 >= partition < count` then it doesn't need to be modulo'd.

However, the module __should__ be applied if it's < 0 because the absolute value might be `> count`.

So I agree, but I think we're stuck with replacing your snippet with an unconditional modulo here while still passing it into the selector.

spring-xd-module/src/main/java/org/springframework/xd/module/ModuleDefinition.java
garyrussell: OK; I simply used eclipse to generate; Sonar complained that `equals()` should never assume the passed-in object is a `ModuleDefinition` (previously the cast was done unconditionally).

If someone calls `myModDef.equals("foo")` we'd get a `ClassCastException`.

Maybe we should revert and add a check for the type; but using `equals()` in this way seems like a bastardization to me.

spring-xd-test-fixtures/src/main/java/org/springframework/xd/test/fixtures/JdbcSink.java
garyrussell: ? `Connection` from `SingleConnectionDataSource` ignores the `close()`. See `CloseSuppressingInvocationHandler`.

That said, since these are test fixtures we could ignore the Sonar violation.

extensions/spring-xd-extension-kafka/src/main/java/org/springframework/integration/x/kafka/KafkaPartitionAllocator.java
garyrussell: Cool; just found `//NOSONAR`

http://docs.sonarqube.org/display/SONAR/Frequently+Asked+Questions#FrequentlyAskedQuestions-NOSONAR

garyrussell:
Rebased, pushed.

spring-xd-dirt/src/main/java/org/springframework/xd/dirt/stream/completion/StacktraceFingerprintingCompletionRecoveryStrategy.java
ericbottard: Ah. ;(

spring-xd-test-fixtures/src/main/java/org/springframework/xd/test/fixtures/JdbcSink.java
ericbottard: This depends on the `suppressClose` flag, no? Anyway, this is a little detail indeed.

spring-xd-messagebus-spi/src/main/java/org/springframework/xd/dirt/integration/bus/MessageBusSupport.java
ericbottard: To clarify, my snippet above is what is currently in `MessageBusSupport.determinePartition()`.

And what I'm saying is that
* Modulo should be applied only once ideally
* and it's our responsibility to make sure it is applied (and correctly, see my point about being negative and less than partitionCount in absolute value)

==> this logic should go in MessageBusSupport (not be left to the selector strategy, where users could screw up)

garyrussell:
Fixing conflicts after rebase this morning...

garyrussell:
Actually - sorry - @ericbottard already merged it as a0ee3ec917260a3cb99b46c7be63d4206047ad12 :smile:

ericbottard:
Sorry, forgot to close that and was on my way back while I thought about it

PR #1438, opened 2015-02-04
[Sprint: 42] XD-2683 Spark streaming module type conversion

ilayaperumalg:
- Set beanName for MessageBusReceiver's message storing channel as "input"
  - Configure the corresponding message converters in MessageBusReceiver's `onStart`
so that the dataType and converters are set and created at spark executor
  - Set beanName for MessageBusSender as "output"
  - Configure message converters inside the MessageBusSender's `start` so that
the dataType and converters are set and created at spark executor
  - Refactored `ModuleTypeConversionPlugin` configurations and made some of the
methods static so that they can be accessed by MessageBusSender/Receiver
  - Add tests

### DIFF
--- a/spring-xd-dirt/src/main/java/org/springframework/xd/dirt/plugins/spark/streaming/MessageBusReceiver.java
+++ b/spring-xd-dirt/src/main/java/org/springframework/xd/dirt/plugins/spark/streaming/MessageBusReceiver.java
@@ -16,7 +16,6 @@
 
 package org.springframework.xd.dirt.plugins.spark.streaming;
 
-import java.io.Serializable;
 import java.util.Properties;
 
 import org.apache.spark.storage.StorageLevel;
@@ -27,6 +26,7 @@ import org.slf4j.LoggerFactory;
 import org.springframework.context.ConfigurableApplicationContext;
 import org.springframework.integration.channel.DirectChannel;
 import org.springframework.messaging.Message;
+import org.springframework.util.MimeType;
 import org.springframework.xd.dirt.integration.bus.MessageBus;
 
 /**
@@ -57,16 +57,20 @@ class MessageBusReceiver extends Receiver {
 
 	private final Properties moduleConsumerProperties;
 
-	public MessageBusReceiver(StorageLevel storageLevel, Properties messageBusProperties, Properties moduleConsumerProperties) {
-		this(null, storageLevel, messageBusProperties, moduleConsumerProperties);
+	private final MimeType contentType;
+
+	public MessageBusReceiver(StorageLevel storageLevel, Properties messageBusProperties,
+			Properties moduleConsumerProperties, MimeType contentType) {
+		this(null, storageLevel, messageBusProperties, moduleConsumerProperties, contentType);
 	}
 
 	public MessageBusReceiver(LocalMessageBusHolder messageBusHolder, StorageLevel storageLevel,
-			Properties messageBusProperties, Properties moduleConsumerProperties) {
+			Properties messageBusProperties, Properties moduleConsumerProperties, MimeType contentType) {
 		super(storageLevel);
 		this.messageBusHolder = messageBusHolder;
 		this.messageBusProperties = messageBusProperties;
 		this.moduleConsumerProperties = moduleConsumerProperties;
+		this.contentType = contentType;
 	}
 
 	public void setInputChannelName(String channelName) {
@@ -76,6 +80,10 @@ class MessageBusReceiver extends Receiver {
 	@Override
 	public void onStart() {
 		logger.info("starting MessageBusReceiver");
+		final MessageStoringChannel messageStoringChannel = new MessageStoringChannel();
+		if (contentType != null) {
+			messageStoringChannel.configureMessageConverter(contentType);
+		}
 		MessageBus messageBus;
 		if (messageBusHolder != null) {
 			messageBus = messageBusHolder.get();
@@ -84,7 +92,7 @@ class MessageBusReceiver extends Receiver {
 			applicationContext = MessageBusConfiguration.createApplicationContext(messageBusProperties);
 			messageBus = applicationContext.getBean(MessageBus.class);
 		}
-		messageBus.bindConsumer(channelName, new MessageStoringChannel(), moduleConsumerProperties);
+		messageBus.bindConsumer(channelName, messageStoringChannel, moduleConsumerProperties);
 	}
 
 	@Override
@@ -96,13 +104,16 @@ class MessageBusReceiver extends Receiver {
 		}
 	}
 
-
 	/**
 	 * The {@link DirectChannel} that stores the received messages into Spark's memory.
 	 */
-	private class MessageStoringChannel extends DirectChannel implements Serializable {
+	private class MessageStoringChannel extends SparkStreamingChannel {
 
-		private static final long serialVersionUID = 1L;
+		private static final String INPUT = "input";
+
+		public MessageStoringChannel() {
+			this.setBeanName(INPUT);
+		}
 
 		@Override
 		protected boolean doSend(Message<?> message, long timeout) {
--- a/spring-xd-dirt/src/main/java/org/springframework/xd/dirt/plugins/spark/streaming/MessageBusSender.java
+++ b/spring-xd-dirt/src/main/java/org/springframework/xd/dirt/plugins/spark/streaming/MessageBusSender.java
@@ -22,6 +22,8 @@ import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
 
 import org.springframework.context.ConfigurableApplicationContext;
+import org.springframework.messaging.Message;
+import org.springframework.util.MimeType;
 import org.springframework.xd.dirt.integration.bus.MessageBus;
 import org.springframework.xd.spark.streaming.SparkMessageSender;
 
@@ -53,19 +55,30 @@ class MessageBusSender extends SparkMessageSender {
 
 	private boolean running = false;
 
-	public MessageBusSender(String outputChannelName, Properties messageBusProperties, Properties moduleProducerProperties) {
-		this(null, outputChannelName, messageBusProperties, moduleProducerProperties);
+	private static final String OUTPUT = "output";
+
+	private final MimeType contentType;
+
+	private final SparkStreamingChannel outputChannel;
+
+	public MessageBusSender(String outputChannelName, Properties messageBusProperties,
+			Properties moduleProducerProperties, MimeType contentType) {
+		this(null, outputChannelName, messageBusProperties, moduleProducerProperties, contentType);
 	}
 
-	public MessageBusSender(LocalMessageBusHolder messageBusHolder, String outputChannelName, Properties messageBusProperties, Properties moduleProducerProperties) {
+	public MessageBusSender(LocalMessageBusHolder messageBusHolder, String outputChannelName,
+			Properties messageBusProperties, Properties moduleProducerProperties, MimeType contentType) {
 		this.messageBusHolder = messageBusHolder;
 		this.outputChannelName = outputChannelName;
 		this.messageBusProperties = messageBusProperties;
 		this.moduleProducerProperties = moduleProducerProperties;
+		this.contentType = contentType;
+		this.outputChannel = new SparkStreamingChannel();
 	}
 
 	@Override
 	public synchronized void start() {
+		outputChannel.setBeanName(OUTPUT);
 		if (messageBus == null) {
 			logger.info("starting MessageBusSender");
 			if (messageBusHolder != null) {
@@ -75,16 +88,24 @@ class MessageBusSender extends SparkMessageSender {
 				applicationContext = MessageBusConfiguration.createApplicationContext(messageBusProperties);
 				messageBus = applicationContext.getBean(MessageBus.class);
 			}
-			messageBus.bindProducer(outputChannelName, this, moduleProducerProperties);
+			if (contentType != null) {
+				outputChannel.configureMessageConverter(contentType);
+			}
+			messageBus.bindProducer(outputChannelName, outputChannel, moduleProducerProperties);
 		}
 		this.running = true;
+	}
+
+	@Override
+	public synchronized void send(Message message) {
+		this.outputChannel.send(message);
 	}
 
 	@Override
 	public synchronized void stop() {
 		if (this.isRunning() && messageBus != null) {
 			logger.info("stopping MessageBusSender");
-			messageBus.unbindProducer(outputChannelName, this);
+			messageBus.unbindProducer(outputChannelName, outputChannel);
 			if (applicationContext != null) {
 				applicationContext.close();
 				applicationContext = null;
@@ -98,5 +119,4 @@ class MessageBusSender extends SparkMessageSender {
 	public synchronized boolean isRunning() {
 		return this.running;
 	}
-
 }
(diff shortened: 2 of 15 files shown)


spring-xd-dirt/src/main/java/org/springframework/xd/dirt/plugins/stream/ModuleTypeConversionPlugin.java
```diff
@@ -66,15 +70,26 @@ public ModuleTypeConversionPlugin(Collection<AbstractFromMessageConverter> conve
...
+		return this.converterFactory;
+	}
+
 	@Override
 	public void postProcessModule(Module module) {
 		String outputType = null;
 		String inputType = null;
 		if (module.getType() == ModuleType.source || module.getType() == ModuleType.processor) {
-			outputType = module.getProperties().getProperty("outputType");
+			outputType = module.getProperties().getProperty(OUTPUT_TYPE);
```
ericbottard: Ideally, we should avoid using `getProperties()`, so the code that does this should be factored out. getMimeType() in SparkStreamingPlugin is exactly what we want, so should convert it to utility code.

spring-xd-dirt/src/main/java/org/springframework/xd/dirt/plugins/stream/ModuleTypeConversionPlugin.java
```diff
@@ -142,14 +166,14 @@ private MimeType resolveContentType(String type, Module module) throws ClassNotF
...
-	private Class<?> resolveJavaType(String type, Module module) throws ClassNotFoundException, LinkageError {
+	private static Class<?> resolveJavaType(String type, Module module) throws ClassNotFoundException, LinkageError {
 		SimpleModule sm = (SimpleModule) module;
 		return ClassUtils.forName(type, sm.getApplicationContext().getClassLoader());
 	}
 
 	@Override
 	public boolean supports(Module module) {
-		return true;
+		return module.shouldBind();
```
ericbottard: I'm not sure I understand the rationale here. Why would a module that doesn't bind locally (if this is what should be understood from `shouldBind()`) not support type conversion? IMO type conversion is a universal XD feature that should apply to all, whether they run on XD proper or elsewhere

spring-xd-spark-streaming-tests/src/test/java/org/springframework/xd/spark/streaming/RabbitTransportSparkStreamingTests.java
```diff
@@ -47,9 +47,8 @@ protected void createStream(String streamName, String stream) {
 		addQueueNames(streamName, stream);
 	}
 
-	@After
-	public void cleanupRabbitQueues() {
-		super.tearDown();
+	@AfterClass
```
ericbottard: Don't agree with this: what if there are 2 tests that use the same queue names. Should be cleared between tests, not at the end of the "suite"

spring-xd-dirt/src/main/java/org/springframework/xd/dirt/plugins/spark/streaming/SparkStreamingMessageConverterSupport.java
```diff
@@ -0,0 +1,72 @@
...
+public class SparkStreamingMessageConverterSupport {
+
+
+	/**
+	 * Configure the data type and message converters to be used by the channel on the payload.
+	 *
+	 * @param mimeType the MIME type configured
+	 * @param channel the message channel to send the converted messages
+	 */
+	protected static void configureMessageConverter(MimeType mimeType, AbstractMessageChannel channel) {
```
ericbottard: Feels like there is something wrong with this IMHO, it kinda violates the DRY principle:
* either   it applies anywhere and it should not live in the spark package,
* or it duplicates a more "springy" way of doing it, and that approach should be used in Sparkland as well

spring-xd-dirt/src/main/java/org/springframework/xd/dirt/plugins/stream/ModuleTypeConversionPlugin.java
dturanski: agree

spring-xd-dirt/src/main/java/org/springframework/xd/dirt/plugins/spark/streaming/MessageBusReceiver.java
```diff
@@ -104,6 +110,10 @@ public void onStop() {
 
 		private static final long serialVersionUID = 1L;
 
+		public MessageStoringChannel() {
+			this.setBeanName("input");
```
dturanski: extract to constant as you did below?

spring-xd-dirt/src/main/java/org/springframework/xd/dirt/plugins/spark/streaming/MessageBusSender.java
```diff
@@ -53,19 +54,27 @@
 
 	private boolean running = false;
 
-	public MessageBusSender(String outputChannelName, Properties messageBusProperties, Properties moduleProducerProperties) {
-		this(null, outputChannelName, messageBusProperties, moduleProducerProperties);
+	private static final String BEAN_NAME = "output";
```
dturanski: Clearer to call this OUTPUT

spring-xd-dirt/src/main/java/org/springframework/xd/dirt/plugins/stream/ModuleTypeConversionPlugin.java
markfisher: I made a comment on the pre-PR commit that was similar, but my point was that "shouldBind()" is not quite capturing what this boolean indicates now that it's expanded to this as well. The fundamental challenge here is that the MessageConverters cannot be created on the XD container node, because they are not serializable (the base class has no public default constructor and that is in core Spring, not XD so we cannot workaround it). That means the converters have to be created within the actual Spark executor process where they will be used. That is why the plugin code is being reused from the lifecycle methods (on start) in those instances that have themselves been serialized and shipped to the cluster nodes. @ilayaperumalg can surely provide more detail, but that's my quick overview of the situation.

spring-xd-dirt/src/main/java/org/springframework/xd/dirt/plugins/spark/streaming/SparkStreamingMessageConverterSupport.java
dturanski: Yeah. Can we DI the ModuleTypeConversionPlugin and expose the appropriate method to do this?

spring-xd-dirt/src/main/java/org/springframework/xd/dirt/plugins/spark/streaming/SparkStreamingPlugin.java
```diff
@@ -100,20 +102,22 @@ public void postProcessModule(Module module) {
 			}
 			LocalMessageBusHolder messageBusHolder = new LocalMessageBusHolder();
 			messageBusHolder.set(module.getComponent(MessageBus.class));
-			receiver = new MessageBusReceiver(messageBusHolder, storageLevel, messageBusProperties, inboundModuleProperties);
+			receiver = new MessageBusReceiver(messageBusHolder, storageLevel, messageBusProperties,
+					inboundModuleProperties, getMimeType(module, true));
```
dturanski: the meaning of the true and false args here are unclear here. Can we pass ModuleTypeConversionPlugin.INPUT_TYPE and ModuleTypeConversionPlugin.OUTPUT_TYPE here instead?

spring-xd-dirt/src/main/java/org/springframework/xd/dirt/plugins/spark/streaming/SparkStreamingMessageConverterSupport.java
markfisher: See my comment here: https://github.com/spring-projects/spring-xd/pull/1438/files#r24163988

I do agree we should try to do this in a more Springy way, but the key challenge is that the ApplicationContext for the MessageConverters must be created on the Spark cluster process. Perhaps we can refactor the plugin in such a way that it can be applied to the module there without having to poke these arbitrary holes in it.

spring-xd-spark-streaming-tests/src/test/java/org/springframework/xd/spark/streaming/AbstractSparkStreamingTests.java
```diff
@@ -108,12 +110,65 @@ public void testSparkProcessor() throws Exception {
...
+			createStream(streamName, stream);
+			source.ensureReady().postData(TEST_MESSAGE);
+			assertThat(sink, XDMatchers.eventually(XDMatchers.hasContentsThat(equalTo("{\"_1\":\"foo\",\"_2\":3}"))));
+		}
+		finally {
+			streamOps.destroyStream(streamName);
+			sink.cleanup();
+		}
+	}
+
```
dturanski: In general I prefer using the processing chain tests for testing a single processor which allow you to deploy a stream containing a processor under test which you can send and receive messages directly. This is much lighter weight and less brittle than deploying using an embedded shell, posting an http message and checking the contents of a file.  For testing a sink, using file output is appropriate, but for processor tests this is unnecessary.

ericbottard: Agreed, although having at least ONE test of each kind makes sure that the rest of the infrastructure works correctly in a real setting..

spring-xd-dirt/src/main/java/org/springframework/xd/dirt/plugins/stream/ModuleTypeConversionPlugin.java
dturanski: I'm not sure I understand the problem. There is no requirement for a public default constructor with Kryo serialization. Does spark use something else?

spring-xd-dirt/src/main/java/org/springframework/xd/dirt/plugins/spark/streaming/SparkStreamingMessageConverterSupport.java
ilayaperumalg: +1 on what Mark said. We also tried to do it through plugin (by setting the dataType and converters). But again they have to serializable. Though Spark streaming module context has both `MessageBusReceiver` and `MessageBusSender` (where the input and output channels are obtained), the channel has to be configured with message converters at the spark cluster's executor node as it is serialized/deserialized.

spring-xd-dirt/src/main/java/org/springframework/xd/dirt/plugins/spark/streaming/MessageBusSender.java
ilayaperumalg: sure.

spring-xd-spark-streaming-tests/src/test/java/org/springframework/xd/spark/streaming/RabbitTransportSparkStreamingTests.java
ilayaperumalg: This is just the rabbit queues that we want to delete after the test run (as we don't normally delete the rabbit queues as part of stream destroy). Since we destroy the streams between the tests, I think we are good here... Moreover, I had an issue (one of the queues wasn't cleaned up) running the cleanup of rabbit queues between the tests.

spring-xd-dirt/src/main/java/org/springframework/xd/dirt/plugins/stream/ModuleTypeConversionPlugin.java
markfisher: Spark does not use Kryo. They have custom serialization, and don't even serialize based on class boundaries since the code ends up running partially in the driver (which in our case is within the XD container process) and partially on the worker nodes.

spring-xd-dirt/src/main/java/org/springframework/xd/dirt/plugins/spark/streaming/SparkStreamingMessageConverterSupport.java
dturanski: by serializable you mean implement java.io.Serializable?

spring-xd-dirt/src/main/java/org/springframework/xd/dirt/plugins/spark/streaming/SparkStreamingPlugin.java
ilayaperumalg: Initially I went through this route. But, I would have to set either of these to empty string when calling the method as we only need either the input or output MIME type. or, do you mean some other way?

markfisher: I agree with David, and I think it might be worth 2 separate methods: getInputMimeType and getOutputMimeType, even if they in turn call a common private method to do part/most of the work.

spring-xd-dirt/src/main/java/org/springframework/xd/dirt/plugins/stream/ModuleTypeConversionPlugin.java
markfisher: Ilaya created this JIRA to capture the post-1.1 refactoring:
https://jira.spring.io/browse/XD-2695

spring-xd-dirt/src/main/java/org/springframework/xd/dirt/plugins/spark/streaming/SparkStreamingMessageConverterSupport.java
```diff
@@ -0,0 +1,48 @@
...
+import org.springframework.xd.dirt.plugins.stream.ModuleTypeConversionPlugin;
+import org.springframework.xd.dirt.plugins.stream.ModuleTypeConversionSupport;
+import org.springframework.xd.dirt.util.ConfigLocations;
+
+/**
+ * Utility class to support message conversion in spark streaming module.
+ *
+ * @author Ilayaperumal Gopinathan
+ */
+public class SparkStreamingMessageConverterSupport {
```
markfisher: Now that the method below has been factored out, I'm wondering if it would make sense to consider these steps even in the short term (as part of this PR):

1. Create some base class like SparkStreamingChannel that extends DirectChannel
2. move the configureMessageConverter() method to that new class (no longer static, and only requiring the single contentType parameter)
3. have SparkMessageSender extend SparkStreamingChannel instead of DirectChannel
4. have MessageStoringChannel extend SparkStreamingChannel instead of DirectChannel
5. replace the 2 current calls to this static configureMessageConverter to then call on the MessageStoringChannel and MessageBusSender instances instead

ilayaperumalg: This is a nice idea. I see some package constraints with this approach. The `SparkMessageSender` (which is in **spring-xd-spark-streaming** and subsequently in spring-xd-module) needs to extend this `SparkStreamingChannel` so that it could be used to send messages in `ModuleExecutor` while the SparkStreamingChannel could only be in **spring-xd-dirt** package as it needs ModuleTypeConversionPlugin

ilayaperumalg: trying...

markfisher:
LGTM; merging

markfisher:
squashed, rebased, and merged as https://github.com/spring-projects/spring-xd/commit/e5df4f49b851703b3d15ebe556bf2b88ed1dbcef

PR #1439, opened 2015-02-05
[Sprint 42] XD-2578 Add Spring OXM to the DIRT libraries

mbogoevici:
This is necessary because Spring OXM is a direct dependency of Spring Batch infrastructure, and batch jobs that depend on it cannot load, even if provided with the job module, due to the parent/child class hierarchy.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -478,6 +478,13 @@ project('spring-xd-dirt') {
 		// are loaded by RestTemplate and RestTemplate is in Dirt
 		runtime "org.codehaus.jackson:jackson-mapper-asl"
 
+
+		// The following are optional dependencies of Spring Batch/Spring Batch Infrastructure
+		// Adding them is solving classloading issues such as https://jira.spring.io/browse/XD-2578
+		runtime ("org.springframework:spring-oxm") {
+			exclude group: "commons-lang", module: "commons-lang"
+		}
+
 		// ************* Test
 		testCompile project(":spring-xd-test")
 		testCompile "org.springframework.integration:spring-integration-test"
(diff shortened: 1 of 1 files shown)


ericbottard:
For all those libs that we include on Dirt, we should clearly mark **why** we did this, so that when the time comes to fix the root problem, we can move them back to modules.

Moreover, I actually don't think that this is a good idea. We can't add the whole maven repository to dirt's CP just in case someone may want to use them.

markpollack:
Yes, we discussed that aspect.  XML processing is very generic/expected out of the box vs say the other dependencies that could have received the same treatment, hibernate, gemfire, ibatis.  In this case, until we can get batch more modular from the packaging perspective, this addition seems quite reasonable to me.  We should perhaps add something to the docs - not sure where - about when to add to XD/lib

mbogoevici:
I'm not happy with this solution either, but in the particular context of the problem it is the only pragmatic solution, I am afraid, and @markpollack has provided the justification. While I agree with @ericbottard  in principle that we can't keep adding dependencies indiscriminately, we also *must* ensure that the out-of-the box experience is not confusing/difficult for the most common cases. (e.g. XML processing).

There is a corollary to this, which is that we also must to have a way to specify versions for non-DIRT dependencies (i.e. what's in dependencies.properties). I'm just looking at the original issue, noticing that the provided example uses an older Spring version (4.0.6.RELEASE vs 4.1.1.RELEASE that the target DIRT version 1.1.0.M2 had). Maybe the used targeted an older DIRT version. Anyhow, that is a separate discussion, but it ties into the fact that it's a bad idea to encourage users to provide parts of the Spring Framework at a different version than what we have in DIRT.

As for *why* we did this, I though mentioning the JIRA item that required it next to the inclusion would be enough. Should we do something else? I am open to suggestions.

markpollack:
Users are required to use the versions of Spring projects that we use in DIRT

mbogoevici:
What I meant was that there's no clear way to communicate to end-users what that version must be. Making sure that such dependencies are included in the <dependency-management/> would help I think (but that's separate discussion).

PR #1442, opened 2015-02-05
[Sprint: 42] XD-2690 - Filter lib files other than .zip and .jar

dturanski:


spring-xd-module/src/main/java/org/springframework/xd/module/support/ModuleUtils.java
```diff
@@ -37,7 +37,11 @@
  */
 public class ModuleUtils {
 
-	private static final AsciiBytes LIB = new AsciiBytes("lib/");
+	private static final String LIB = "lib/";
+
+	public static final String DOT_JAR = ".jar";
+
+	public static final String DOT_ZIP = ".zip";
```
ericbottard: why even accept .zip? We won't add them to the CP currently

mbogoevici: `.zip` is a legitimate extension for a CP element. See http://docs.oracle.com/javase/7/docs/technotes/tools/windows/classpath.html. I don't see what we could possibly gain by not accepting them. 

@ericbottard can you clarify the second part of the comment: org.springframework.xd.module.support.ModuleUtils#createModuleClassLoader(org.springframework.core.io.Resource, java.lang.ClassLoader, boolean) doesn't seem to do any filtering of archives of its own.

ericbottard: My bad, I was under the impression we only looked up .jar
So yeah, we shall accept .zip
The more, the merrier...

mbogoevici: OK :) I'll LGTM it, and add it to the merge pipeline.

mbogoevici:
Merged w/https://github.com/spring-projects/spring-xd/commit/97aa52a5a610572fa8179d9f6a4790122bc3a701

PR #1445, opened 2015-02-06
[Sprint 42] XD-2698 Kafka tests use an external broker by default

mbogoevici:
- Make existing rule configurable, using the default Zookeeper settings;
- When running externalized, Kafka tests will not start if server is not available;
- allow the `XD_KAFKA_TEST_EMBEDDED=true` system property for enabling embedded tests, if necessary;
- ensure that other tests are not affected by the running ZK

### DIFF
--- a/spring-xd-spark-streaming-tests/src/test/java/org/springframework/xd/spark/streaming/KafkaTransportSparkStreamingTests.java
+++ b/spring-xd-spark-streaming-tests/src/test/java/org/springframework/xd/spark/streaming/KafkaTransportSparkStreamingTests.java
@@ -16,10 +16,6 @@
 
 package org.springframework.xd.spark.streaming;
 
-import java.util.Properties;
-
-import kafka.utils.TestUtils;
-
 import org.junit.ClassRule;
 
 import org.springframework.xd.dirt.integration.bus.KafkaConnectionPropertyNames;
@@ -31,16 +27,11 @@ import org.springframework.xd.test.kafka.KafkaTestSupport;
 public class KafkaTransportSparkStreamingTests extends AbstractSparkStreamingTests {
 
 	@ClassRule
-	public static KafkaTestSupport kafkaTestSupport;
+	public static KafkaTestSupport kafkaTestSupport = new KafkaTestSupport();
 
 	static {
-		kafkaTestSupport = new KafkaTestSupport();
-		Properties brokerConfigProperties = TestUtils.createBrokerConfig(0, TestUtils.choosePort());
-		kafkaTestSupport.setBrokerConfig(brokerConfigProperties);
-		System.setProperty(KafkaConnectionPropertyNames.KAFKA_BROKERS,
-				brokerConfigProperties.getProperty("host.name") + ":" + brokerConfigProperties.getProperty("port"));
-		System.setProperty(KafkaConnectionPropertyNames.KAFKA_ZK_ADDRESS,
-				brokerConfigProperties.getProperty("zookeeper.connect"));
+		System.setProperty(KafkaConnectionPropertyNames.KAFKA_BROKERS, kafkaTestSupport.getBrokerAddress());
+		System.setProperty(KafkaConnectionPropertyNames.KAFKA_ZK_ADDRESS, kafkaTestSupport.getZkConnectString());
 	}
 
 	public KafkaTransportSparkStreamingTests() {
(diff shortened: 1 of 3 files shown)


spring-xd-spark-streaming-tests/src/test/java/org/springframework/xd/spark/streaming/KafkaTransportSparkStreamingTests.java
```diff
@@ -31,16 +27,11 @@
...
-		kafkaTestSupport = new KafkaTestSupport();
-		Properties brokerConfigProperties = TestUtils.createBrokerConfig(0, TestUtils.choosePort());
-		kafkaTestSupport.setBrokerConfig(brokerConfigProperties);
-		System.setProperty(KafkaConnectionPropertyNames.KAFKA_BROKERS,
-				brokerConfigProperties.getProperty("host.name") + ":" + brokerConfigProperties.getProperty("port"));
-		System.setProperty(KafkaConnectionPropertyNames.KAFKA_ZK_ADDRESS,
-				brokerConfigProperties.getProperty("zookeeper.connect"));
+		System.setProperty(KafkaConnectionPropertyNames.KAFKA_BROKERS, kafkaTestSupport.getBrokerAddress());
+		System.setProperty(KafkaConnectionPropertyNames.KAFKA_ZK_ADDRESS, kafkaTestSupport.getZkConnectString());
 	}
```
mminella: Why is this stuff done in a static block instead of using @BeforeClass.  That way you could clean it up via @AfterClass?

spring-xd-test/src/main/java/org/springframework/xd/test/kafka/KafkaTestSupport.java
```diff
@@ -49,81 +62,99 @@
...
+				log.debug("Creating Kafka server");
+				Properties brokerConfigProperties = brokerConfig;
+				kafkaServer = TestUtils.createServer(new KafkaConfig(brokerConfigProperties), SystemTime$.MODULE$);
+				log.debug("Created Kafka server at " + kafkaServer.config().hostName() + ":" + kafkaServer.config().port());
+			}
+			catch (Exception e) {
+				zookeeper.shutdown();
+				zkClient.close();
+				throw new RuntimeException("Issues creating the Kafka server", e);
+			}
```
mminella: Nit picking here, but why wrap all of these Exceptions in RuntimeException if the method is throwing Exception anyways?

spring-xd-test/src/main/java/org/springframework/xd/test/kafka/KafkaTestSupport.java
```diff
@@ -49,81 +62,99 @@
...
+		else {
+			try {
+				this.zkClient = new ZkClient(DEFAULT_ZOOKEEPER_CONNECT, 5000, 5000, ZKStringSerializer$.MODULE$);
+				if (ZkUtils.getAllBrokersInCluster(zkClient).size() == 0) {
+					throw new RuntimeException("Kafka server not available on localhost");
+				}
+			}
+			catch (Exception e) {
+				throw new RuntimeException("Issues connecting to ZK server");
+			}
```
mminella: See comment above about wrapping exception (in fact this one doesn't even wrap the original exception).

spring-xd-spark-streaming-tests/src/test/java/org/springframework/xd/spark/streaming/KafkaTransportSparkStreamingTests.java
mbogoevici: I think that the original intent of the code was to control how the `KafkaTestSupport` `@ClassRule` rule is instantiated, and to retrieve its configuration parameters, so that they can be set as system properties. Done this way, the instantiation has to take place in a static block, because `@ClassRule`s are executed around the `@BeforeClass`/`@AfterClass` cycle.

After moving the instantiation to the field declaration, there is little reason to keep a static block, that is correct.

Also, looking again at this, I notice that we need to move this into a `@BeforeClass` block, for the embedded tests to work properly.

spring-xd-test/src/main/java/org/springframework/xd/test/kafka/KafkaTestSupport.java
mbogoevici: Not necessary to be runtime exceptions, it's just for adding the error message.

spring-xd-test/src/main/java/org/springframework/xd/test/kafka/KafkaTestSupport.java
mbogoevici: Actually, this can be removed

markpollack:
squashed and merged as 8cf436e9d45d9999d7ebd892edb529143380144d

PR #1454, opened 2015-02-09
XD-2707 Check the payload type for Message

ilayaperumalg:
- When the spark streaming RDD items are iterated, add a check to see if the
item is of type spring `Message` before we proceed with building one.

### DIFF
--- a/spring-xd-spark-streaming/src/main/java/org/springframework/xd/spark/streaming/java/ModuleExecutor.java
+++ b/spring-xd-spark-streaming/src/main/java/org/springframework/xd/spark/streaming/java/ModuleExecutor.java
@@ -25,6 +25,7 @@ import org.apache.spark.api.java.function.VoidFunction;
 import org.apache.spark.streaming.api.java.JavaDStreamLike;
 
 import org.springframework.beans.factory.NoSuchBeanDefinitionException;
+import org.springframework.messaging.Message;
 import org.springframework.messaging.support.MessageBuilder;
 import org.springframework.xd.spark.streaming.SparkMessageSender;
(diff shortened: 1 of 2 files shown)


spring-xd-spark-streaming/src/main/java/org/springframework/xd/spark/streaming/java/ModuleExecutor.java
```diff
@@ -72,7 +73,10 @@ public void call(Iterator<?> results) throws Exception {
 									}
 								}
 								while (results.hasNext()) {
-									messageSender.send(MessageBuilder.withPayload(results.next()).build());
+									Object next = results.next();
+									Message message = (next instanceof Message) ? (Message) next :
+											MessageBuilder.withPayload(next).build();
```
garyrussell: Not critical for this release but we should always use `getMessageBuilderFactory(beanFactory)` to get the `MessageBuilder` to use (not each time of course, during initialization).

This allows customization of the message builder globally.

markfisher: This particular use of MessageBuilder occurs on the worker node on the Spark cluster so it would not be within the same context as XD itself (on the container side). That said, we should consider whether that instance from the context could be serialized. Post 1.1 of course. @ilayaperumalg can you create a JIRA for that?

ilayaperumalg: https://jira.spring.io/browse/XD-2708

markfisher:
LGTM; rebased, squashed, and merged

PR #1459, opened 2015-02-11
XD-2710 Kafka on-demand queue tests should not assume offset 0

mbogoevici:
- record initial offsets at the beginning of the tests - this ensures that the reading happens after the test has begun writing of the topics;
- this accounts for the case where the topic exists at the beginning of the test and already contains data (typical for CI environment)


ilayaperumalg:
LGTM, merging!

ilayaperumalg:
Rebased and merged as 8598f48265c5b5e8fe744869b3ab1c5e7cd420da

PR #1460, opened 2015-02-11
Fix STS Indent Toggling

garyrussell:
Don't set both "format on save" and "correct indentation on save".

They have different rules and cause toggling of (somes) indents.

We will get some one-time adjustments, such as the RabbitMessageBus
(included in this commit).


ghillert:
LGTM - merging.

ghillert:
Merged.

PR #1461, opened 2015-02-12
ShellCommandProcessor handles quoted strings

nebhale:
Previously the `ShellCommandProcessor` did not handle quoted strings properly.  If you attempted to pass the string
```
bash -c 'while read LINE ; do echo $LINE ; done'
```

as a `--command`, the processor would naively parse the string into
```
bash
-c
'while
read
LINE
;
do
echo
$LINE
;
done'
```

when it should be parsed into

```
bash
-c
while read LINE ; do echo $LINE ; done
```

This change enhances the parsing algorithm to make it compatible with the [POSIX / SUSv3 standard][1].

[1]: http://pubs.opengroup.org/onlinepubs/009695399/toc.htm

### DIFF
--- a/extensions/spring-xd-extension-process/src/main/java/org/springframework/xd/extension/process/ShellCommandProcessor.java
+++ b/extensions/spring-xd-extension-process/src/main/java/org/springframework/xd/extension/process/ShellCommandProcessor.java
@@ -22,7 +22,6 @@ import java.io.IOException;
 import java.io.InputStream;
 import java.io.InputStreamReader;
 import java.io.OutputStream;
-import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
 import java.util.concurrent.atomic.AtomicBoolean;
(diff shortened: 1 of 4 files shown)


ericbottard:
Can you point us to the exact location of http://pubs.opengroup.org/onlinepubs/009695399/toc.htm where that algorithm appears?

trisberg:
My guess is that Ben is referring to this paragraph:

2.2.2 Single-Quotes

Enclosing characters in single-quotes ( '' ) shall preserve the literal value of each character within the single-quotes. A single-quote cannot occur within single-quotes.

dturanski:
Good start. I think we need to pull out the parser into its own class and add some more unit tests (that don't require ProcessBuilder).

ericbottard:
I'd like to understand how one goes from that a statement to such a crazy regex :)

If the regex appears verbatim in the doc, then that's fine with me

nebhale:
@ericbottard https://github.com/rubysl/rubysl-shellwords/blob/2.0/lib/rubysl/shellwords/shellwords.rb#L32-L45 is a highly regarded implementation of the spec.  As you can see, my change is simply a Java-ification of that.  If you want specs to go long with it (to Java-ify and include) you can find them [here](https://github.com/rubysl/rubysl-shellwords/blob/2.0/spec/shellwords_spec.rb).

nebhale:
I've updated to PR by splitting out the parser into a separate class with a port of the Ruby test suite.

dturanski:
LGTM; merged

PR #1464, opened 2015-02-18
[Sprint: 43] XD-2735 Make local messagebus props configurable

ilayaperumalg:
- Add property keys in servers.yml so that they can be configurable via servers.yml
 - Fix the prefix to align with other messagebus property keys.

### DIFF
--- a/spring-xd-dirt/src/main/resources/application.yml
+++ b/spring-xd-dirt/src/main/resources/application.yml
@@ -93,6 +93,8 @@ xd:
   messageRateMonitoring:
     enabled: false
   messagebus:
+    local:
+      polling:                     1000
     rabbit:
       compressionLevel:            1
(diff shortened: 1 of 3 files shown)


spring-xd-dirt/src/main/resources/application.yml
```diff
@@ -93,6 +93,8 @@ xd:
   messageRateMonitoring:
     enabled: false
   messagebus:
+    local:
+      polling:                     1000
```
markpollack: do you want to add queuesize here?

ilayaperumalg: I thought about it. Since it is the integer max, I kept it this way: `${xd.messagebus.local.queueSize: #{T(Integer).MAX_VALUE}}`

PR #1465, opened 2015-02-19
Fix spring-xd-exec task

ipsi:
The spring-xd-exec task was not doing what it was supposed to, so I've added the necessary code to fix it (for now - I imagine later releases may break it).

Also fixed up some issues with SLF4J complaining about both log4j-over-slf4j and slf4j-log4j12 being on the classpath, and Kafka using a different version of ZooKeeper to everything else.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -379,12 +379,15 @@ project('spring-xd-dirt') {
 		compile "org.springframework.boot:spring-boot-actuator"
 		compile("org.springframework.boot:spring-boot-starter-security") {
 			exclude group: 'org.springframework.boot', module: "spring-boot-starter-logging"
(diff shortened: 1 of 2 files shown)


ericbottard:
The exec task is a remnant from early integration with Cloud Foundry and is no longer really useful.
We'd be interested in understanding how you use it (if you do).

Also, could you split this PR in 3, one for each concern that it addresses (it appears the 2 others are indeed genuine, irrespective of what we decide with regard to the `exec` task).

ericbottard:
"exec" related stuff will be taken care of as part of https://github.com/spring-projects/spring-xd/pull/1496

PR #1466, opened 2015-02-23
XD-2723: Revert timeout to 1hr

ericbottard:


spring-xd-module-spi/src/main/java/org/springframework/xd/module/options/mixins/BatchJobSinglestepPartitionSupportOptionMixin.java
```diff
@@ -29,7 +29,7 @@
 @Mixin(BatchJobRestartableOptionMixin.class)
 public class BatchJobSinglestepPartitionSupportOptionMixin {
 
-	private long partitionResultsTimeout = 300000;
+	private long partitionResultsTimeout = 3_600_000; // 1 hour
```
mminella: Neat trick with the _ characters.  Never knew that :)

PR #1468, opened 2015-02-23
[Sprint: 44] XD-2741 - Redis sink default to spring.redis configuration in servers.yml

markpollack:


garyrussell:
Shouldn't we do this the same way as it's done in the [RabbitConnectionMixin](https://github.com/spring-projects/spring-xd/blob/master/spring-xd-dirt/src/main/java/org/springframework/xd/dirt/modules/metadata/RabbitConnectionMixin.java#L33) ?

Also, this is misnamed; it's 2741 not 2471.

markpollack:
I was following the pattern for jdbc I think.  There maybe some overall inconsistency.  I'll look into it more.

garyrussell:
Yes, I see - I have opened https://jira.spring.io/browse/XD-2768 to resolve the inconsistencies and will merge this as-is.

garyrussell:
Merged and cherry-picked to 1.1.x.

PR #1469, opened 2015-02-23
[Sprint: 44] XD-2719: Add Rabbit Bus Cleanup REST Method

garyrussell:
JIRA: https://jira.spring.io/browse/XD-2719

    curl -X DELETE 'http://localhost:9393/streams/clean/rabbit/foo\
         ?user=guest&pw=guest&vhost=/&busPrefix=xdbus.&adminUri=http://localhost:15672'

Query params are optional (defaults as shown).

Looks for queues named xdbus.<stream>.0, 1, 2, 3 etc and possibly their associated
.dlq companions.

Verifies that none of the queues are currently in use (but does not check if they are empty.

If no queues have consumers, all that were found are deleted.

### DIFF
--- a/spring-xd-dirt/src/main/java/org/springframework/xd/dirt/integration/bus/RabbitAdminException.java
+++ b/spring-xd-dirt/src/main/java/org/springframework/xd/dirt/integration/bus/RabbitAdminException.java
@@ -1,0 +1,36 @@
+/*
+ * Copyright 2015 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.xd.dirt.integration.bus;
+
+import org.springframework.xd.dirt.XDRuntimeException;
+
+
+/**
+ *
+ * @author Gary Russell
+ */
+public class RabbitAdminException extends XDRuntimeException {
+
+	public RabbitAdminException(String message, Throwable cause) {
+		super(message, cause);
+	}
+
+	public RabbitAdminException(String message) {
+		super(message);
+	}
+
+}
--- a/spring-xd-dirt/src/main/java/org/springframework/xd/dirt/integration/bus/RabbitBusCleaner.java
+++ b/spring-xd-dirt/src/main/java/org/springframework/xd/dirt/integration/bus/RabbitBusCleaner.java
@@ -1,0 +1,150 @@
+/*
+ * Copyright 2015 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.xd.dirt.integration.bus;
+
+import java.net.URI;
+import java.net.URISyntaxException;
+import java.util.ArrayList;
+import java.util.Collections;
+import java.util.List;
+import java.util.Map;
+
+import org.apache.commons.logging.Log;
+import org.apache.commons.logging.LogFactory;
+import org.apache.http.HttpHost;
+import org.apache.http.auth.AuthScope;
+import org.apache.http.auth.UsernamePasswordCredentials;
+import org.apache.http.client.AuthCache;
+import org.apache.http.client.HttpClient;
+import org.apache.http.client.protocol.HttpClientContext;
+import org.apache.http.impl.auth.BasicScheme;
+import org.apache.http.impl.client.BasicAuthCache;
+import org.apache.http.impl.client.BasicCredentialsProvider;
+import org.apache.http.impl.client.HttpClients;
+import org.apache.http.protocol.HttpContext;
+
+import org.springframework.http.HttpMethod;
+import org.springframework.http.HttpStatus;
+import org.springframework.http.client.HttpComponentsClientHttpRequestFactory;
+import org.springframework.http.converter.HttpMessageConverter;
+import org.springframework.http.converter.json.MappingJackson2HttpMessageConverter;
+import org.springframework.web.client.HttpClientErrorException;
+import org.springframework.web.client.RestTemplate;
+import org.springframework.web.util.UriComponentsBuilder;
+
+
+/**
+ *
+ * @author Gary Russell
+ */
+public class RabbitBusCleaner implements BusCleaner {
+
+	final static Log logger = LogFactory.getLog(RabbitBusCleaner.class);
+
+	@Override
+	public List<String> clean(String stream) {
+		return clean("http://localhost:15672", "guest", "guest", "/", "xdbus.", stream);
+	}
+
+	@SuppressWarnings("unchecked")
+	public List<String> clean(String adminUri, String user, String pw, String vhost,
+			String busPrefix, String stream) {
+		List<String> results = new ArrayList<>();
+		RestTemplate restTemplate = buildRestTemplate(adminUri, user, pw);
+		int n = 0;
+		while (true) {
+			String queueName = busPrefix + stream + "." + n++;
+			URI uri = UriComponentsBuilder.fromUriString(adminUri + "/api").pathSegment("queues", "{vhost}", "{stream}")
+					.buildAndExpand(vhost, queueName).encode().toUri();
+			try {
+				Map<String, Object> queue = restTemplate.getForObject(uri, Map.class);
+				if (queue.get("consumers") != Integer.valueOf(0)) {
+					throw new RabbitAdminException("Queue " + queueName + " is in use");
+				}
+				results.add(queueName);
+				queueName += ".dlq";
+				uri = UriComponentsBuilder.fromUriString(adminUri + "/api").pathSegment("queues", "{vhost}", "{stream}")
+						.buildAndExpand(vhost, queueName).encode().toUri();
+				try {
+					queue = restTemplate.getForObject(uri, Map.class);
+					if (queue.get("consumers") != Integer.valueOf(0)) {
+						throw new RabbitAdminException("Queue " + queueName + " is in use");
+					}
+					results.add(queueName);
+				}
+				catch (HttpClientErrorException e) {
+					if (e.getStatusCode().equals(HttpStatus.NOT_FOUND)) {
+						continue; // DLQs are not mandatory
+					}
+					throw new RabbitAdminException("Failed to lookup queue", e);
+				}
+			}
+			catch (HttpClientErrorException e) {
+				if (e.getStatusCode().equals(HttpStatus.NOT_FOUND)) {
+					break; // No more for this stream
+				}
+				throw new RabbitAdminException("Failed to lookup queue", e);
+			}
+		}
+		for (String queueName : results) {
+			URI uri = UriComponentsBuilder.fromUriString(adminUri + "/api").pathSegment("queues", "{vhost}", "{stream}")
+					.buildAndExpand(vhost, queueName).encode().toUri();
+			restTemplate.delete(uri);
+			if (logger.isDebugEnabled()) {
+				logger.debug("deleted queue: " + queueName);
+			}
+		}
+		return results;
+	}
+
+	public static RestTemplate buildRestTemplate(String adminUri, String user, String password) {
+		BasicCredentialsProvider credsProvider = new BasicCredentialsProvider();
+		credsProvider.setCredentials(
+				new AuthScope(AuthScope.ANY_HOST, AuthScope.ANY_PORT),
+				new UsernamePasswordCredentials(user, password));
+		HttpClient httpClient = HttpClients.custom().setDefaultCredentialsProvider(credsProvider).build();
+		// Set up pre-emptive basic Auth because the rabbit plugin doesn't currently support challenge/response for PUT
+		// Create AuthCache instance
+		AuthCache authCache = new BasicAuthCache();
+		// Generate BASIC scheme object and add it to the local; from the apache docs...
+		// auth cache
+		BasicScheme basicAuth = new BasicScheme();
+		URI uri;
+		try {
+			uri = new URI(adminUri);
+		}
+		catch (URISyntaxException e) {
+			throw new RabbitAdminException("Invalid URI", e);
+		}
+		authCache.put(new HttpHost(uri.getHost(), uri.getPort(), uri.getScheme()), basicAuth);
+		// Add AuthCache to the execution context
+		final HttpClientContext localContext = HttpClientContext.create();
+		localContext.setAuthCache(authCache);
+		RestTemplate restTemplate = new RestTemplate(new HttpComponentsClientHttpRequestFactory(httpClient) {
+
+			@Override
+			protected HttpContext createHttpContext(HttpMethod httpMethod, URI uri) {
+				return localContext;
+			}
+
+		});
+		restTemplate.setMessageConverters(Collections.<HttpMessageConverter<?>> singletonList(
+				new MappingJackson2HttpMessageConverter()));
+		return restTemplate;
+	}
+
+}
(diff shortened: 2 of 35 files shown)


spring-xd-dirt/src/main/java/org/springframework/xd/dirt/rest/StreamsController.java
```diff
@@ -67,4 +75,27 @@ public StreamsController(StreamDeployer streamDeployer, StreamDefinitionReposito
 	protected StreamDefinition createDefinition(String name, String definition) {
 		return new StreamDefinition(name, definition);
 	}
+
+	@ResponseBody
+	@RequestMapping(value = "/clean/{transport}/{stream}")
```
ericbottard: This is a mutative operation, so should not use GET IMO.

spring-xd-dirt/src/main/java/org/springframework/xd/dirt/rest/StreamsController.java
```diff
@@ -67,4 +75,27 @@ public StreamsController(StreamDeployer streamDeployer, StreamDefinitionReposito
...
+	@ResponseStatus(HttpStatus.OK)
+	public List<String> clean(@PathVariable String transport, @PathVariable String stream,
+			@RequestParam(required = false) String adminUri,
+			@RequestParam(required = false) String user,
+			@RequestParam(required = false) String pw,
+			@RequestParam(required = false) String vhost,
+			@RequestParam(required = false) String busPrefix) {
+		Assert.isTrue("rabbit".equals(transport), "Only rabbit transports can be cleaned");
+		String adminUri2 = adminUri == null ? "http://localhost:15672" : adminUri;
```
ericbottard: All those defaults should not reside in the Controller layer, but rather _e.g._ in the cleaner itself

spring-xd-dirt/src/main/java/org/springframework/xd/dirt/rest/StreamsController.java
```diff
@@ -67,4 +75,27 @@ public StreamsController(StreamDeployer streamDeployer, StreamDefinitionReposito
...
+		String user2 = user == null ? "guest" : user;
+		String pw2 = pw == null ? "guest" : pw;
+		String vhost2 = vhost == null ? "/" : vhost;
+		String busPrefix2 = busPrefix == null ? "xdbus." : busPrefix;
+		List<String> results = busCleaner.clean(adminUri2, user2, pw2, vhost2, busPrefix2, stream);
+		if (results.size() == 0) {
+			results.add("No queues named " + busPrefix2 + stream + ".* found");
```
ericbottard: Should not convey this kind of result the same way a normal result is conveyed. Different HTTP code seems like a better option

spring-xd-dirt/src/main/java/org/springframework/xd/dirt/integration/bus/RabbitBusCleaner.java
```diff
@@ -0,0 +1,150 @@
...
+					.buildAndExpand(vhost, queueName).encode().toUri();
+			restTemplate.delete(uri);
+			if (logger.isDebugEnabled()) {
+				logger.debug("deleted queue: " + queueName);
+			}
+		}
+		return results;
+	}
+
+	public static RestTemplate buildRestTemplate(String adminUri, String user, String password) {
```
ericbottard: Should not be public

spring-xd-dirt/src/main/java/org/springframework/xd/dirt/rest/StreamsController.java
artembilan: I guess `DELETE` is the best choice for `clean`.
But does `curl` support such a HTTP method ?

spring-xd-dirt/src/main/java/org/springframework/xd/dirt/integration/bus/RabbitBusCleaner.java
```diff
@@ -0,0 +1,150 @@
...
+		return clean("http://localhost:15672", "guest", "guest", "/", "xdbus.", stream);
+	}
+
+	@SuppressWarnings("unchecked")
+	public List<String> clean(String adminUri, String user, String pw, String vhost,
+			String busPrefix, String stream) {
+		List<String> results = new ArrayList<>();
+		RestTemplate restTemplate = buildRestTemplate(adminUri, user, pw);
+		int n = 0;
+		while (true) {
```
ericbottard: Aren't we afraid of staying trapped in this while loop? Maybe use Spring Retry?

spring-xd-dirt/src/main/java/org/springframework/xd/dirt/integration/bus/RabbitBusCleaner.java
```diff
@@ -0,0 +1,150 @@
...
+import org.springframework.http.client.HttpComponentsClientHttpRequestFactory;
+import org.springframework.http.converter.HttpMessageConverter;
+import org.springframework.http.converter.json.MappingJackson2HttpMessageConverter;
+import org.springframework.web.client.HttpClientErrorException;
+import org.springframework.web.client.RestTemplate;
+import org.springframework.web.util.UriComponentsBuilder;
+
+
+/**
+ *
```
ericbottard: Needs javadoc

spring-xd-dirt/src/main/java/org/springframework/xd/dirt/integration/bus/RabbitBusCleaner.java
```diff
@@ -0,0 +1,150 @@
...
+ * @author Gary Russell
+ */
+public class RabbitBusCleaner implements BusCleaner {
+
+	final static Log logger = LogFactory.getLog(RabbitBusCleaner.class);
```
artembilan: `private`.
And looks like there is need `@since 1.2` for all new classes

spring-xd-dirt/src/main/java/org/springframework/xd/dirt/integration/bus/RabbitAdminException.java
```diff
@@ -0,0 +1,36 @@
...
+ * limitations under the License.
+ */
+
+package org.springframework.xd.dirt.integration.bus;
+
+import org.springframework.xd.dirt.XDRuntimeException;
+
+
+/**
+ *
```
ericbottard: Needs javadoc

spring-xd-shell/src/test/java/org/springframework/xd/shell/bus/RabbitBusCleanerTests.java
```diff
@@ -0,0 +1,127 @@
...
+		connectionFactory.destroy();
+		List<String> cleaned = cleaner.clean(uuid);
+		assertEquals(10, cleaned.size());
+		for (int i = 0; i < 5; i++) {
+			assertEquals("xdbus." + uuid + "." + i, cleaned.get(i * 2));
+			assertEquals("xdbus." + uuid + "." + i + ".dlq", cleaned.get(i * 2 + 1));
+		}
+	}
+
+	public class Queue {
```
ericbottard: Should be static I believe

spring-xd-dirt/src/main/java/org/springframework/xd/dirt/rest/StreamsController.java
```diff
@@ -67,4 +75,27 @@ public StreamsController(StreamDeployer streamDeployer, StreamDefinitionReposito
 	protected StreamDefinition createDefinition(String name, String definition) {
 		return new StreamDefinition(name, definition);
 	}
+
+	@ResponseBody
+	@RequestMapping(value = "/clean/{transport}/{stream}")
+	@ResponseStatus(HttpStatus.OK)
+	public List<String> clean(@PathVariable String transport, @PathVariable String stream,
+			@RequestParam(required = false) String adminUri,
```
ericbottard: Given the specific options that are here, I believe we should have a dedicated endpoint (ie `"/clean/rabbit/{stream}"`) as this won't accommodate other transports anyway

spring-xd-dirt/src/main/java/org/springframework/xd/dirt/integration/bus/RabbitBusCleaner.java
```diff
@@ -0,0 +1,150 @@
...
+				if (e.getStatusCode().equals(HttpStatus.NOT_FOUND)) {
+					break; // No more for this stream
+				}
+				throw new RabbitAdminException("Failed to lookup queue", e);
+			}
+		}
+		for (String queueName : results) {
+			URI uri = UriComponentsBuilder.fromUriString(adminUri + "/api").pathSegment("queues", "{vhost}", "{stream}")
+					.buildAndExpand(vhost, queueName).encode().toUri();
+			restTemplate.delete(uri);
```
ericbottard: Shouldn't we check the result of this?

spring-xd-dirt/src/main/java/org/springframework/xd/dirt/integration/bus/RabbitBusCleaner.java
```diff
@@ -0,0 +1,150 @@
...
+	}
+
+	@SuppressWarnings("unchecked")
+	public List<String> clean(String adminUri, String user, String pw, String vhost,
+			String busPrefix, String stream) {
+		List<String> results = new ArrayList<>();
+		RestTemplate restTemplate = buildRestTemplate(adminUri, user, pw);
+		int n = 0;
+		while (true) {
+			String queueName = busPrefix + stream + "." + n++;
```
ericbottard: This is a big violation of DRY. Can't we hook this into the  MessageBus class somehow?

spring-xd-shell/src/test/java/org/springframework/xd/shell/bus/RabbitBusCleanerTests.java
```diff
@@ -0,0 +1,127 @@
...
+import com.rabbitmq.client.DefaultConsumer;
+
+/**
+ *
+ * @author Gary Russell
+ */
+public class RabbitBusCleanerTests {
+
+	@Test
+	@Ignore
```
ericbottard: Is this ignored for the reason explained in the comment below? If so, can we create a `@Rule` for that?

spring-xd-shell/src/test/java/org/springframework/xd/shell/bus/RabbitBusCleanerTests.java
```diff
@@ -0,0 +1,127 @@
...
+			template.put(uri, new Queue(false, true));
+		}
+		CachingConnectionFactory connectionFactory = new CachingConnectionFactory("localhost");
+		new RabbitTemplate(connectionFactory).execute(new ChannelCallback<Void>() {
+
+			@Override
+			public Void doInRabbit(Channel channel) throws Exception {
+				String consumerTag = channel.basicConsume("xdbus." + uuid + ".4", new DefaultConsumer(channel));
+				try {
+					Thread.sleep(5000); // wait for consumer to show up in rest API
```
ericbottard: Can we poll until the condition is met? This test adds 10s of build time on its own :(

spring-xd-dirt/src/main/java/org/springframework/xd/dirt/integration/bus/RabbitBusCleaner.java
garyrussell: I needed it for the test case; I guess I could copy the code or use the google `@VisibleForTesting` annotation.

spring-xd-dirt/src/main/java/org/springframework/xd/dirt/integration/bus/RabbitBusCleaner.java
garyrussell: For what? The format of the queue name or the busPrefix? Different containers can have different bus prefixes so I left it to the user to supply the bus prefix. I guess we could create some helper function `SomeUtils.formatQueueName(busPrefix, stream, index)` and invoke it from both places.

spring-xd-dirt/src/main/java/org/springframework/xd/dirt/integration/bus/RabbitBusCleaner.java
ericbottard: Can it be made package private?

spring-xd-dirt/src/main/java/org/springframework/xd/dirt/integration/bus/RabbitBusCleaner.java
garyrussell: It will throw an exception if there's a problem.

spring-xd-dirt/src/main/java/org/springframework/xd/dirt/rest/StreamsController.java
garyrussell: good point

spring-xd-dirt/src/main/java/org/springframework/xd/dirt/rest/StreamsController.java
garyrussell: :+1:

spring-xd-dirt/src/main/java/org/springframework/xd/dirt/integration/bus/RabbitBusCleaner.java
ericbottard: The format of the queue name. Are you saying that for a single XD install (using rabbit as a transport), there could be two different naming schemes?

spring-xd-shell/src/test/java/org/springframework/xd/shell/bus/RabbitBusCleanerTests.java
garyrussell: Yeah - the funky formatting forced the comment to the next line; I'll add the rule.

spring-xd-shell/src/test/java/org/springframework/xd/shell/bus/RabbitBusCleanerTests.java
garyrussell: Another reason for the ignore :smile: 

Will do.

spring-xd-shell/src/test/java/org/springframework/xd/shell/bus/RabbitBusCleanerTests.java
ericbottard: Great!

For the record, the `@Ignore` annotation has a `value` attribute that can be used to convey just about anything you want.

spring-xd-dirt/src/main/java/org/springframework/xd/dirt/rest/StreamsController.java
garyrussell: Yeah, but we might get some args but not all - moving to the cleaner.

spring-xd-dirt/src/main/java/org/springframework/xd/dirt/integration/bus/RabbitBusCleaner.java
garyrussell: That's how I had it originally, but I had to move the test to xd-shell because dirt no longer has rabbit on the cp; I guess I could put the test in the same package there, but it makes me feel a little queasy using a dirt package there.

spring-xd-dirt/src/main/java/org/springframework/xd/dirt/integration/bus/RabbitBusCleaner.java
garyrussell: Yes, the bus prefix is in the servers.yml so different container groups could use different queue prefixes. (Not sure why they would, but certainly possible - I'll add a utility method).

garyrussell: Well, it's not so simple, the bus knows nothing about the `.0` part - that's created by the stream plugin. So now this utility method would simply concatenate the prefix + name; so it's not clear to me how useful that would be. So I'm leaving as-is unless you have other suggestions.

spring-xd-dirt/src/main/java/org/springframework/xd/dirt/integration/bus/RabbitBusCleaner.java
garyrussell: Hmmm - looks like there __are__ rabbit tests in dirt; I'll move it back there and change the visibility

spring-xd-dirt/src/main/java/org/springframework/xd/dirt/integration/bus/RabbitBusCleaner.java
ericbottard: Maybe a utility method on the stream plugin then? My point was that the format specifier for that name should appear only once in the codebase, ideally.

garyrussell: Unfortunately, the stream plugin "knows" about the ".0" part, but only the bus knows about the prefix part. I'll give it some more thought.

garyrussell:
All good points; pushed; added integration test.

garyrussell:
Pushed support for cleaning up rabbit remnants for jobs (job channel and event tap exchanges).

    curl -X DELETE localhost:9393/jobs/clean/rabbit/jobxxx

ericbottard:
Rebased, squashed and merged as 2d5f3f706330a6ead8e91c9a7a23d4372715614d

PR #1470, opened 2015-02-24
[Sprint: 45] XD-2287: Centrally shared ModuleRegistry

ericbottard:
This PR does the following:
* Rework ResourceModuleRegistry so that it is not tied to `java.io.File` (contrary to `ArchiveModuleRegistry`, which is removed in this PR)
* Add md5 file generation for uploaded modules
* Introduce `SynchronizingModuleRegistry` which acts as a registry but knows about two delegates: the source (remote) and the target (local). It will do as if it queries the source registry (_e.g._ `hdfs:`), but fetch the module bits back to the target registry (which is meant to use `file:`)
* Replace the registry that stores custom modules with such an instance if needs be


spring-xd-dirt/src/main/java/org/springframework/xd/dirt/module/ExtendedResource.java
```diff
@@ -0,0 +1,124 @@
...
+import org.springframework.core.io.Resource;
+import org.springframework.util.FileSystemUtils;
+import org.springframework.util.ReflectionUtils;
+
+/**
+ * A wrapper around a Resource that adds capabilities lacking in the original interface.
+ *
+ * @author Eric Bottard
+ */
+public abstract class ExtendedResource {
```
dturanski: Might want to add HDFS to the class name since it has a hard dependency

spring-xd-dirt/src/main/java/org/springframework/xd/dirt/module/ExtendedResource.java
```diff
@@ -0,0 +1,124 @@
...
+
+	private static final Class<?> HDFS_RESOURCE;
+
+	static {
+		try {
+			HDFS_RESOURCE = Class.forName("org.springframework.data.hadoop.fs.HdfsResource");
```
dturanski: Or encapsulate this stuff in HdfsExtendedResource

spring-xd-dirt/src/main/java/org/springframework/xd/dirt/module/ExtendedResource.java
```diff
@@ -0,0 +1,124 @@
...
+
+		private static final Field PATH_FIELD;
+
+		private static final Field FS_FIELD;
+
+		static {
+			PATH_FIELD = ReflectionUtils.findField(HDFS_RESOURCE, "path");
+			ReflectionUtils.makeAccessible(PATH_FIELD);
+			FS_FIELD = ReflectionUtils.findField(HDFS_RESOURCE, "fs");
+			ReflectionUtils.makeAccessible(FS_FIELD);
```
dturanski: This is brittle. Should we consider adding public accessors in spring data hadoop?

spring-xd-dirt/src/main/java/org/springframework/xd/dirt/module/StalenessCheck.java
```diff
@@ -0,0 +1,35 @@
...
+ *
+ */
+
+package org.springframework.xd.dirt.module;
+
+import org.springframework.xd.module.SimpleModuleDefinition;
+
+/**
+ * A strategy interface for deciding whether a terget definition is stale compared to a source definition.
+ *
```
dturanski: target

dturanski:
General question. The standalone module projects exclude spring-data-hadoop. Does this create a hard runtime dependency even for single node?

spring-xd-dirt/src/main/java/org/springframework/xd/dirt/module/ResourceModuleRegistry.java
```diff
@@ -16,25 +16,319 @@
...
+	public static final String ARCHIVE_AS_FILE_EXTENSION = ".jar";
+
+	/**
+	 * The extension to use for storing an uploaded module hash.
+	 */
+	public static final String HASH_EXTENSION = ".md5";
+
+	private final static String[] SUFFIXES = new String[] {"", ARCHIVE_AS_FILE_EXTENSION};
+
+	final protected static byte[] hexArray = "0123456789ABCDEF".getBytes();
```
dturanski: Should this be all caps by convention?

spring-xd-dirt/src/main/java/org/springframework/xd/dirt/module/ExtendedResource.java
```diff
@@ -0,0 +1,124 @@
...
+	/**
+	 * Makes sure that the resource and all its parent "directories" exist, creating them if necessary.
+	 */
+	public abstract boolean mkdirs() throws IOException;
+
+	public static ExtendedResource wrap(Resource original) {
+		if (original instanceof FileSystemResource) {
+			return new FileSystemExtendedResource((FileSystemResource) original);
+		}
+		else if (original.getClass().equals(HDFS_RESOURCE)) {
```
ericbottard: It's needed here, sadly.
But the ExtendedResource itself is not "Hdfs" centric so...

We could also pay Janne a bear so that he makes some members of HdfsResource public..

spring-xd-dirt/src/main/java/org/springframework/xd/dirt/module/ResourceModuleRegistry.java
ericbottard: Correct

spring-xd-dirt/src/main/java/org/springframework/xd/dirt/module/ResourceModuleRegistry.java
```diff
@@ -16,25 +16,319 @@
...
+
+	private String root;
 
 	public ResourceModuleRegistry(String root) {
-		super(root);
+		this(root, false);
+	}
+
+	public ResourceModuleRegistry(String root, boolean allowWrite) {
+		this.allowWrite = this.createDirectoryStructure = this.requireHashFiles = allowWrite;
```
dturanski: This many conditionals usually indicates refactoring to a sub class; possibly a factory pattern.

ericbottard: Hmmm, not sure I follow.

These are 3 concerns, with this constructor providing a quick and easy way of creating a "write enabled" registry with sensible defaults for writing. What subclass(es) would you see here?

spring-xd-dirt/src/main/java/org/springframework/xd/dirt/module/SynchronizingModuleRegistry.java
```diff
@@ -0,0 +1,151 @@
...
+		SimpleModuleDefinition targetDefinition = (SimpleModuleDefinition) currentTargetDefinition;
+		if (staleness.isStale(targetDefinition, simpleRemoteDefinition)) {
+			InputStream is = null;
+			try {
+				is = resolver.getResource(simpleRemoteDefinition.getLocation()).getInputStream();
+			}
+			catch (IOException e) {
+				throw new RuntimeIOException("Error while copying module", e);
+			}
+			UploadedModuleDefinition vehicle = new UploadedModuleDefinition(remoteDefinition.getName(), remoteDefinition.getType(), is);
```
dturanski: `vehicle` is confusing to me in this context

ericbottard:
> General question. The standalone module projects exclude spring-data-hadoop. Does this create a hard runtime dependency even for single node?

I believe so. Maybe we could prevent the __hard__ dependency by hardcoding the HdfsResource class name. I'd really like to see this class expose its path and filesystem though.

spring-xd-dirt/src/main/java/org/springframework/xd/dirt/module/SynchronizingModuleRegistry.java
ericbottard: `toTransfer` ?
Had a hard time coming with a name..

spring-xd-dirt/src/main/java/org/springframework/xd/dirt/module/ResourceModuleRegistry.java
```diff
@@ -16,25 +16,319 @@
...
+
+	}
+
+	public void setRequireHashFiles(boolean requireHashFiles) {
+		this.requireHashFiles = requireHashFiles;
+	}
+
+	public void setCreateDirectoryStructure(boolean createDirectoryStructure) {
+		this.createDirectoryStructure = createDirectoryStructure;
+	}
```
dturanski: Do we really need to make these properties independently configurable? Seems like it could lead to some inconsistencies. I don't see these setters used anywhere.

spring-xd-dirt/src/main/java/org/springframework/xd/dirt/module/SynchronizingModuleRegistry.java
```diff
@@ -0,0 +1,151 @@
...
+ * its source and the {@code file://} one as its target.</p>
+ *
+ * @author Eric Bottard
+ */
+public class SynchronizingModuleRegistry implements WritableModuleRegistry {
+
+	private final WritableModuleRegistry sourceRegistry;
+
+	private final WritableModuleRegistry targetRegistry;
+
```
dturanski: Seems like `localRegistry` and `remoteRegistry` would be more accurate names here.

spring-xd-dirt/src/test/java/org/springframework/xd/dirt/module/ResourceModuleRegistryLogicTests.java
```diff
@@ -18,26 +18,33 @@
 
 package org.springframework.xd.dirt.module;
 
-import static org.hamcrest.Matchers.contains;
-import static org.junit.Assert.assertThat;
+import static org.hamcrest.Matchers.*;
+import static org.junit.Assert.*;
+import static org.springframework.xd.module.ModuleType.*;
```
dturanski: I recently discovered a setting in Idea to disable wild card static imports.

spring-xd-dirt/src/main/java/org/springframework/xd/dirt/module/ResourceModuleRegistry.java
ericbottard: This relates to my comment above. I chose to allow independent configuration for cases where we don't have rights to the FS, or there is a 'virtual' FS.
For example, one could imagine a http remote server where "creating" http://foo/source does not make sense but which accepts the payload for an uploaded module.

The requireHashFiles setting is 100% orthogonal though, unless I'm mistaken (but we'll typically want it in a distributed scenario)

spring-xd-dirt/src/main/java/org/springframework/xd/dirt/module/SynchronizingModuleRegistry.java
ericbottard: Yeah, I've been using local and remote in some places. But note that they could both be "local" (e.g. different paths of the same (main) filesystem, as the unit tests do).

spring-xd-dirt/src/test/java/org/springframework/xd/dirt/module/ResourceModuleRegistryLogicTests.java
ericbottard: If I'm not mistaken (IDEA prefs being such a mess to understand, at least for me), I'm using the Spring Core settings.
Which we should put under git somehow, by the way, so that we all use the same.

spring-xd-dirt/src/main/java/org/springframework/xd/dirt/module/SynchronizingModuleRegistry.java
```diff
@@ -0,0 +1,151 @@
...
+		return result;
+	}
+
+
+	/**
+	 * Checks whether the target version of the definition is stale, and if it is, triggers an update (as a deletion then
+	 * new registration).
+	 */
+	private synchronized void copyIfStale(ModuleDefinition remoteDefinition) {
+		SimpleModuleDefinition simpleRemoteDefinition = (SimpleModuleDefinition) remoteDefinition;
```
markfisher: we either need asserts prior to these casts or have the specific type as the method param here; too brittle otherwise

spring-xd-dirt/src/main/java/org/springframework/xd/dirt/module/ResourceModuleRegistry.java
```diff
@@ -16,25 +16,319 @@
...
+
+	public ResourceModuleRegistry(String root, boolean allowWrite) {
+		this.allowWrite = this.createDirectoryStructure = this.requireHashFiles = allowWrite;
+		this.root = StringUtils.trimTrailingCharacter(root, '/');
+	}
+
+	@Override
+	public boolean delete(ModuleDefinition definition) {
+		if (!allowWrite) {
```
markfisher: It seems like this should throw an UnsupportedOperationException

ericbottard: Actually not. Returning `false` is a signal to `DelegatingModuleRegistry` to move to the next candidate

spring-xd-dirt/src/main/java/org/springframework/xd/dirt/module/SynchronizingModuleRegistry.java
ericbottard: Will move the cast to the caller 14 lines above. `!isComposed()` protects us here

spring-xd-dirt/src/main/java/org/springframework/xd/dirt/module/ResourceModuleRegistry.java
```diff
@@ -16,25 +16,319 @@
...
+		catch (IllegalArgumentException e) {
+			// Not an actual type name, skip
+			return;
+		}
+		String locationToUse = resource.getURL().toString();
+
+		ModuleDefinition found = ModuleDefinitions.simple(name, type, locationToUse);
+
+		if (holder.contains(found)) {
+			SimpleModuleDefinition one = (SimpleModuleDefinition) found;
```
markfisher: all of these unprotected downcasts are a code smell; looks like we need to reconsider what we expose on the interface

spring-xd-dirt/src/main/java/org/springframework/xd/dirt/module/CustomModuleRegistryFactoryBean.java
```diff
@@ -0,0 +1,82 @@
...
+	public boolean isSingleton() {
+		return true;
+	}
+
+	@Override
+	public void afterPropertiesSet() throws Exception {
+		Matcher matcher = NO_SYNCHRONIZATION_PATTERN.matcher(root);
+		if (matcher.matches()) {
+			registry = new ResourceModuleRegistry(root, true);
+			((ResourceModuleRegistry)registry).afterPropertiesSet();
```
markfisher: Why not change the instance type to ResourceModuleRegistry?

spring-xd-dirt/src/main/java/org/springframework/xd/dirt/module/CustomModuleRegistryFactoryBean.java
```diff
@@ -0,0 +1,82 @@
...
+			logger.info("Custom modules will be written directly to {}", root);
+		} else {
```
markfisher: should be a newline for else

spring-xd-dirt/src/main/java/org/springframework/xd/dirt/module/ResourceModuleRegistry.java
ericbottard: The list will only contain SimpleModuleDefinitions (and we're responsible for its creation). But in java, a `List<Goose>` is not a `List<Animal>`.

spring-xd-dirt/src/main/java/org/springframework/xd/dirt/module/ExtendedResource.java
markfisher: not sure if it's relevant, but Spring's Hadoop support includes an HdfsResource (not currently public). Could that be useful here, even if for delegation/composition?

spring-xd-dirt/src/main/java/org/springframework/xd/dirt/module/ExtendedResource.java
```diff
@@ -0,0 +1,124 @@
...
+public abstract class ExtendedResource {
+
+	private static final Class<?> HDFS_RESOURCE;
+
+	static {
+		try {
+			HDFS_RESOURCE = Class.forName("org.springframework.data.hadoop.fs.HdfsResource");
+		}
+		catch (ClassNotFoundException e) {
+			throw new IllegalStateException(e);
```
markfisher: should add some context in the exception message here

spring-xd-dirt/src/main/java/org/springframework/xd/dirt/module/MD5StalenessCheck.java
```diff
@@ -0,0 +1,71 @@
+/*
+ *
+ *  * Copyright 2011-2014 the original author or authors.
```
markfisher: Copyright range should include the year the class was introduced and any subsequent years. Was this copied from code that originated in 2011, and if not, what's the rationale?

Also, the formatting looks off here with double asterisks per line.

spring-xd-dirt/src/main/java/org/springframework/xd/dirt/module/MD5StalenessCheck.java
```diff
@@ -0,0 +1,71 @@
...
+import org.springframework.util.FileCopyUtils;
+import org.springframework.xd.dirt.core.RuntimeIOException;
+import org.springframework.xd.module.SimpleModuleDefinition;
+
+/**
+ * A {@link org.springframework.xd.dirt.module.StalenessCheck} that expects to find a {@code .md5} file next to a
+ * module archive and returns true if the computed hashes differ.
+ *
+ * @see org.springframework.xd.dirt.module.ResourceModuleRegistry
+ * @author Eric Bottard
```
markfisher: should be using "since" attributes for any new classes/interfaces

spring-xd-dirt/src/main/java/org/springframework/xd/dirt/module/MD5StalenessCheck.java
ericbottard: IDEA playing tricks on me

spring-xd-dirt/src/main/java/org/springframework/xd/dirt/module/ResourceModuleRegistry.java
markfisher: That seems wrong if it means the boolean return is overloaded. Maybe DelegatingModuleRegistry should catch exceptions as well.

spring-xd-dirt/src/main/java/org/springframework/xd/dirt/module/SynchronizingModuleRegistry.java
markfisher: Still a code smell, if isComposed is interchangeable with an instanceof check.

spring-xd-dirt/src/main/java/org/springframework/xd/dirt/module/ResourceModuleRegistry.java
markfisher: Why don't the simple() and composed() methods in ModuleDefinitions have specific return types? That would help.

spring-xd-dirt/src/main/java/org/springframework/xd/dirt/module/ResourceModuleRegistry.java
ericbottard: Seems ok with the contract:

```
public interface WritableModuleRegistry extends ModuleRegistry {

	/**
	 * Attempt to delete the given definition, if that registry is responsible for it.
	 *
	 * @return true if this registry was responsible for the definition and it was actually deleted
	 */
	public boolean delete(ModuleDefinition definition);
```

ericbottard: Disclaimer: I wrote the contract :)

spring-xd-dirt/src/main/java/org/springframework/xd/dirt/module/SynchronizingModuleRegistry.java
ericbottard: We had this discussion way back then, and this is exactly what I said. You were against instanceof checks. But there are really only 3 ways to do this, IMHO:
1. instanceof checks at usage site
1.1. hiding the instanceof check behind a method (what we do here)
2. double dispatch (ie visitor), which seems overkill here as we only have 2 implementations
3. collection-style exposition, which is actually a variation of double dispatch in a way, ie

```
CompositeModuleDefinition {
Iterable<String> or Optional<String> getLocation() {return "none"}
}

SimpleModuleDefinition {
 Iterable<ModuleDefinition> getChildren() {return empty()}
}
```

spring-xd-dirt/src/main/java/org/springframework/xd/dirt/module/ResourceModuleRegistry.java
markfisher: well, I guess what I'm questioning is the contract then. Anytime a boolean is based on X and Y, it's worth questioning IMO.

ericbottard: Rewinding back a bit, this whole situation exists because RMR is a WritableMR. Could create a subclass (like we had before with ArchiveMR & WriteCapableAMR) to cover write behavior.

dturanski: This is where I was going with my comments as well. Rather than have 3 boolean flags, I think it's cleaner to create subclasses (maybe even more than 2 along with corresponding interfaces to address creating hashes, creating directories).

ericbottard: Ok, I'll create a subclass for writing.
Can also implement the "require" part in requireHashes though the staleness strategy I guess

spring-xd-dirt/src/main/java/org/springframework/xd/dirt/module/ExtendedResource.java
```diff
@@ -0,0 +1,125 @@
...
+public abstract class ExtendedResource {
+
+	private static final Class<?> HDFS_RESOURCE;
+
+	static {
+		try {
+			HDFS_RESOURCE = Class.forName("org.springframework.data.hadoop.fs.HdfsResource");
+		}
+		catch (ClassNotFoundException e) {
+			throw new IllegalStateException("Spring Data Hadoop not found on the classpath, but was expected", e);
```
dturanski: This breaks with SingleNodeApplication tests in module projects as they currently do not import any hadoop dependent libs. It may be easiest to optionally disable this registry in SingleNodeIntegrationTestSupport. Can we do anything here?

ericbottard: One way to do it would be override the bean definition(s) for the registry(ies) in the test.

I wonder if removing the hard class dependency (using classnames) would be enough though

spring-xd-dirt/src/main/java/org/springframework/xd/dirt/module/ExtendedResource.java
ericbottard: And by "bear", I meant "beer" of course. Just seeing that now, wonder if this explains why that solution hasn't been pursued yet :)

dturanski:
squashed and merged

PR #1471, opened 2015-02-24
[Sprint: 44] XD-2712 Default LocalMessageBus for singlenode

dturanski:


spring-xd-messagebus-local/src/main/resources/META-INF/spring-xd/bus/local-bus.xml
```diff
@@ -1,15 +1,17 @@
...
 
-	<bean id="messageBus" class="org.springframework.xd.dirt.integration.bus.local.LocalMessageBus">
-		<property name="queueSize"
-			value="${xd.local.transport.named.queueSize: #{T(Integer).MAX_VALUE}}" />
-		<property name="poller">
-			<int:poller fixed-rate="${xd.local.transport.named.polling:1000}" />
-		</property>
-	</bean>
-
+	<beans profile="singlenode">
```
ericbottard: Why the need of a profile if it's only gonna be included when transport == local anyway?

dturanski: True, I don't think we need it.

ericbottard: And then, the next question is: why was the addition of that profile to JobPluginTests.java needed

dturanski: That's removed in the latest commit

ericbottard:
Rebased and merged as 61060d157238329856f772730b2017cff93be009

PR #1474, opened 2015-02-25
[Sprint: 44] XD-2730: Allow deployment properties as a file reference

ericbottard:


spring-xd-shell/src/main/java/org/springframework/xd/shell/command/DeploymentOptionKeys.java
```diff
@@ -0,0 +1,28 @@
...
+
+/**
+ * Used to factor out names of the options that are shared by all deploy commands.
+ *
+ * @author Eric Bottard
+ */
+public class DeploymentOptionKeys {
+	static final String PROPERTIES_OPTION = "properties";
+
+	static final String PROPERTIES_FILE_OPTION = "propertiesFile";
```
garyrussell: Perhaps `propertiesFiles` 

    --propertiesFiles=foo.properties,myCommon.properties

??

spring-xd-shell/src/main/java/org/springframework/xd/shell/command/JobCommands.java
```diff
@@ -333,9 +341,29 @@ public String restartJobExecution(
...
+
+		int which = Assertions.atMostOneOf(PROPERTIES_OPTION, properties, PROPERTIES_FILE_OPTION, propertiesFile);
+		Map<String, String> propertiesToUse;
+		switch (which) {
+			case 0:
+				propertiesToUse = DeploymentPropertiesFormat.parseDeploymentProperties(properties);
+				break;
+			case 1:
+				Properties props = new Properties();
+				try (FileInputStream fis = new FileInputStream(propertiesFile)) {
```
garyrussell: `new PatternMatchingResourceResolvert().getResource(propertiesFile).getInputStream()`   ??

Would allow `file:`, `classpath:`, `http:`.

garyrussell:
Otherwise LGTM except a bunch of files need copyright updates.

ericbottard:
Both suggestions are interesting BUT 
1. we'd lose TAB completion for Files if we were to accept any resource (a String)
2. not sure `classpath:` makes much sense in this case (would mean the CP of the shell, which the user is not supposed to edit. And if he were, this is pretty equivalent to `file:`). So the question is: is there really any value for non-File protocol here?

I think we could work something out at shell level to keep TAB completion and conversion for a `Foo[]`.

To recap, 
1. needs Shell support, which I think we should add
2. may not be a good idea after all?

What do you think?

spring-xd-shell/src/main/java/org/springframework/xd/shell/command/JobCommands.java
```diff
@@ -333,9 +341,29 @@ public String restartJobExecution(
...
+					props.load(fis);
+				}
+				propertiesToUse = DeploymentPropertiesFormat.convert(props);
+				break;
+			default:
+				throw new AssertionError();
+		}
+
+		jobOperations().deploy(name, propertiesToUse);
+		return String.format("Deployed stream '%s'", name);
```
ericbottard: Whoops, copy/paste, should be "job"

ericbottard:
WRT copyright, what do you mean? Missing copyrights or need year update?
If the latter, shouldn't we do a full pass now and on Jan 1st?

garyrussell:
No problem; they were just ideas; I am ok with deferring 1 and dropping 2.

Regarding copyright, no, it's not a blanket update; we update the copyright if a file is __changed__ in a calendar year; 2014 becomes 2014-2015; 2013-2014 becomes 2013-2015.

I generally don't bother if it's an automated updated (e.g. a class moves and causes a bunch of import changes) but any changes of substance in a class should be updated.

ericbottard:
Rebased, updated copyrights.

ericbottard:
Created https://jira.spring.io/browse/SHL-173 to track requirement for 1.

I updated the copyright headers, so should be ready to merge if no further comment

garyrussell:
Squashed; merged as 550197a6b2488f6cab9cd1cf007779cbe858c3e0

ilayaperumalg:
It looks like we need to have this merged into 1.1.x as well. Re-opening.

ilayaperumalg:
Also, this change needs to be applied as well: c97ddba591a373c168e63a345b1c5897caa966ad

ericbottard:
This is a new feature, I don't see why it should go into the maintenance branch.

ilayaperumalg:
I see the JIRA https://jira.spring.io/browse/XD-2730 fix version says 1.1.x

ilayaperumalg:
@ericbottard I agree, I think it should be 1.2 then.

ericbottard:
I'll stick to what the JIRA God said, but seems odd to me

ilayaperumalg:
Closing this as this is in fact was intended for 1.2

PR #1478, opened 2015-02-26
[Sprint: 45] [Backport] XD-2761 fix kryo registration issue

dturanski:
This PR is for 1.1.x. Should be applied to the master as well. 

* PojoCodec does not register classes to Kryo inside ser/deser methods
* An extension to shared server context is enabled. The context imports from `classpath*:/META-INF/spring-xd/bus/ext/*.xml`.  This mechanism enables bean definitions in `codec.xml` (and other beans in the shared context technically) to be added/overriden. 
* A bean named `kryoRegistrar` may be registered and configured by the user if registration is necessary.
* PojoCodec is configured with a KryoRegistrar constructor and no implementation OOTB so it registers 
no classes by default. Two implementations are included, one backed by a List and the other a Map. Another implementation might be Properties based which opens up a range of possibilities, but is not included yet. This mechanism is internal to XD 1.1.x and we will revisit this in relation to other customization features for 1.2.

It is not necessary to register classes targeted for serialization, but it is offered as a performance optimization to reduce the size of the byte stream.  A simple way to register classes is to create a `/META-INF/spring-xd/bus/ext directory` in `$XD_HOME/config` and drop an XML file in there:
````
<beans ...>
   	<bean id="kryoRegistrar" class="org.springframework.xd.dirt.integration.bus.serializer.kryo.KryoClassListRegistrar">
			<constructor-arg>
				<util:list>
					<value>org.springframework.xd.extensions.test.Foo</value>
					<value>org.springframework.xd.extensions.test.Bar</value>
				</util:list>
			</constructor-arg>
		</bean>
</beans>
````
The bean id must be "kryoRegistrar". The name of the XML file does not matter. A Map backed version is also provided to allow the user to specify class IDs for each class. This is initially more effort but may give the user more  control. This also allows the user to implement alternate strategies if necessary.

This resource may also be packaged in a jar and dropped into `xd/lib`. Note that the referenced domain classes must also be on the XD class path (xd/lib) else the context will fail to initialize.

### DIFF
--- a/spring-xd-dirt/src/main/java/org/springframework/xd/dirt/integration/bus/serializer/kryo/KryoClassListRegistrar.java
+++ b/spring-xd-dirt/src/main/java/org/springframework/xd/dirt/integration/bus/serializer/kryo/KryoClassListRegistrar.java
@@ -1,0 +1,74 @@
+/*
+ * Copyright 2015 the original author or authors.
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.xd.dirt.integration.bus.serializer.kryo;
+
+import java.util.List;
+
+import com.esotericsoftware.kryo.Kryo;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+import reactor.util.Assert;
+
+/**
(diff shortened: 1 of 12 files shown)


kdowbecki:
- PojoCodec supports only one KryoRegistrar instance. In our case Spring XD is a shared, multi-tenant environment to which we can deploy multiple projects. Each of them might want to register their own types without depending on other projects.
- Have you considered adding a custom annotation or interface that would allow classes to define their own Kryo ID?

dturanski:
Re. One KryoRegistrar instance. I believe this is the way to go in order to allow validation that different projects don't provide conflicting registrations. One could even write an implementation to maintain registrations in an external data store. We need to see what the demand for the registration feature will be. Meanwhile, I prefer to leave this to the user. Also, we plan to revisit the whole user customization theme for 1.2, so this will likely be it for 1.1.x. Possibly the Map implementation may work better for you in the grand scheme of things. 

We did consider the annotation idea but this has limited application if the domain types are part of an external/shared code base which I expect would be frequently the case. I agree it would be pretty slick but seems to me like a feature for Kryo itself (if it's not there already).

markpollack:
Squashed and merged into master as ae64c32feb4a6ea2e28f3179500a5d5917819d46 and into 1.1.x as 6f9a8108dd782798f6823df72fd2266de96794a1

PR #1480, opened 2015-02-26
Properly read initial offset from partitions

mbogoevici:
Fixes an issue if the test partition already exists


garyrussell:
Rebased; merged as a9de9a0ac2006e163433978da067ee5a052bf027

PR #1484, opened 2015-02-27
[Sprint 44] XD-2764 Avoid escaping topic name twice

mbogoevici:
Due to a bug in KafkaMessageBus.escapeTopicName(), "_" is re-converted on already escaped topic names (even if it is a valid character). This causes the initialization of the topic states to read the wrong initial offsets, subsequently causing the test to fail. This behaviour is observable only on already cleaned (pruned) topics.

### DIFF
--- a/spring-xd-dirt/src/test/java/org/springframework/xd/dirt/stream/KafkaSingleNodeStreamDeploymentIntegrationTests.java
+++ b/spring-xd-dirt/src/test/java/org/springframework/xd/dirt/stream/KafkaSingleNodeStreamDeploymentIntegrationTests.java
@@ -18,13 +18,13 @@ package org.springframework.xd.dirt.stream;
 
 import static org.junit.Assert.assertNotNull;
 import static org.junit.Assert.assertTrue;
+import static org.springframework.xd.dirt.integration.kafka.KafkaMessageBus.escapeTopicName;
 
 import java.util.Collections;
 import java.util.HashMap;
 import java.util.Map;
 
 import kafka.api.OffsetRequest;
-
 import org.junit.AfterClass;
 import org.junit.BeforeClass;
 import org.junit.ClassRule;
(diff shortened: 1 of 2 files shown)


ericbottard:
I was gonna merge this but now realize the initial behavior is correct:
When there is an escape mechanism, the escaping character must always be escaped itself, or else you can't distinguish between "normal" and "escaped" cases. (e.g. in java, a backslash is `\\`).



Moreover, not sure I understand what you mean with "re-escaping". What should be passed to the escape method are un-escaped names only.

mbogoevici:
Well, the fix to the test is necessary anyway. As for the escaping, I think it's a misnomer anyway. What we are doing is re-encoding topic names by replacing invalid characters with valid sequences of characters, so in that context, I don't see the need for replacing `_`, but it's such a minor concern that from my point of view it can go one way or another.

ericbottard:
Well, we're not encoding topic names. (or at least that's not what the method did when I wrote it).
We're encoding *bus binding names* to be *kafka compliant topic names*.

mbogoevici:
Okay. Let me know if you'd prefer me to revert the changes to KafkaMessageBus. The other commit stands on its own.

ericbottard:
I think I see what you mean. And also that part about it being misnamed.

The change to KafkaMessageBus should not be applied indeed. But are you saying the rest fixes something on its own?

ericbottard:
I can test that on my own.

Also, the "fix" on KMB is currently harmless because we never *un*escape, but..

mbogoevici:
Yeah, there are two commits. One that fixes the problem that we had in the test, because escaping was applied twice and the test read initial offsets from the wrong topic. So I've changed that.

I've added the second commit later, since I couldn't see the need to modify a topic name that was already valid. But again, that, in and by itself is a different discussion (since, yes, we never unescape).

ericbottard:
Merged on master as 683e35e6e71a3ff77908455cdf811df1caff8f6b. Another PR for maintenance to come

PR #1486, opened 2015-03-03
Fixed deployment commands when no deployment properties set

dturanski:


spring-xd-shell/src/main/java/org/springframework/xd/shell/command/JobCommands.java
```diff
@@ -358,6 +360,9 @@ public String deployJob(
 				}
 				propertiesToUse = DeploymentPropertiesFormat.convert(props);
 				break;
+			case -1: // Neither option specified
+				propertiesToUse = Collections.EMPTY_MAP;
```
garyrussell: Needs to be `propertiesToUse = Collections.<String, String> emptyMap();` for type safety.

IDEA doesn't seem to be as good as eclipse at flagging such things :smile: 

Will fix during merge.

garyrussell:
Merged as c97ddba591a373c168e63a345b1c5897caa966ad

spring-xd-shell/src/main/java/org/springframework/xd/shell/command/JobCommands.java
ericbottard: I think `propertiesToUse = Collections.emptyMap();` would work too.

In any case, nice catch David, thx

PR #1487, opened 2015-03-03
[Sprint 45] [master + backport] XD-2748 Reliable spark streaming receiver

ilayaperumalg:
- Acknowledge the messages only after they are stored/replicated in spark storage
   - For Rabbit messagebus set acknowledgement mode to `MANUAL` at `MessageBusConfiguration` for spark streaming
module. This way, it is the responsibility of the `MessageBusReceiver` to acknowledge the messages
after the data blocks are persisted in spark storage.
   - For kafka messagebus,  update offset only after the messages are persisted
   - Add message headers that can be used to send acknowledgement into temporary object in the messagebus
   - Implement a `BlockGeneratorListener` and acknowledge the messages after the data is pushed
to spark storage

### DIFF
--- a/spring-xd-dirt/src/main/java/org/springframework/xd/dirt/plugins/spark/streaming/MessageBusConfiguration.java
+++ b/spring-xd-dirt/src/main/java/org/springframework/xd/dirt/plugins/spark/streaming/MessageBusConfiguration.java
@@ -43,6 +43,8 @@ import org.springframework.xd.dirt.util.ConfigLocations;
 		ConfigLocations.XD_CONFIG_ROOT + "bus/codec.xml"})
 class MessageBusConfiguration {
 
+	private static final String RABBIT_ACKMODE_PROPERTY = "xd.messagebus.rabbit.default.ackMode";
+
 	/**
 	 * This method called by {@link MessageBusReceiver} and {@link MessageBusSender} to setup
 	 * message bus configuration at the spark module executor process. This configuration
@@ -54,6 +56,10 @@ class MessageBusConfiguration {
 	 */
 	static ConfigurableApplicationContext createApplicationContext(final Properties properties) {
 		String transport = properties.getProperty("XD_TRANSPORT");
+		if (properties.getProperty(RABBIT_ACKMODE_PROPERTY) != null) {
+			properties.setProperty(RABBIT_ACKMODE_PROPERTY, "MANUAL");
+		}
+		properties.setProperty("isKafkaAutoCommitEnabled", "false");
 		SpringApplicationBuilder application = new SpringApplicationBuilder()
 				.sources(MessageBusConfiguration.class)
 				// ensure the properties are added at the first precedence level
--- a/spring-xd-dirt/src/main/java/org/springframework/xd/dirt/plugins/spark/streaming/MessageBusConfiguration.java
+++ b/spring-xd-dirt/src/main/java/org/springframework/xd/dirt/plugins/spark/streaming/MessageBusConfiguration.java
@@ -56,9 +56,8 @@ class MessageBusConfiguration {
 	 */
 	static ConfigurableApplicationContext createApplicationContext(final Properties properties) {
 		String transport = properties.getProperty("XD_TRANSPORT");
-		if (properties.getProperty(RABBIT_ACKMODE_PROPERTY) != null) {
-			properties.setProperty(RABBIT_ACKMODE_PROPERTY, "MANUAL");
-		}
+		// Set Rabbit message bus acknowledgement mode to 'MANUAL'
+		properties.setProperty(RABBIT_ACKMODE_PROPERTY, "MANUAL");
 		properties.setProperty("isKafkaAutoCommitEnabled", "false");
 		SpringApplicationBuilder application = new SpringApplicationBuilder()
 				.sources(MessageBusConfiguration.class)
--- a/spring-xd-dirt/src/main/java/org/springframework/xd/dirt/plugins/spark/streaming/MessageBusConfiguration.java
+++ b/spring-xd-dirt/src/main/java/org/springframework/xd/dirt/plugins/spark/streaming/MessageBusConfiguration.java
@@ -45,6 +45,8 @@ class MessageBusConfiguration {
 
 	private static final String RABBIT_ACKMODE_PROPERTY = "xd.messagebus.rabbit.default.ackMode";
 
+	private static final String KAFKA_AUTOCOMMIT_PROPERTY = "xd.messagebus.kafka.default.autoCommitEnabled";
+
 	/**
 	 * This method called by {@link MessageBusReceiver} and {@link MessageBusSender} to setup
 	 * message bus configuration at the spark module executor process. This configuration
@@ -58,7 +60,7 @@ class MessageBusConfiguration {
 		String transport = properties.getProperty("XD_TRANSPORT");
 		// Set Rabbit message bus acknowledgement mode to 'MANUAL'
 		properties.setProperty(RABBIT_ACKMODE_PROPERTY, "MANUAL");
-		properties.setProperty("isKafkaAutoCommitEnabled", "false");
+		properties.setProperty(KAFKA_AUTOCOMMIT_PROPERTY, "false");
 		SpringApplicationBuilder application = new SpringApplicationBuilder()
 				.sources(MessageBusConfiguration.class)
 				// ensure the properties are added at the first precedence level
--- a/spring-xd-dirt/src/main/java/org/springframework/xd/dirt/plugins/spark/streaming/MessageBusReceiver.java
+++ b/spring-xd-dirt/src/main/java/org/springframework/xd/dirt/plugins/spark/streaming/MessageBusReceiver.java
@@ -145,7 +145,7 @@ class MessageBusReceiver extends Receiver {
 					headersListToAck.add(headersList.take());
 				}
 				catch (InterruptedException ie) {
-					logger.error("Interrupted exception while getting message headers from the local blocking queue.");
+					Thread.currentThread().interrupt();
 				}
 			}
 			((MessageBusSupport) messageBus).doManualAck(headersListToAck);
@@ -175,7 +175,7 @@ class MessageBusReceiver extends Receiver {
 				headersList.put(message.getHeaders());
 			}
 			catch (InterruptedException ie) {
-				logger.error("Interrupted exception while adding message headers to local blocking queue.");
+				Thread.currentThread().interrupt();
 			}
 			blockGenerator.addDataWithCallback(message.getPayload(), message.getHeaders());
 			return true;
--- a/spring-xd-dirt/src/main/java/org/springframework/xd/dirt/plugins/spark/streaming/MessageBusReceiver.java
+++ b/spring-xd-dirt/src/main/java/org/springframework/xd/dirt/plugins/spark/streaming/MessageBusReceiver.java
@@ -18,6 +18,7 @@ package org.springframework.xd.dirt.plugins.spark.streaming;
 
 import java.util.LinkedList;
 import java.util.Properties;
+import java.util.concurrent.LinkedBlockingQueue;
 
 import org.apache.spark.SparkConf;
 import org.apache.spark.storage.StorageLevel;
@@ -71,7 +72,7 @@ class MessageBusReceiver extends Receiver {
 
 	private BlockGenerator blockGenerator = null;
 
-	private LinkedList<MessageHeaders> headersList = new LinkedList<MessageHeaders>();
+	private LinkedBlockingQueue<MessageHeaders> headersList = new LinkedBlockingQueue<MessageHeaders>();
 
 	public MessageBusReceiver(StorageLevel storageLevel, Properties messageBusProperties,
 			Properties moduleConsumerProperties, MimeType contentType) {
@@ -140,7 +141,12 @@ class MessageBusReceiver extends Receiver {
 			store(dataBuffer);
 			LinkedList<MessageHeaders> headersListToAck = new LinkedList<MessageHeaders>();
 			for (int i = 0; i < dataBuffer.size(); i++) {
-				headersListToAck.add(headersList.pop());
+				try {
+					headersListToAck.add(headersList.take());
+				}
+				catch (InterruptedException ie) {
+					logger.error("Interrupted exception while getting message headers from the local blocking queue.");
+				}
 			}
 			((MessageBusSupport) messageBus).doManualAck(headersListToAck);
 		}
@@ -164,9 +170,14 @@ class MessageBusReceiver extends Receiver {
 		}
 
 		@Override
-		protected boolean doSend(Message<?> message, long timeout) {
+		protected synchronized boolean doSend(Message<?> message, long timeout) {
+			try {
+				headersList.put(message.getHeaders());
+			}
+			catch (InterruptedException ie) {
+				logger.error("Interrupted exception while adding message headers to local blocking queue.");
+			}
 			blockGenerator.addDataWithCallback(message.getPayload(), message.getHeaders());
-			headersList.add(message.getHeaders());
 			return true;
 		}
 	}
--- a/spring-xd-dirt/src/main/java/org/springframework/xd/dirt/plugins/spark/streaming/MessageBusReceiver.java
+++ b/spring-xd-dirt/src/main/java/org/springframework/xd/dirt/plugins/spark/streaming/MessageBusReceiver.java
@@ -18,7 +18,11 @@ package org.springframework.xd.dirt.plugins.spark.streaming;
 
 import java.util.Properties;
 
+import org.apache.spark.SparkConf;
 import org.apache.spark.storage.StorageLevel;
+import org.apache.spark.storage.StreamBlockId;
+import org.apache.spark.streaming.receiver.BlockGenerator;
+import org.apache.spark.streaming.receiver.BlockGeneratorListener;
 import org.apache.spark.streaming.receiver.Receiver;
 import org.slf4j.Logger;
 import org.slf4j.LoggerFactory;
@@ -28,6 +32,9 @@ import org.springframework.integration.channel.DirectChannel;
 import org.springframework.messaging.Message;
 import org.springframework.util.MimeType;
 import org.springframework.xd.dirt.integration.bus.MessageBus;
+import org.springframework.xd.dirt.integration.bus.MessageBusSupport;
+
+import scala.collection.mutable.ArrayBuffer;
 
 /**
  * Spark {@link Receiver} implementation that binds to the MessageBus as a consumer.
@@ -60,6 +67,9 @@ class MessageBusReceiver extends Receiver {
 
 	private final MimeType contentType;
 
+	private BlockGenerator blockGenerator = null;
+
+
 	public MessageBusReceiver(StorageLevel storageLevel, Properties messageBusProperties,
 			Properties moduleConsumerProperties, MimeType contentType) {
 		this(null, storageLevel, messageBusProperties, moduleConsumerProperties, contentType);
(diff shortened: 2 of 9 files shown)


spring-xd-messagebus-kafka/src/main/java/org/springframework/xd/dirt/integration/kafka/KafkaMessageBus.java
```diff
@@ -789,4 +806,43 @@ private int roundRobin() {
 
 	}
 
+	@Override
+	public void storeForManualAck(MessageHeaders headers) {
+		String topic = (String) headers.get(KafkaHeaders.TOPIC);
+		Partition partition = new Partition(topic, (int) headers.get(KafkaHeaders.PARTITION_ID));
+		Long offset = (Long) headers.get(KafkaHeaders.OFFSET);
```
mbogoevici: Unfortunately, the proper header here should be "nextOffset", which is missing from the current implementation of SI-Kafka. The known issue here is that if we store the current offset, then in case of a crash, reading will replay the last item. This is addressed as part of https://github.com/spring-projects/spring-integration-kafka/pull/20

ilayaperumalg: yeah, but it is nice that you already have it addressed in the SI-kafka PR. Once that PR is merged, I can either update this PR or submit a new PR once the SI changes are merged.

spring-xd-messagebus-rabbit/src/main/java/org/springframework/xd/dirt/integration/rabbit/RabbitMessageBus.java
```diff
@@ -874,4 +878,30 @@ public boolean getAutoBindDLQ(boolean defaultValue) {
...
 
+	@Override
+	public void storeForManualAck(MessageHeaders headers) {
+		if (headers.containsKey("amqp_channel")) {
+			Channel channel = (com.rabbitmq.client.Channel) headers.get("amqp_channel");
+			if (!channelsForManualAck.containsKey(channel)) {
+				channelsForManualAck.put(channel, 1);
+			}
+			else {
+				channelsForManualAck.put(channel, channelsForManualAck.get(channel) + 1);
```
garyrussell: You can't use a simple incrementing value; you need to retrieve the delivery tag from the headers (`amqp_deliveryTag`). Also, instead of hard-coded strings, you should use the `AmqpHeaders` constants.

spring-xd-dirt/src/main/java/org/springframework/xd/dirt/plugins/spark/streaming/MessageBusConfiguration.java
```diff
@@ -54,6 +56,10 @@
 	 */
 	static ConfigurableApplicationContext createApplicationContext(final Properties properties) {
 		String transport = properties.getProperty("XD_TRANSPORT");
+		if (properties.getProperty(RABBIT_ACKMODE_PROPERTY) != null) {
+			properties.setProperty(RABBIT_ACKMODE_PROPERTY, "MANUAL");
+		}
```
garyrussell: Should we even allow an override if `MANUAL` is required to make it work?

If not, we could either abort the deploy or just log a WARN.

spring-xd-messagebus-spi/src/main/java/org/springframework/xd/dirt/integration/bus/MessageBusSupport.java
```diff
@@ -1103,4 +1104,19 @@ public void handleMessage(Message<?> message) throws MessagingException {
...
+	 */
+	public void storeForManualAck(MessageHeaders messageHeaders) {
+	}
+
+	/**
+	 * Perform manual acknowledgement based on the metadata stored in message bus.
+	 */
+	public void doManualAck() {
+	}
+
```
garyrussell: Should these eventually be pulled up to `MessageBus` ? (Obviously not for 1.1.x)

Also, please move them above the `DefaultPartitionSelectorClass`; convention is to have classes at the end of the file.

spring-xd-messagebus-rabbit/src/main/java/org/springframework/xd/dirt/integration/rabbit/RabbitMessageBus.java
```diff
@@ -269,6 +271,8 @@ public MessageProperties toMessageProperties(AMQP.BasicProperties source, Envelo
 	private static final ExpressionParser expressionParser =
 			new SpelExpressionParser(new SpelParserConfiguration(true, true));
 
+	private Map<Channel, Integer> channelsForManualAck = new HashMap<Channel, Integer>();
```
garyrussell: Needs to be a `LinkedHashMap` so the iterator runs in insertion order.

spring-xd-messagebus-rabbit/src/main/java/org/springframework/xd/dirt/integration/rabbit/RabbitMessageBus.java
```diff
@@ -874,4 +878,30 @@ public boolean getAutoBindDLQ(boolean defaultValue) {
...
+			}
+			else {
+				channelsForManualAck.put(channel, channelsForManualAck.get(channel) + 1);
+			}
+		}
+	}
+
+	@Override
+	public void doManualAck() {
+		for (final com.rabbitmq.client.Channel channel : channelsForManualAck.keySet()) {
```
garyrussell: It's more efficient to iterate over the `entrySet()` to save the hit of `get()`.

spring-xd-dirt/src/main/java/org/springframework/xd/dirt/plugins/spark/streaming/MessageBusConfiguration.java
ilayaperumalg: We copy the message bus properties from XD configuration (along with what end user provides) and set here for Spark. At this level we don't know if the property comes from XD config or user level settings. That's the reason, we always override irrespective what's being set already. What do you think?

spring-xd-messagebus-rabbit/src/main/java/org/springframework/xd/dirt/integration/rabbit/RabbitMessageBus.java
```diff
@@ -874,4 +878,30 @@ public boolean getAutoBindDLQ(boolean defaultValue) {
...
+			try {
+				channel.basicAck(channelsForManualAck.get(channel), true);
```
garyrussell: When `multiple` is true; you should only send the highest `deliveryTag` and all previous will be ackd.

spring-xd-messagebus-spi/src/main/java/org/springframework/xd/dirt/integration/bus/MessageBusSupport.java
ilayaperumalg: Yeah, this is going to in MessageBus. We want to have this change in 1.1.x. Could you tell me why we can't have it in 1.1.x?

spring-xd-messagebus-rabbit/src/main/java/org/springframework/xd/dirt/integration/rabbit/RabbitMessageBus.java
garyrussell: Hmmm - I just looked at Spring AMQP and the channel is not a good key for a Map (the hashCode() might change).

garyrussell: Is it possible to have multiple `MessageBusReceiver` per bus (> 1 stream) ?

If so, don't we need to keep track by stream?

ilayaperumalg: No, when the Spark streaming receiver runs on spark cluster's executor node, each `MessageBusReceiver` is configured to have its own `MessageBus`.

spring-xd-messagebus-rabbit/src/main/java/org/springframework/xd/dirt/integration/rabbit/RabbitMessageBus.java
garyrussell: Also, the underlying channel might be reset so it's not necessarily the case that deliveryTags will be incrementing.

spring-xd-dirt/src/main/java/org/springframework/xd/dirt/plugins/spark/streaming/MessageBusConfiguration.java
garyrussell: Sorry, I missed the `!` but don't we __always__ want to set it to `MANUAL`, regardless of whether the bus/user set it?

spring-xd-messagebus-spi/src/main/java/org/springframework/xd/dirt/integration/bus/MessageBusSupport.java
garyrussell: We (Spring) generally don't allow breaking changes to interfaces in point releases - it will break someone's code if, for example, they have implemented a custom MB.

Perhaps unlikely but...

spring-xd-dirt/src/main/java/org/springframework/xd/dirt/plugins/spark/streaming/MessageBusConfiguration.java
ilayaperumalg: Yeah, we always want to set it. I thought of checking if the property is being used. Since we always set this property (via application.yml), I will remove this check and set it always.

spring-xd-messagebus-spi/src/main/java/org/springframework/xd/dirt/integration/bus/MessageBusSupport.java
ilayaperumalg: I see. Lets discuss if we are ok to have this change to include reliable receiver in 1.1.x.

spring-xd-messagebus-rabbit/src/main/java/org/springframework/xd/dirt/integration/rabbit/RabbitMessageBus.java
garyrussell: I removed the comment about the channel not being a good candidate for a map key; my mistake I was looking at the wrong class; it's ok.

spring-xd-messagebus-rabbit/src/main/java/org/springframework/xd/dirt/integration/rabbit/RabbitMessageBus.java
```diff
@@ -673,6 +677,34 @@ public void destroy() {
...
+	public void storeForManualAck(MessageHeaders headers) {
+		if (headers.containsKey(AmqpHeaders.CHANNEL)) {
+			Channel channel = (com.rabbitmq.client.Channel) headers.get(AmqpHeaders.CHANNEL);
+			Long deliveryTag = (Long) headers.get(AmqpHeaders.DELIVERY_TAG);
+			if (channelsForManualAck.containsKey(channel) && (channelsForManualAck.get(channel) < deliveryTag)) {
+				channelsForManualAck.put(channel, deliveryTag);
+			}
+			else {
+				channelsForManualAck.put(channel, deliveryTag);
+			}
```
garyrussell: The code is identical regardless of the if condition.

This is actually correct because you can't assume deliveryTags are incrementing.

Let's say your "batch size" is 10 and you've already pushed 10.

Now we deliver 11-15 and there's a network glitch (connection reset).

11-15 will be redelivered but they will have delivery tags 1-5.

ilayaperumalg: oh, ok. thanks for clarifying. I will remove the if condition check.

spring-xd-messagebus-rabbit/src/main/java/org/springframework/xd/dirt/integration/rabbit/RabbitMessageBus.java
```diff
@@ -673,6 +677,34 @@ public void destroy() {
...
+		}
+	}
+
+	@Override
+	public void doManualAck() {
+		for (Map.Entry<Object, Long> entry : channelsForManualAck.entrySet()) {
```
garyrussell: AFAICT this will only work if concurrency = 1 on the bus (and then there is only one channel).

Consider concurrency = 2.

Correct me if I'm wrong; let's say the batch size is 10 each thread delivers 5 messages and this triggers a push. The spark thread starts to call this method but, in the meantime, message 6 on thread 1 arrives, updating its deliveryTag to 6. Won't the spark push then ack 6 for thread 1 and 5 for thread 2 (which is wrong)?

It seems to me you need to keep a bucket of acks somehow and then figure out at ack time which ones were actually part of the block. Do you have access to the streamBlockId that receives a message?

Am I missing something?

ilayaperumalg: That's a great catch. We do have access to `streamBlockId` for each block being pushed. I can add blockId as a key to each batch of messages that get acked.

ilayaperumalg: Looking at it again, we only can access `blockId`s after the streaming blocks are generated and ready for push. So, it doesn't seem possible to bucket the messages based on some reference unique to the block. Investigating...

ilayaperumalg:
@garyrussell Pushed the changes to take care of acknowledging **only** the messages that were part of the Spark streaming `block` that was pushed to Spark storage.

There is one comment I would like to discuss which I left as **todo** here: https://github.com/spring-projects/spring-xd/pull/1487/files#diff-8148ef32d0904696965ca7af237580a5R694

spring-xd-dirt/src/main/java/org/springframework/xd/dirt/plugins/spark/streaming/MessageBusReceiver.java
```diff
@@ -60,6 +69,10 @@
 
 	private final MimeType contentType;
 
+	private BlockGenerator blockGenerator = null;
+
+	private LinkedList<MessageHeaders> headersList = new LinkedList<MessageHeaders>();
```
garyrussell: `LinkedList` is not thread-safe - `LinkedBlockingQueue` ?

spring-xd-dirt/src/main/java/org/springframework/xd/dirt/plugins/spark/streaming/MessageBusReceiver.java
```diff
@@ -121,7 +165,8 @@ public MessageStoringChannel() {
 
 		@Override
 		protected boolean doSend(Message<?> message, long timeout) {
-			store(message.getPayload());
+			blockGenerator.addDataWithCallback(message.getPayload(), message.getHeaders());
+			headersList.add(message.getHeaders());
```
garyrussell: This looks like a race condition - we could get the callback before we add the headers.

Even if you swap the statements, consider concurrency; it would be possible that one thread puts its header in first, while the second initiates the push.

You might have to synchronize this entire method.

spring-xd-messagebus-rabbit/src/main/java/org/springframework/xd/dirt/integration/rabbit/RabbitMessageBus.java
```diff
@@ -673,6 +682,31 @@ public void destroy() {
...
+		Iterator<MessageHeaders> iterator = messageHeadersList.iterator();
+		Map<Object, Long> channelsToAck = new HashMap<Object, Long>();
+		while(iterator.hasNext()) {
+			MessageHeaders messageHeaders = iterator.next();
+			if (messageHeaders.containsKey(AmqpHeaders.CHANNEL)) {
+				Channel channel = (com.rabbitmq.client.Channel) messageHeaders.get(AmqpHeaders.CHANNEL);
+				Long deliveryTag = (Long) messageHeaders.get(AmqpHeaders.DELIVERY_TAG);
+				// todo: If the order of messages isn't preserved in spite of storing the incoming
+				// todo: messages into LinkedList, how do we handle the deliveryTag?
+				channelsToAck.put(channel, deliveryTag);
```
garyrussell: The delivery tags (by channel) should be in the right (delivery) order so I believe it's safe to (multiple) ack the last tag for each channel.

Understand, though, that we might get duplicates messages. Connection resets half-way through a block, the already unacked messages will be redelivered and added to the same block.

But the delivery tag handling will be correct.

ilayaperumalg:
Updated the PR based on the review.

spring-xd-dirt/src/main/java/org/springframework/xd/dirt/plugins/spark/streaming/MessageBusReceiver.java
```diff
@@ -107,6 +124,39 @@ public void onStop() {
...
+		}
+
+		public void onPushBlock(StreamBlockId streamBlockId, ArrayBuffer<?> dataBuffer) {
+			store(dataBuffer);
+			LinkedList<MessageHeaders> headersListToAck = new LinkedList<MessageHeaders>();
+			for (int i = 0; i < dataBuffer.size(); i++) {
+				try {
+					headersListToAck.add(headersList.take());
+				}
+				catch (InterruptedException ie) {
```
garyrussell: Whenever you catch `InterruptedException` you must invoke `Thread.currentThread().interrupt()`.

spring-xd-dirt/src/main/java/org/springframework/xd/dirt/plugins/spark/streaming/MessageBusReceiver.java
```diff
@@ -120,8 +170,14 @@ public MessageStoringChannel() {
 		}
 
 		@Override
-		protected boolean doSend(Message<?> message, long timeout) {
-			store(message.getPayload());
+		protected synchronized boolean doSend(Message<?> message, long timeout) {
+			try {
+				headersList.put(message.getHeaders());
+			}
+			catch (InterruptedException ie) {
```
garyrussell: ditto

spring-xd-messagebus-rabbit/src/main/java/org/springframework/xd/dirt/integration/rabbit/RabbitMessageBus.java
```diff
@@ -673,6 +682,29 @@ public void destroy() {
...
+			if (messageHeaders.containsKey(AmqpHeaders.CHANNEL)) {
+				Channel channel = (com.rabbitmq.client.Channel) messageHeaders.get(AmqpHeaders.CHANNEL);
+				Long deliveryTag = (Long) messageHeaders.get(AmqpHeaders.DELIVERY_TAG);
+				channelsToAck.put(channel, deliveryTag);
+			}
+		}
+		for (Map.Entry<Object, Long> entry : channelsToAck.entrySet()) {
+			try {
+				Channel channel = (Channel) entry.getKey();
+				channel.basicAck(entry.getValue(), true);
```
garyrussell: You don't really need the `channel` variable here; you can concat these 2 lines.

garyrussell:
Otherwise, LGTM from the rabbit side; over to @mbogoevici for Kafka (if not done already)

mbogoevici:
My comments are captured already - we need the manual ack support that's forthcoming with INTEXT-143.

mbogoevici:
Upgraded SI Kafka version to 1.1.0.RELEASE
Squashed.
Merged master here w/https://github.com/spring-projects/spring-xd/commit/0d672aa50ec16888a66fd886186fa816b41e43c0
Merged 1.1.x branch https://github.com/spring-projects/spring-xd/commit/cab775911bd4f735218c44f788cd03d9a490ae9e

PR #1488, opened 2015-03-04
[Sprint 44] [master + backport] XD-2671 Handle stream/job undeploy when deployment in progress

ilayaperumalg:
- Check for `NoNodException` on the stream/job deployment status path
and if the exception is thrown, then the stream/job is considered to
have been un-deployed. Then, proceed with deletion of stream deployment
path and module deployment request paths.


ilayaperumalg:
Based on the JIRA XD-2671 which states the modules left orphaned when the stream/job is deployed/un-deployed in quick succession. The PR changes are applied to `InitialDeploymentListener` to handle the deployment requests initiated by the REST calls.

There is one other edge case when the deployment is handled by the `ModuleRedeployer` and the undeployment happens at the same time. This is not addressed along with this PR as I think that requires some discussion/investigation.

ilayaperumalg:
Closing this PR as I am working on a different approach after some suggestions from @pperalta and @markfisher

PR #1490, opened 2015-03-07
XD-2794: Add MongoDB Source

agandhinit:
The diff.


ericbottard:
Thanks for this contribution, looks very good overall.

You'll need to change the gradle build to have your module project appear there (similar to the mongo sink)
Also, do you think you could be able to factor out the module options that relate to connecting/authenticating to a mongo instance out of `FromMongoDbOptionMixin` and `IntoMongoDbOptionMixin`, leaving out only the read (respectively write) details in those classes?

ericbottard:
Applied the refactorings discussed above and merged as eb733edf34747e4c7bf35895d31f1211ba2f71f2

PR #1492, opened 2015-03-09
[Sprint: 45] XD-2774 Fix error handling in jdbchdfs job

trisberg:


mminella:
Why doesn't this writer take the same approach as the {{FlatFileItemWriter}} with a transaction aware buffer of some kind ({{TransactionAwareBufferedWriter}} in it's case)?

trisberg:
Don't know, that's a different issue though, maybe create a new story for that. We are also considering re-writing this job to have a more direct path from jdbc->hdfs and not use the chunk approach to speed things up.

mminella:
I ask because of the lifecycle of a transaction.  However, if we're going to gut this for a more high speed approach, then nevermind.  Merged.

trisberg:
We need this backported to 1.1.x branch as well

mminella:
Coming right up

mminella:
Ported to 1.1.x

PR #1494, opened 2015-03-10
[Sprint: 45] [Backport] XD-2704: Have documentation reside in the main repo

ericbottard:
This PR does the following:
- remove everything related to pulling the docs from the wiki repo
- use a single asciidoctor task, configured with two backends, expecting to find the asciidoctor files in `src/docs/asciidoc`
- provide a shell script that automates the following, which the poor soul merging this PR should run:
 1. clone the latest version of the docs
 2. rename `FullGuide` to `index`
 3. rewrite all `link`s to `xref:`
 4. put images directly inside `images`, where they are correctly picked up in all settings, including live-preview in the browser with extensions, for example
- re-installs the link checking task, which somehow got lost previously :(

I've also explored the following:
  1. upgrading the asciidoctor plugin. Sadly, this requires newer groovy (ie newer gradle, even the compat package does not work). In turn, this requires changes to https://github.com/spring-projects/spring-xd/blob/master/spring-xd-starters/spring-xd-module-parent/publish-maven.gradle#L70 
  2. so I looked into https://gradle.org/docs/current/userguide/publishing_maven.html, seems like a big change
  3. Also looked into getting rid of the docbook toolchain altogether. There are still some issues, but this could be worth it

To merge this PR:
 1. run `./convertDocs.sh`
 2. verify
 3. check in src/docs under source control
 4. convertDocs.sh can be deleted
 5. I suggest pasting in this PR comments the sha1 of the wiki repo at the time of merging, as the wiki should then be considered decommissioned

I have already fixed quite a few issues on the wiki directly (master), so this should pass properly

### DIFF
--- a/gradle/build-docs.gradle
+++ b/gradle/build-docs.gradle
@@ -9,211 +9,78 @@ buildscript {
 	dependencies {
 		classpath 'io.spring.gradle:docbook-reference-plugin:0.3.0'
 		classpath 'org.asciidoctor:asciidoctor-gradle-plugin:1.5.0'
-		classpath 'org.ajoberstar:gradle-git:0.8.0'
+		//classpath 'org.asciidoctor:asciidoctorj-pdf:1.5.0-alpha.6' // asciidoctor pdf still has issues with callouts, etc. Still relying on docbook+pdf for now
 	}
 
 }
 
-import org.ajoberstar.grgit.*
-import org.ajoberstar.grgit.auth.AuthConfig
-import org.apache.tools.ant.filters.TokenFilter
-import org.asciidoctor.gradle.*
+def docsDir = 'src/docs/asciidoc' // Will be default with newer asciidoctor plugin
 
-task(pullDocs) {
-	if (project.gradle.startParameter.offline) {
-		enabled = false
-	}
-	ext.cloneDestination = file("$buildDir/asciidoc-raw")
-}
-
-pullDocs << {
-
-	def xdGit = Grgit.open(rootProject.projectDir)
-	def springXdBranchName = xdGit.branch.getCurrent().name
-
-	println "Spring XD Branch: $springXdBranchName"
-	println "Spring XD Version: $version"
-
-	if (rootProject.hasProperty('wikiBranch')) {
-		ext.wikiBranch = rootProject.getProperty('wikiBranch')
-		println "Using custom wiki branch: ${ext.wikiBranch}"
-	}
-	else {
-		ext.wikiBranch = 'origin/' + springXdBranchName;
-		println "Using default wiki branch: ${ext.wikiBranch}"
-	}
-
-	def uri = "https://github.com/spring-projects/spring-xd.wiki.git"
-
-	if (cloneDestination.exists()) {
-		cloneDestination.deleteDir()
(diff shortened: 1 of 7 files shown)


gradle/build-docs.gradle
```diff
@@ -9,211 +9,78 @@ buildscript {
...
 }
 
+asciidoctor.dependsOn ([checkDocsLinks, shellReferenceDoc, moduleOptionsReferenceDoc])
+
 apply plugin: DocbookReferencePlugin
 
 reference {
 	sourceFileName = 'index.xml'
-	sourceDir = file("$buildDir/docbook")
+	sourceDir = file("$buildDir/html")
```
ghillert: May need to change this - currently the Docbook xml file is placed under `/build/html/index.xml`, thus it would be published to the docs website, maybe something we don't want to.

gradle/build-docs.gradle
```diff
@@ -9,211 +9,78 @@ buildscript {
 	dependencies {
 		classpath 'io.spring.gradle:docbook-reference-plugin:0.3.0'
 		classpath 'org.asciidoctor:asciidoctor-gradle-plugin:1.5.0'
-		classpath 'org.ajoberstar:gradle-git:0.8.0'
+		//classpath 'org.asciidoctor:asciidoctorj-pdf:1.5.0-alpha.6' // asciidoctor pdf still has issues with callouts, etc. Still relying on docbook+pdf for now
```
ghillert: should this be removed?

ghillert:
@ericbottard Overall - looks good and works well. See 2 minor comments above.

gradle/build-docs.gradle
ericbottard: Wanted to leave that here reminding the next reader about
- the fact that asciidoctor now has direct support for pdf...
- ... but it's not ready yet

gradle/build-docs.gradle
ericbottard: Yeah, saw that and thought about it.
I guess it doesn't hurt *that much*, especially given that accessing `thedocs/` will display `thedocs/index.html`, so only people aware of it would see it

ericbottard:
@ghillert Added 2dfaa8d to automate wiki deprecation.

ghillert:
Looks good. Will need to adapt the migration scripts during merge for the 1.1 backport.

ghillert:
Migrated docs for MASTER - Relevant spring-xd.wiki commits:

* Pre-migration hash: 83767e3dd339fb7eeaf6f7fa3f5853b766ac72b1
* Post migration hash: 115a5dcf724c9118571fc815c0082f86bdd5f6c8

ericbottard:
`\o/ \o/ \o/`

ghillert:
Migrated docs for `1.1.x`

    * Pre-migration hash: 5757108a23d4acd1cd30a6b79f88abd9a897683b
    * Post migration hash: 737f9cbee30d10c1912111796283b6e6b17b4e83

Commits:

* https://github.com/spring-projects/spring-xd/commit/6f6a6dd0d00ed69fef85fc2f1167f0a33169eb06
* https://github.com/spring-projects/spring-xd/commit/fa1b77ff9c40c762d118262d679f8acb7cf3c871

PR #1496, opened 2015-03-12
[Sprint: 45] XD-2807: fix build

ericbottard:
The major benefit is that the ui project is not cleaned everytime.
Also removed -exec related leftovers
Optimized hadoop distro related copyings


build.gradle
```diff
@@ -1042,17 +997,17 @@ project('spring-xd-ui') {
...
 	installGrunt.shouldRunAfter 'npmInstall'
 
 	task cleanUI(dependsOn: [
-			'clean',
 			'cleanDist',
 			'cleanNodeModules',
 			'cleanBowerFiles'
 	]) { description = "Clean the Admin UI dist, node modules and bower files" }
+	
+	//tasks['clean'].dependsOn(['cleanUI']) // 'dist' files are checked into VCS for now, don't clean
```
ghillert: @ericbottard This is commented out? How does that relate to `The major benefit is that the ui project is not cleaned everytime.`

build.gradle
```diff
@@ -1042,17 +997,17 @@ project('spring-xd-ui') {
 	task setupUI(dependsOn: ['grunt_build', 'build']) { description = "Build the Admin UI using Grunt" }
 
 	tasks['build'].shouldRunAfter 'grunt_build'
-	tasks['compileJava'].dependsOn(['clean'])
```
ericbottard: This line previously triggered a clean on every build

build.gradle
ericbottard: See my comment above.

Line 1010 is commented out because artifacts are checked into source control. If they were not, we'd de-comment it

ghillert:
Rebased and Merged, fixed a trailing white-space during merge.

PR #1499, opened 2015-03-16
[Sprint: 45] [Backport] XD-2767: JMS Source Recovery

garyrussell:
JIRA: https://jira.spring.io/browse/XD-2767

Previously, the JMS source could lose messages.

Expose the `acknowledge` property to enable transactions.

Also, when not using transactions, use a `SimpleMessageListenerContainer`
instead of a `DefaultMessageListenerContainer` so that acknowledgement
occurs after the listener completes instead of before.

Tested with a local bus and:

    xd:>stream create jmstest --definition "jms | transform --expression='headers[jms_redelivered] ? payload : 1/0' | log --expression=#root" --deploy

    10:20:23,498  INFO ActiveMQ Session Task-1 sink.jmstest - GenericMessage [payload=foo, headers={jms_redelivered=true, JMSXDeliveryCount=2, id=db7d4d51-8e25-4b0e-0217-ab6c26c195a3, priority=0, jms_timestamp=1426515616699, jms_messageId=ID:arwen4-39565-1426346536573-15:1:1:1:1, timestamp=1426515623497}]

and

    xd:>stream create jmstest --definition "jms --acknowledge=transacted | transform --expression='headers[jms_redelivered] ? payload : 1/0' | log --expression=#root" --deploy

    10:21:47,231  INFO org.springframework.jms.listener.DefaultMessageListenerContainer#0-1 sink.jmstest - GenericMessage [payload=foo, headers={jms_redelivered=true, JMSXDeliveryCount=2, id=8a63d421-7059-53bc-a352-3426427b2d23, priority=0, jms_timestamp=1426515700437, jms_messageId=ID:arwen4-39565-1426346536573-17:1:1:1:1, timestamp=1426515707231}]

showing that in both cases, the failed message (due to the divide by zero) is successfully recovered.

### DIFF
--- a/spring-xd-dirt/src/main/java/org/springframework/xd/dirt/modules/metadata/JmsSourceModuleOptionsMetadata.java
+++ b/spring-xd-dirt/src/main/java/org/springframework/xd/dirt/modules/metadata/JmsSourceModuleOptionsMetadata.java
@@ -19,6 +19,7 @@ package org.springframework.xd.dirt.modules.metadata;
 import static org.springframework.xd.module.options.spi.ModulePlaceholders.XD_STREAM_NAME;
 
 import javax.validation.constraints.NotNull;
+import javax.validation.constraints.Pattern;
 
 import org.springframework.xd.module.options.spi.ModuleOption;
 import org.springframework.xd.module.options.spi.ProfileNamesProvider;
@@ -76,6 +77,7 @@ public class JmsSourceModuleOptionsMetadata implements ProfileNamesProvider {
 		return clientId;
 	}
 
+	@Pattern(regexp = "(auto|transacted|dups-ok|client)")
 	public String getAcknowledge() {
 		return acknowledge;
 	}
--- a/spring-xd-dirt/src/main/java/org/springframework/xd/dirt/modules/metadata/JmsSourceModuleOptionsMetadata.java
+++ b/spring-xd-dirt/src/main/java/org/springframework/xd/dirt/modules/metadata/JmsSourceModuleOptionsMetadata.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2014 the original author or authors.
+ * Copyright 2014-2015 the original author or authors.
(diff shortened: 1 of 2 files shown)


spring-xd-dirt/src/main/java/org/springframework/xd/dirt/modules/metadata/JmsSourceModuleOptionsMetadata.java
```diff
@@ -42,6 +44,8 @@
 
 	private String clientId = null;
 
+	private String acknowledge = "auto";
```
ericbottard: Wanted to see if we could use enums, but could not because of `dups-ok`.

Then, got me thinking about each combination. Does `dmlc ` only make sense with `transacted`?

```
auto
client
dups-ok
transacted
```

At the very least, I'd like to add jsr303 validation to that option. But not sure about the profiles to activate for each?

garyrussell: Yes, the DMLC is only used for `transacted`. The basic difference is that with the SMLC, the message listener is a true JMS `MessageListener` and the ack (auto or dups-ok) is sent when the listener exits.

With the DMLC, `consumer.receive()` is called (with the ack emitted when receive exits) and __then__ the listener is invoked.

If you run the test stream without this commit, you will see that the failed message is never redelivered because it was already ack'd.

garyrussell:
Rebased; added validation.

ericbottard:
Ah! When I said 
> I'd like to add jsr303 validation

I really meant _**I**_. Thanks for adding this!

Merged as 139ba2837be7c22906df0d4e8bf0b47948335431 with case insensitive match + explicit message

ericbottard:
603619b on the 1.1.x branch

garyrussell:
Thanks for polishing

PR #1506, opened 2015-03-19
[Sprint 45] XD-2832 job launch for @Configuration job modules 1.1.x

dturanski:


dturanski:
@ilayaperumalg - actually I suspect there is still a problem with the Step. e.g. 

````
bean instanceof StepParserStepFactoryBean<?, ?>
````
How can we test for that? tap stepExecution channel?

**Comment by dturanski** [comment]
The tests pass but something similar needs to happen at the Step level.  Currently the XD code deals with StepListener, the parent of StepExecutionListener @mminella - I was looking into the batch API. XML config provides a StepParserFactoryBean which XD uses at runtime to set an array of StepListener[] including StepExecutionListeners, ChunkListeners, etc.  I'm having trouble finding the equivalent API for @Configuration, starting from StepBuilderFactory.  (This is a general weakness in spring that the bean provider logic is typically coupled to the parsing logic). I'm sure there's a way to do it but would appreciate your thoughts.

**Comment by mminella** [comment]
@dturanski I'd say let's merge this as a start and log the other issue with the lack of support for the listeners for java configured jobs.  A quick glance through the batch code didn't give me anything obvious without reflection hacks.

**Comment by ilayaperumalg** [comment]
@dturanski the step listeners can be tested by creating a tap on the out of the box step listeners:

```
xd>job create --name mycustomJob --definition "mycustomJob" --deploy
xd>stream create --name stepExecutionEvents --definition "mycustomJob.step >log" --deploy
xd>job launch mycustomJob
```

ilayaperumalg:
Created JIRA https://jira.spring.io/browse/XD-2841 to investigate the listeners in case of job modules with java config. Merging the changes...

ilayaperumalg:
Merged into 1.1.x f8507b95cfaa8e0edcf761ecbacc67ad772e370b

PR #1511, opened 2015-03-20
[Sprint 45] XD-2837 Admin scripts to include hadoop classpath

ilayaperumalg:
- Setup and add hadoop classpath into XD-admin JVM classpath

### DIFF
--- a/scripts/xd/xd-admin
+++ b/scripts/xd/xd-admin
@@ -67,17 +67,42 @@ cd "$SAVED" >&-
 
 APP_HOME_LIB=$APP_HOME/lib
 CLASSPATH=$APP_HOME/modules/processor/scripts:$APP_HOME/config:$APP_HOME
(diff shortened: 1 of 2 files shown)


ericbottard:
This works (tested the *nix version) and allows one to successfully upload a module to HDFS with `xd-admin`.

However,
* The admin does not need full "distro specific" support, merely the ability to write to Hdfs using `HdfsResource`. I don't know if this is possible without needing to choose a specific distro
* The way this is done (similar to container) is really brittle at the moment. But there is a story dealing with hadoop classpath in the next sprint, so merging this as is
* For the same reasons, not worrying too much about the inclusion of `hadoopDistro` in the usage message

ericbottard:
Merged as 47fa882c2fdcf9c1880c044b1035bc78a4f62166

PR #1512, opened 2015-03-22
[XD-2769] Making receive() and receivePayload() consistent with Spring Integration

kdowbecki:
Should tests for changed classes be provided with this pull request? I searched the project but I haven't found existing tests for NamedChannelSink, please point me in the right direction.

### DIFF
--- a/spring-xd-dirt/src/main/java/org/springframework/xd/dirt/test/process/SingleNodeProcessingChain.java
+++ b/spring-xd-dirt/src/main/java/org/springframework/xd/dirt/test/process/SingleNodeProcessingChain.java
@@ -57,7 +57,6 @@ public class SingleNodeProcessingChain extends AbstractSingleNodeProcessingChain
 		return sink.receivePayload(timeout);
 	}
 
-
 	@Override
 	public void send(Message<?> message) {
 		source.send(message);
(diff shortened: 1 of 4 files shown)


garyrussell:
We can't use blocking `receive()` in tests (see my comment on the JIRA); use a long enough timeout instead.

dturanski:
I would even be ok removing receive() and receivePayload() from NamedChannelSink. The tests for these classes are in ExampleProcessingChainTests.

dturanski:
@garyrussell, @kdowbecki  I think it's best if we remove receive() and receivePayload() altogether since  blocking methods should never be used for tests and backing with a long timeout is 1) inconsistent and 2) difficult to choose an appropriate value.

kdowbecki:
I like those blocking methods because I prefer to enforce hard timeouts globally. From where I stand using a blocking ```receive()``` or ```receive(int)``` with huge timeout is equally bad.

Ideally there should be a way to distinguish between timeout and no message produced by the component (e.g. Spring Integration <service-activator requires-reply=false/> or <filter/> inside the module). Currently it's always timeout and null is returned from receive(int ).

If you think that removing them will improve the test framework usability and guard developers from doing a build-halting mistake please go ahead. As it is written now ```receive()``` can't be used so no one should complain.

dturanski:
closed https://github.com/spring-projects/spring-xd/commit/a5a6a556420a7a07024a08bcc2df526d608c8171

PR #1514, opened 2015-03-23
[Sprint: 46] XD-2853 Deployment message consumer ordering

ilayaperumalg:
- Make sure the ZK distributed queue consumer for
processing deployment messages is initialized after
the path cache is setup for module, stream and job
deployment requests paths.

### DIFF
--- a/spring-xd-dirt/src/main/java/org/springframework/xd/dirt/server/admin/deployment/zk/DeploymentSupervisor.java
+++ b/spring-xd-dirt/src/main/java/org/springframework/xd/dirt/server/admin/deployment/zk/DeploymentSupervisor.java
@@ -456,9 +456,6 @@ public class DeploymentSupervisor implements ApplicationListener<ApplicationEven
 				jobDeployments = instantiatePathChildrenCache(client, Paths.JOB_DEPLOYMENTS);
 				jobDeployments.start(PathChildrenCache.StartMode.BUILD_INITIAL_CACHE);
 
-				deploymentQueueForConsumer = new DeploymentQueue(client, deploymentMessageConsumer, Paths.DEPLOYMENT_QUEUE,
-						executorService);
-				deploymentQueueForConsumer.start();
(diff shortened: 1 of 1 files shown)


cppwfs:
The update makes sense to me.  But would want someone who has further knowledge of Leadership logic to greenlight it.

spring-xd-dirt/src/main/java/org/springframework/xd/dirt/server/admin/deployment/zk/DeploymentSupervisor.java
```diff
@@ -459,6 +456,9 @@ public void takeLeadership(CuratorFramework client) throws Exception {
 				jobDeployments = instantiatePathChildrenCache(client, Paths.JOB_DEPLOYMENTS);
 				jobDeployments.start(PathChildrenCache.StartMode.BUILD_INITIAL_CACHE);
 
+				deploymentQueueForConsumer = new DeploymentQueue(client, deploymentMessageConsumer, Paths.DEPLOYMENT_QUEUE,
+						executorService);
+				deploymentQueueForConsumer.start();
```
pperalta: The exception was thrown because `ZKDeploymentHandler.moduleDeploymentRequests` was null. This reference is set when `onSupervisorElected` is invoked. That is invoked on line 468.

However, the deployment queue is still started before this event is fired. This shortens the window for this condition but it does not eliminate it.

I think I would move the queue creation and start to the bottom of this method (just before the sleep).

ilayaperumalg: I am fine with moving the consumer start part before sleep. Is the reason you are suggesting this because the `PathChildrenCache` start is asynchronous and we could delay starting the consumer for some more time? If this operation were synchronous, then we don't have to worry about this I think?

pperalta:
Squashed and merged as 4ee44a00e998224b163acd3660a34b3cb7f932a5

PR #1515, opened 2015-03-24
[master + backport] XD-2854 Check for ZK data with with no module defs

ilayaperumalg:
- Add null check if the module definitions weren't set for the given stream
 - Use `dummy` resources if the module definition is null

### DIFF
--- a/spring-xd-dirt/src/main/java/org/springframework/xd/dirt/stream/zookeeper/ZooKeeperStreamDefinitionRepository.java
+++ b/spring-xd-dirt/src/main/java/org/springframework/xd/dirt/stream/zookeeper/ZooKeeperStreamDefinitionRepository.java
@@ -21,10 +21,6 @@ import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 
-import com.fasterxml.jackson.core.type.TypeReference;
-import com.fasterxml.jackson.databind.ObjectMapper;
-import com.fasterxml.jackson.databind.ObjectReader;
-import com.fasterxml.jackson.databind.ObjectWriter;
 import org.apache.curator.framework.CuratorFramework;
 import org.apache.curator.framework.api.BackgroundPathAndBytesable;
 import org.apache.zookeeper.KeeperException.NoNodeException;
@@ -47,6 +43,11 @@ import org.springframework.xd.dirt.zookeeper.Paths;
 import org.springframework.xd.dirt.zookeeper.ZooKeeperConnection;
 import org.springframework.xd.dirt.zookeeper.ZooKeeperUtils;
 import org.springframework.xd.module.ModuleDefinition;
+
+import com.fasterxml.jackson.core.type.TypeReference;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.databind.ObjectReader;
+import com.fasterxml.jackson.databind.ObjectWriter;
 
 /**
  * @author Mark Fisher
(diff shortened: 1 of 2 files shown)


spring-xd-dirt/src/main/java/org/springframework/xd/dirt/stream/zookeeper/ZooKeeperStreamDefinitionRepository.java
```diff
@@ -154,8 +155,10 @@ public StreamDefinition findOne(String id) {
 			}
 			Map<String, String> map = ZooKeeperUtils.bytesToMap(bytes);
 			StreamDefinition streamDefinition = new StreamDefinition(id, map.get(DEFINITION_KEY));
-			List<ModuleDefinition> moduleDefinitions = objectReader.readValue(map.get(MODULE_DEFINITIONS_KEY));
-			streamDefinition.setModuleDefinitions(moduleDefinitions);
+			if (map.get(MODULE_DEFINITIONS_KEY) != null) {
```
cppwfs: I may be wrong (which wouldn't be the first time ), wouldn't we see NPE's down stream?  i.e. it will just get thrown somewhere else.

spring-xd-module/src/main/java/org/springframework/xd/module/ModuleDescriptor.java
```diff
@@ -575,6 +575,9 @@ public Builder setModuleDefinition(ModuleDefinition moduleDefinition) {
 		}
 
 		public ModuleDefinition getModuleDefinition() {
+			if (moduleDefinition == null) {
+				moduleDefinition = ModuleDefinitions.dummy(moduleName, type);
```
cppwfs: By adding this code will we still hit the filehandle issue they saw before?

spring-xd-dirt/src/main/java/org/springframework/xd/dirt/stream/zookeeper/ZooKeeperStreamDefinitionRepository.java
ilayaperumalg: The value will be null here only if the stream definition was stored using the previous version of XD. In other cases, we always set the module definitions based on the fix for XD-2731. This method handles the find() method only. In case of other loadStream operations, the null check here https://github.com/spring-projects/spring-xd/pull/1515#discussion_r26995231 would address it.

spring-xd-module/src/main/java/org/springframework/xd/module/ModuleDescriptor.java
ilayaperumalg: I hope my above comment addresses this as well.

cppwfs:
LGTM.  Ran regressions and they looked good.
I'll merge in the morning!

spring-xd-module/src/main/java/org/springframework/xd/module/ModuleDescriptor.java
ericbottard: Please don't use `ModuleDefinitions.dummy()`

If there really is a migration issue, then recreate the correct definitions (or tell the users to manually redeploy), but don't use dummy resources  in production code

ericbottard:
-1, see my comment

spring-xd-module/src/main/java/org/springframework/xd/module/ModuleDescriptor.java
markfisher: `ModuleDefinitions.dummy()` itself should be removed and replaced with something that is only on the src/test side. The fact that something "should not be used" is an indicator that it's in the wrong place.

cppwfs:
Removed my LGTM comment from before.  I've updated the https://jira.spring.io/browse/XD-2854 to included the steps to reproduce this bug.

ilayaperumalg:
ok, Closing this PR as the migration issue reported in https://jira.spring.io/browse/XD-2854 can only be addressed by following the steps mentioned there.

ilayaperumalg:
@ericbottard The creation of module definition is only needed on the getter as all it requires is the type information to be used during the deployment. Also, the module definition will be null only for the case where we port the stream definitions from previous versions of XD.

markpollack:
Squashed and merged on master as 387e2336d2ca39cc7848667e4e6a213d66279d26 and on 1.1.x as 61657f7998be1b5f8c2a8f7a8832f8cad492bdd8

PR #1522, opened 2015-03-25
XD-2854 [1.1.x] Add migration code changes

ilayaperumalg:
- This code checks for existing stream definitions and update
the module definitions if they don't exist.

### DIFF
--- a/spring-xd-dirt/src/main/java/org/springframework/xd/dirt/stream/AbstractDeployer.java
+++ b/spring-xd-dirt/src/main/java/org/springframework/xd/dirt/stream/AbstractDeployer.java
@@ -24,8 +24,6 @@ import java.util.Set;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 
-import org.apache.curator.framework.CuratorFramework;
-import org.apache.curator.framework.api.BackgroundPathAndBytesable;
 import org.apache.zookeeper.CreateMode;
 import org.apache.zookeeper.KeeperException;
 import org.slf4j.Logger;
@@ -44,12 +42,7 @@ import org.springframework.xd.dirt.zookeeper.Paths;
 import org.springframework.xd.dirt.zookeeper.ZooKeeperConnection;
 import org.springframework.xd.dirt.zookeeper.ZooKeeperUtils;
 import org.springframework.xd.module.ModuleDefinition;
-import org.springframework.xd.module.ModuleDefinitions;
 import org.springframework.xd.module.ModuleDescriptor;
-
-import com.fasterxml.jackson.core.type.TypeReference;
-import com.fasterxml.jackson.databind.ObjectMapper;
-import com.fasterxml.jackson.databind.ObjectWriter;
 
 /**
  * Abstract implementation of the @link {@link org.springframework.xd.dirt.core.ResourceDeployer} interface. It provides
--- a/spring-xd-dirt/src/main/java/org/springframework/xd/dirt/stream/AbstractDeployer.java
+++ b/spring-xd-dirt/src/main/java/org/springframework/xd/dirt/stream/AbstractDeployer.java
@@ -24,6 +24,8 @@ import java.util.Set;
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 
+import org.apache.curator.framework.CuratorFramework;
+import org.apache.curator.framework.api.BackgroundPathAndBytesable;
 import org.apache.zookeeper.CreateMode;
 import org.apache.zookeeper.KeeperException;
 import org.slf4j.Logger;
@@ -44,6 +46,10 @@ import org.springframework.xd.dirt.zookeeper.ZooKeeperUtils;
 import org.springframework.xd.module.ModuleDefinition;
 import org.springframework.xd.module.ModuleDefinitions;
 import org.springframework.xd.module.ModuleDescriptor;
+
+import com.fasterxml.jackson.core.type.TypeReference;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.databind.ObjectWriter;
 
 /**
  * Abstract implementation of the @link {@link org.springframework.xd.dirt.core.ResourceDeployer} interface. It provides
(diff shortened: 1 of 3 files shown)


spring-xd-dirt/src/main/java/org/springframework/xd/dirt/stream/StreamDeployer.java
```diff
@@ -86,4 +104,44 @@ protected String getDeploymentPath(StreamDefinition definition) {
 		return Paths.build(Paths.STREAM_DEPLOYMENTS, definition.getName());
 	}
 
+	/**
+	 * The migration code to run against existing stream definitions that don't have module definitions set already.
+	 * The following method will update the module definitions for those stream definitions.
+	 * See https://jira.spring.io/browse/XD-2854
+	 */
+	private void updateModuleDefinitions() {
+		if (this.parser != null && this.zkConnection != null && this.zkConnection.getClient() != null) {
```
pperalta: Are all of these null checks really required?

ilayaperumalg: No, the client (CuratorFramewor) was null in some of the tests. So, I added null check on `zkConnection.getClient()`. Otherwise, just to safeguard the migration operation (especially on the tests), I added the null check.

spring-xd-dirt/src/main/java/org/springframework/xd/dirt/stream/AbstractDeployer.java
```diff
@@ -24,6 +24,8 @@
 import java.util.regex.Matcher;
 import java.util.regex.Pattern;
 
+import org.apache.curator.framework.CuratorFramework;
+import org.apache.curator.framework.api.BackgroundPathAndBytesable;
```
markpollack: why all these new imports.  they don't seem to be used.

spring-xd-dirt/src/main/java/org/springframework/xd/dirt/stream/StreamDeployer.java
```diff
@@ -86,4 +104,44 @@ protected String getDeploymentPath(StreamDefinition definition) {
...
+			ObjectWriter objectWriter = new ObjectMapper().writerWithType(new TypeReference<List<ModuleDefinition>>() {});
```
cppwfs: Small detail but we could move the object writer after the first if statement.

spring-xd-dirt/src/main/java/org/springframework/xd/dirt/stream/AbstractDeployer.java
ilayaperumalg: yeah, Initially I had the migration code in here and then moved to `StreamDeployer`. Fixing this now.

spring-xd-dirt/src/main/java/org/springframework/xd/dirt/stream/StreamDeployer.java
```diff
@@ -86,4 +105,59 @@ protected String getDeploymentPath(StreamDefinition definition) {
...
+									List<ModuleDescriptor> moduleDescriptors = this.parser.parse(streamName,
+											definition.getDefinition(), definitionKind);
+									List<ModuleDefinition> moduleDefinitions = createModuleDefinitions(moduleDescriptors);
+									if (!moduleDefinitions.isEmpty()) {
+										map.put(DEFINITION_KEY, definition.getDefinition());
+										try {
+											map.put(MODULE_DEFINITIONS_KEY,
+													objectWriter.writeValueAsString(moduleDefinitions));
+										}
+										catch (JsonProcessingException jpe) {
```
pperalta: If a JSON exception is thrown, do we want to proceed with updating the stream anyway? Or should we skip the stream, log the error, and move on to the next one?

spring-xd-dirt/src/main/java/org/springframework/xd/dirt/stream/StreamDeployer.java
```diff
@@ -86,4 +105,59 @@ protected String getDeploymentPath(StreamDefinition definition) {
...
+											logger.error("Error writing module definitions " + moduleDefinitions +
```
pperalta: The exception should be passed to the logger so we don't lose the stack trace.

spring-xd-dirt/src/main/java/org/springframework/xd/dirt/stream/StreamDeployer.java
```diff
@@ -86,4 +105,59 @@ protected String getDeploymentPath(StreamDefinition definition) {
...
+										}
+										byte[] binary = ZooKeeperUtils.mapToBytes(map);
+										BackgroundPathAndBytesable<?> op = client.checkExists().forPath(path) == null
+												? client.create() : client.setData();
+										op.forPath(path, binary);
+									}
+								}
+							}
+						}
+						catch(Exception e) {
```
pperalta: Missing space after `catch`.

spring-xd-dirt/src/main/java/org/springframework/xd/dirt/stream/StreamDeployer.java
```diff
@@ -86,4 +105,59 @@ protected String getDeploymentPath(StreamDefinition definition) {
...
+							logger.error("Exception when updating module definitions for the stream " + streamName );
```
pperalta: Exception should be passed to the log message, extra space after paren.

spring-xd-dirt/src/main/java/org/springframework/xd/dirt/stream/StreamDeployer.java
```diff
@@ -59,7 +75,9 @@
 	public StreamDeployer(ZooKeeperConnection zkConnection, StreamDefinitionRepository repository,
 			StreamRepository streamRepository, XDParser parser) {
 		super(zkConnection, repository, streamRepository, parser, stream);
+		this.zkConnection = zkConnection;
 		this.parser = parser;
+		updateModuleDefinitions();
```
pperalta: I don't think this method should be called in a constructor.

spring-xd-dirt/src/main/java/org/springframework/xd/dirt/stream/StreamDeployer.java
ilayaperumalg: Yeah, we can skip the stream as the `streamDefinition` key should have been there in ZK data.

spring-xd-dirt/src/main/java/org/springframework/xd/dirt/stream/StreamDeployer.java
ilayaperumalg: Yeah, I agree. I am fixing this now with the use of `@PostConstruct`

spring-xd-dirt/src/main/java/org/springframework/xd/dirt/stream/StreamDeployer.java
ilayaperumalg: Since this exception could happen, we would still need to handle `null` value for ModuleDefinition in ModuleDescriptor. I will add that check.

markpollack:
squashed and merged as ce5ff05e87e7b0a917811104ab41e87d82e3419a

PR #1525, opened 2015-03-25
XD-2864 added check for @Configuration module

dturanski:
Should be backported


markpollack:
merged into master and into 1.1.x as 81b652a2f29e16cf8f40dc2c9005b03a2c503069

PR #1527, opened 2015-03-26
[Sprint: 46] XD-2868: Fix Local Bus and Partitioned Jobs

garyrussell:
JIRA: https://jira.spring.io/browse/XD-2868

When using a `LocalMessageBus` with partitioned jobs:

In 1.0.x, the request/reply binding used an unbounded task executor, causing
many/all partitions in a partitioned job to be processed concurrently.

In 1.1.x, the executor was changed to a thread pool, but with default configuration
that causes a single thread to be used by default = no concurrency. The executor
could be reconfigured, but it was bus-wide and thus users could not control the
concurrency for an individual job.

Change the local bus such that partitioned jobs each get a dedicated thread pool
(with configurable concurrency). Destroy the thread pool when the job is undeployed.

Note that I uncovered a problem with binding partitioned jobs where the producer instead
of consumer properties were passed in on the replier side.


mbogoevici:
Rebased, merged w/https://github.com/spring-projects/spring-xd/commit/fbd9272bb51f34604c9efd956da918913725effd

PR #1528, opened 2015-03-26
[Sprint: 46] XD-2829: Support XML Payloads in xd/lib

garyrussell:
JIRA: https://jira.spring.io/browse/XD-2829

Add spring-integration-xml (and hence spring-xml) to xd/lib
to support `xpath` expressions in standard modules that support
expressions (transform, filter etc).


dturanski:
LGTM; merged

PR #1532, opened 2015-03-30
[Sprint: 46] XD-2777: Document Trigger Source

garyrussell:
JIRA: https://jira.spring.io/browse/XD-2777

Add documentation for trigger source.

Change the way the default date value is set up to provide a useful default in the documentation.

Tested with

    xd:>stream create foo --definition "trigger --payload=foo | log" --deploy

for immediate trigger, and

    xd:>stream create foo --definition "trigger --payload=foo --date='03/30/15 17:52:00' | log" --deploy

for a delayed trigger.

### DIFF
--- a/modules/source/trigger/config/trigger.xml
+++ b/modules/source/trigger/config/trigger.xml
@@ -12,15 +12,15 @@
 	</int:inbound-channel-adapter>
 
 	<beans profile="use-date">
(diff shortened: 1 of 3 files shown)


ericbottard:
Merged as is as 8c2c021.
Will create another JIRA/PR to
* handle cron validation early
* handle 1 out of 3 mutually exclusive options

PR #1533, opened 2015-03-30
[Sprint: 47] XD-2852 XD-2298 A basic gpload tasklet implementation

trisberg:
- provides option for overriding INPUT.SOURCE.FILE as a job parameter


src/docs/asciidoc/Batch-Jobs.asciidoc
```diff
@@ -738,3 +738,77 @@ $$resourceManagerPort$$:: $$the Port for Hadoop's ResourceManager$$ *($$String$$
...
 //$job.sqoop
+
+==== Running gpload as a batch job (`gpload`)
+The gpload utility can be deployed and launched from Spring XD as a batch job. The gpload job uses a `GploadTasklet` that submits a gpload job as an external process. The Spring XD gpload batch job aims to support most of the gpload functionality.
+
+We need to provide the following required options:
+
+- `gploadHome` - this must be the path to where gpload utility is installed. This is usually /usr/local/greenplum-loaders-<version>.
+- `controlFile` - this file defines the gpload options in effect for this load job and is documented in the _Greenplum Load Tools Reference_ documentation.
+- `password` or `passswordFile` - you can either speciy the passord or provide a password file that must follow he general format for a PostgreSQL password file.
```
jvalkeal: 'he' typo

extensions/spring-xd-extension-gpload/src/main/java/org/springframework/xd/gpload/GploadEnvironmentProvider.java
```diff
@@ -0,0 +1,50 @@
...
+
+	public GploadEnvironmentProvider(String gploadHome, boolean usePasswordFile, String passwordInfo) {
+		this.gploadHome = gploadHome;
+		this.usePasswordFile = usePasswordFile;
+		this.passwordInfo = passwordInfo;
+	}
+
+	@Override
+	public void setEnvironment(Map<String, String> env) {
+		env.put("GPHOME_LOADERS", gploadHome);
```
jvalkeal: Don't do this because it will result errors in system which don't have this. I have a GPDB demo vm's and its `gpload` is:
```
$ cat /usr/local/greenplum-db/bin/gpload
#!/bin/sh
if [ ! -z "$GPHOME" ]; then
    . $GPHOME/greenplum_path.sh
fi
if [ ! -z "$GPHOME_LOADERS" ]; then
    . $GPHOME_LOADERS/greenplum_loaders_path.sh
fi
if [ `uname -s` = "AIX" -a ! -z "$GP_LIBPATH_FOR_PYTHON" ]; then
    LIBPATH=$GP_LIBPATH_FOR_PYTHON:$LIBPATH
    export LIBPATH
fi
```

Don't have `greenplum_loaders_path.sh` and job then fails with:
```
2015-04-08 03:29:14,432 1.2.0.SNAP ERROR task-scheduler-3 gpload.GploadTasklet - gpload job finished with exit code: 1
2015-04-08 03:29:14,434 1.2.0.SNAP ERROR task-scheduler-3 gpload.GploadTasklet - gpload err: /usr/local/greenplum-db/bin/gpload: line 6: /usr/local/greenplum-db/greenplum_loaders_path.sh: No such file or directory
```

extensions/spring-xd-extension-gpload/src/main/java/org/springframework/xd/gpload/GploadTasklet.java
```diff
@@ -0,0 +1,294 @@
...
+			throw new IllegalArgumentException("You can't specify both 'password' and 'passwordFile' options");
+		}
+		if (StringUtils.hasText(password)) {
+			addEnvironmentProvider(new GploadEnvironmentProvider(gploadHome, false, password));
+		}
+		else if (StringUtils.hasText(passwordFile)) {
+			addEnvironmentProvider(new GploadEnvironmentProvider(gploadHome, true, passwordFile));
+		}
+		else {
+			throw new IllegalArgumentException("You must specify either 'password' or 'passwordFile' option");
```
jvalkeal: I surely don't need password or passwordFile so forcing to set it is a bit pita.

jvalkeal:
Other than that, gpload job ran fine.

extensions/spring-xd-extension-gpload/src/main/java/org/springframework/xd/gpload/GploadEnvironmentProvider.java
trisberg: Hmm, that's a problem with that startup script. I guess we could check for ${gploadHome}/greenplum_loaders_path.sh and if it is not there check for ${gploadHome}/greenplum_path.sh and set GPHOME in that case.

extensions/spring-xd-extension-gpload/src/main/java/org/springframework/xd/gpload/GploadTasklet.java
trisberg: The password handling is bad overall since there is no way to specify to never ever prompt for password. So the best compromise I could come up with was to force either password or passwordFile to avoid the majority of situations where we would get that password prompt.

jvalkeal: Oh yeah, I remember now when you were having a fight with passwords.

trisberg:
I pushed a new commit where I check for the existence of the greenplum_path.sh or greenplum_loaders_path.sh and set the appropriate env var based on that. We might want to add Windows file name support at some point if we want to support that.

jvalkeal:
forget to say that lgtm.

jvalkeal:
merged

PR #1534, opened 2015-03-30
[Sprint: 46] XD-2854 Migrate Stream definitions that don't have MDs

ilayaperumalg:
- This code checks for existing stream definitions and update
the module definitions if they don't exist.
 - Add better exception handling
 - Use PostConstruct to run migration

### DIFF
--- a/spring-xd-dirt/src/main/java/org/springframework/xd/dirt/stream/StreamDeployer.java
+++ b/spring-xd-dirt/src/main/java/org/springframework/xd/dirt/stream/StreamDeployer.java
@@ -21,6 +21,26 @@ import static org.springframework.xd.dirt.stream.ParsingContext.stream;
 import org.springframework.xd.dirt.server.admin.deployment.DeploymentHandler;
 import org.springframework.xd.dirt.zookeeper.Paths;
 import org.springframework.xd.dirt.zookeeper.ZooKeeperConnection;
+import java.util.List;
+import java.util.Map;
+
+import javax.annotation.PostConstruct;
+
+import org.apache.curator.framework.CuratorFramework;
+import org.apache.curator.framework.api.BackgroundPathAndBytesable;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
+
+import org.springframework.xd.dirt.zookeeper.Paths;
+import org.springframework.xd.dirt.zookeeper.ZooKeeperConnection;
+import org.springframework.xd.dirt.zookeeper.ZooKeeperUtils;
+import org.springframework.xd.module.ModuleDefinition;
+import org.springframework.xd.module.ModuleDescriptor;
+
+import com.fasterxml.jackson.core.JsonProcessingException;
+import com.fasterxml.jackson.core.type.TypeReference;
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.databind.ObjectWriter;
 
 /**
  * Default implementation of {@link StreamDeployer} that uses provided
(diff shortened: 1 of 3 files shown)


spring-xd-dirt/src/main/java/org/springframework/xd/dirt/stream/StreamDeployer.java
```diff
@@ -80,4 +114,70 @@ protected String getDeploymentPath(StreamDefinition definition) {
...
+	 * See https://jira.spring.io/browse/XD-2854
+	 */
+	@PostConstruct
+	private void updateModuleDefinitions() {
+		if (this.zkConnection.getClient() != null) {
+			try {
+				CuratorFramework client = this.zkConnection.getClient();
+				if (client.checkExists().forPath(Paths.STREAMS) != null) {
+					Iterable<StreamDefinition> streamDefinitions = findAll();
+					for (StreamDefinition definition : streamDefinitions) {
```
pperalta: These two lines can be simplified to:
```
for (StreamDefinition definition : findAll()) {
```

spring-xd-dirt/src/main/java/org/springframework/xd/dirt/stream/StreamDeployer.java
```diff
@@ -80,4 +114,70 @@ protected String getDeploymentPath(StreamDefinition definition) {
...
+						setModuleDefinitions(client, definition);
+					}
+				}
+			}
+			catch (Exception e) {
+				logger.error("Exception migrating stream definitions. This migration is done when the existing " +
+						"stream definitions that don't have module definitions set. " + e);
```
pperalta: Pass the exception as a parameter to the logger so that we don't lose the stack trace.

As in, do `", e);` instead of `" + e);`.

spring-xd-dirt/src/main/java/org/springframework/xd/dirt/stream/StreamDeployer.java
```diff
@@ -80,4 +114,70 @@ protected String getDeploymentPath(StreamDefinition definition) {
...
+	 *
+	 * @param client the curator framework client
+	 * @param definition the stream definition
+	 */
+	private void setModuleDefinitions(CuratorFramework client, StreamDefinition definition) {
+		String streamName = definition.getName();
+		String path = Paths.build(Paths.STREAMS, streamName);
+		try {
+			byte[] bytes = client.getData().forPath(path);
+			if (bytes != null) {
```
pperalta: If the stream node exists but is empty, would it be desirable to populate it with the new format? If so, here's my suggestion:

```
byte[] bytes = client.getData().forPath(path);
Map<String, String> map = bytes == null
		? new HashMap<String, String>()
		: ZooKeeperUtils.bytesToMap(bytes);
```

spring-xd-dirt/src/main/java/org/springframework/xd/dirt/stream/StreamDeployer.java
```diff
@@ -80,4 +114,70 @@ protected String getDeploymentPath(StreamDefinition definition) {
...
+				Map<String, String> map = ZooKeeperUtils.bytesToMap(bytes);
+				if (map.get(MODULE_DEFINITIONS_KEY) == null) {
```
pperalta: Perhaps we should add a comment here indicating the old format vs the new format?

spring-xd-dirt/src/main/java/org/springframework/xd/dirt/stream/StreamDeployer.java
```diff
@@ -80,4 +114,70 @@ protected String getDeploymentPath(StreamDefinition definition) {
...
+						try {
+							map.put(MODULE_DEFINITIONS_KEY,
+									objectWriter.writeValueAsString(moduleDefinitions));
+							byte[] binary = ZooKeeperUtils.mapToBytes(map);
+							BackgroundPathAndBytesable<?> op = client.checkExists().forPath(path) == null
+									? client.create() : client.setData();
+							op.forPath(path, binary);
+						}
+						catch (JsonProcessingException jpe) {
+							logger.error("Exception writing module definitions " + moduleDefinitions +
```
pperalta: Same comment regarding exception logging as above.

spring-xd-dirt/src/main/java/org/springframework/xd/dirt/stream/StreamDeployer.java
```diff
@@ -80,4 +114,70 @@ protected String getDeploymentPath(StreamDefinition definition) {
...
+									" for the stream " + streamName + " : " + jpe);
+						}
+					}
+				}
+			}
+		}
+		catch (Exception e) {
+			logger.error("Exception when updating module definitions for the stream "
+					+ streamName + " : " + e);
```
pperalta: Same comment regarding exception logging as above.

spring-xd-dirt/src/main/java/org/springframework/xd/dirt/stream/StreamDeployer.java
ilayaperumalg: If the stream node exists then it would never be empty as it will always at least have the stream definition with it.

pperalta:
Squashed & merged as 0d184685af1d5d3e26e8822e5ff66027336cef4c

PR #1535, opened 2015-03-31
XD-2885: Fix modules packaging to only include relevant files

ericbottard:
You'll also want to make sure that you don't have `${checkout_dir}/modules/lib`, as this was a previous bogus dir (fixed in a different PR), but not cleaned. This can safely be `rm -fR`ed, and it should not re-appear (and not be packaged)


dturanski:
merged

PR #1538, opened 2015-04-02
Allow "--path=..." configuration in HTTP channel adapter

whummer:
Currently, each HTTP stream channel uses its own port, which is a limitation if we want to use large numbers of HTTP streams with fine granularity. It would be nice to (additionally) use URI paths to identify HTTP channels, allowing to re-use a single HTTP port for multiple channels in XD. 

Example usage:
```
stream create --name s1 --definition "http --port=9495 --path=/foo  | log" --deploy
```

This pull request contains a working proof-of-concept implementation, including a set of simple integration tests. 

Any comments/suggestions welcome! Thanks

### DIFF
--- a/extensions/spring-xd-extension-http/src/main/java/org/springframework/integration/x/http/NettyHttpInboundChannelAdapter.java
+++ b/extensions/spring-xd-extension-http/src/main/java/org/springframework/integration/x/http/NettyHttpInboundChannelAdapter.java
@@ -24,8 +24,13 @@ import static org.jboss.netty.handler.codec.http.HttpResponseStatus.OK;
 import static org.jboss.netty.handler.codec.http.HttpResponseStatus.REQUEST_ENTITY_TOO_LARGE;
 import static org.jboss.netty.handler.codec.http.HttpVersion.HTTP_1_1;
 
+import java.lang.reflect.Method;
 import java.net.InetSocketAddress;
 import java.security.KeyStore;
+import java.util.HashMap;
+import java.util.LinkedList;
+import java.util.List;
+import java.util.Map;
 import java.util.Properties;
 import java.util.concurrent.Executor;
 import java.util.concurrent.Executors;
@@ -109,6 +114,13 @@ public class NettyHttpInboundChannelAdapter extends MessageProducerSupport {
 
 	private final int port;
 
+	/* begin whu */
+	private final String path;
+
+	private static final String KEY_MODULE_STATE = NettyHttpInboundChannelAdapter.class.getName() + ".moduleState";
+
+	/* end whu */
+
 	private final boolean ssl;
 
 	private volatile ServerBootstrap bootstrap;
@@ -139,9 +151,32 @@ public class NettyHttpInboundChannelAdapter extends MessageProducerSupport {
 	}
 
 	public NettyHttpInboundChannelAdapter(int port, boolean ssl) {
+		/* begin whu */
+		this(port, "/", ssl);
+		/* end whu */
+	}
+
+	/* begin whu */
+	public NettyHttpInboundChannelAdapter(int port, String path) {
+		this(port, path, false);
+	}
+
+	public NettyHttpInboundChannelAdapter(int port, String path, boolean ssl) {
 		this.port = port;
 		this.ssl = ssl;
+		this.path = path;
+		putToChannelAdapter();
 	}
+
+	private void putToChannelAdapter() {
+		Map<String, Object> pathToProducer = getModuleState();
+		if (pathToProducer.containsKey(path)) {
+			logger.warn("Channel for path '" + path + "' already exists, overwriting.");
+		}
+		pathToProducer.put(path, this);
+	}
+
+	/* end whu */
 
 	/**
 	 *
@@ -198,11 +233,21 @@ public class NettyHttpInboundChannelAdapter extends MessageProducerSupport {
 			this.messageConverter = new NettyInboundMessageConverter(getMessageBuilderFactory());
 		}
 		executionHandler = new ExecutionHandler(executor);
-		bootstrap = new ServerBootstrap(new NioServerSocketChannelFactory(Executors.newCachedThreadPool(),
-				Executors.newCachedThreadPool()));
-		bootstrap.setOption("child.tcpNoDelay", true);
-		bootstrap.setPipelineFactory(new PipelineFactory());
-		bootstrap.bind(new InetSocketAddress(this.port));
+		/* begin whu */
+		try {
+			/* end whu */
+			bootstrap = new ServerBootstrap(new NioServerSocketChannelFactory(Executors.newCachedThreadPool(),
+					Executors.newCachedThreadPool()));
+			bootstrap.setOption("child.tcpNoDelay", true);
+			bootstrap.setPipelineFactory(new PipelineFactory());
+			bootstrap.bind(new InetSocketAddress(this.port));
+			/* begin whu */
+		}
+		catch (Exception e) {
+			/* bind exception -> server already created by another channel */
+			bootstrap = null;
+		}
+		/* end whu */
 	}
 
 	@Override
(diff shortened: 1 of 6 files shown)


garyrussell:
First of all, thanks for the contribution; I think this is useful functionality but, as you say, this is more of a PoC than a PR ready to merge. I don't think it needs a huge amount of work, but I have some concerns in some areas of your code.

Second, I suggest you read the [contributor guidelines](https://github.com/spring-projects/spring-xd/blob/master/CONTRIBUTING.md). We generally prefer to accept PRs from topic branches rather than your master; it's not that it's easier for us necessarily, it's easier for you when your code is eventually merged to master after polishing and squashing. When submitting PRs from master, you will likely be forced to reset your master, rewriting history. We also prefer to track features like this through JIRA so that the feature appears in the release contents. We will open a story in JIRA for this feature.

I will be making some general comments on your code, but it will likely be tomorrow, now.

Thanks again; looking forward to seeing more!!

whummer:
Thanks for the response. Apologies that my PR was a bit of a quick shot, off the guidelines. I signed the CLA (let me know if you need the confirmation number):

```
```

How do we proceed - I'd be happy if you could take it from there, or do I need to improve the PR, make it comply with the guidelines, etc.? Looking forward to your further comments. Thanks

whummer:
Could not find existing JIRA issue, created a new one: https://jira.spring.io/browse/XD-2976

garyrussell:
Great; thanks - we can continue to work with this PR for now - I'll make my comments and we can take it from there. Github doesn't allow you to move a PR to a new source branch. After I make my comments, it's up to you whether you want to add a new commit here, or close this one and start a new one from branch XD-2976.

extensions/spring-xd-extension-http/src/main/java/org/springframework/integration/x/http/NettyHttpInboundChannelAdapter.java
```diff
@@ -109,6 +114,13 @@
 
 	private final int port;
 
+	/* begin whu */
```
garyrussell: We don't need these `begin`, `end` comments - `git blame` provides adequate history as to who changed what code. IDEs such as eclipse even provide a mechanism to show blame in the editor pane.

extensions/spring-xd-extension-http/src/main/java/org/springframework/integration/x/http/NettyHttpInboundChannelAdapter.java
```diff
@@ -139,10 +151,35 @@ public NettyHttpInboundChannelAdapter(int port) {
...
+	}
+
+	public NettyHttpInboundChannelAdapter(int port, String path, boolean ssl) {
 		this.port = port;
 		this.ssl = ssl;
+		this.path = path;
+		putToChannelAdapter();
 	}
 
+	private void putToChannelAdapter() {
```
garyrussell: Although deployments are currently single-threaded, it would be safer to synchronize this method.

Also, perhaps something like `ensurePathIsUnique()` would be a more suitable method name.

extensions/spring-xd-extension-http/src/main/java/org/springframework/integration/x/http/NettyHttpInboundChannelAdapter.java
```diff
@@ -198,11 +235,21 @@ protected void doStart() {
...
+			bootstrap = new ServerBootstrap(new NioServerSocketChannelFactory(Executors.newCachedThreadPool(),
+					Executors.newCachedThreadPool()));
+			bootstrap.setOption("child.tcpNoDelay", true);
+			bootstrap.setPipelineFactory(new PipelineFactory());
+			bootstrap.bind(new InetSocketAddress(this.port));
+			/* begin whu */
+		}
+		catch (Exception e) {
+			/* bind exception -> server already created by another channel */
+			bootstrap = null;
```
garyrussell: This needs to be fatal - we can't just eat the exception.

extensions/spring-xd-extension-http/src/main/java/org/springframework/integration/x/http/NettyHttpInboundChannelAdapter.java
```diff
@@ -210,6 +257,8 @@ protected void doStop() {
 		if (bootstrap != null) {
 			bootstrap.shutdown();
 		}
+		Map<String, Object> pathToProducer = getModuleState();
+		pathToProducer.remove(path);
```
garyrussell: This needs to be symmetrical with the `put()`, which is currently done in the constructor.

We could implement `DisposableBean` and do this in the `destroy()` method, but it would probably be better to move the `ensurePathIsUnique()` logic (`put`) to `doStart()`.

extensions/spring-xd-extension-http/src/main/java/org/springframework/integration/x/http/NettyHttpInboundChannelAdapter.java
garyrussell: Oh, I see why you're catching it now. Hmmm...

This won't work; consider...

deploy stream a
deploy stream b
undeploy stream a

the bootstrap will be torn down, breaking stream b

I believe we need to store the bootstrap in some shared state and only shut it down when the last module using that port is undeployed.

extensions/spring-xd-extension-http/src/main/java/org/springframework/integration/x/http/NettyHttpInboundChannelAdapter.java
```diff
@@ -139,10 +151,35 @@ public NettyHttpInboundChannelAdapter(int port) {
...
 	}
 
+	private void putToChannelAdapter() {
+		Map<String, Object> pathToProducer = getModuleState();
+		if (pathToProducer.containsKey(path)) {
+			String msg = "Channel for HTTP port " + port + ", path '" + path + "' already exists.";
+			logger.warn(msg);
+			throw new RuntimeException(msg);
+		}
+		pathToProducer.put(path, this);
```
garyrussell: It would be better to store the `Method` instead of `this` - avoids the reflection to look up the method on every request.

extensions/spring-xd-extension-http/src/main/java/org/springframework/integration/x/http/NettyHttpInboundChannelAdapter.java
```diff
@@ -286,6 +335,58 @@ public void writeRequested(ChannelHandlerContext ctx, MessageEvent e) throws Exc
...
 
+	/* start whu */
+	void wrapSendMessage(Message<?> message) {
+		sendMessage(message);
+	}
+
+	private boolean sendMessageToResponsibleSender(HttpRequest request, Message<?> message) {
+		String name = "wrapSendMessage";
+		Object sender = getResponsibleSender(request);
+		for (Method m : sender.getClass().getDeclaredMethods()) {
```
garyrussell: See my note above about storing the method instead of the "sender" to avoid this each time.

Also, consider using `ReflectionUtils.doWithMethods()` with a method filter instead of this raw iteration.

extensions/spring-xd-extension-http/src/main/java/org/springframework/integration/x/http/NettyHttpInboundChannelAdapter.java
```diff
@@ -286,6 +335,58 @@ public void writeRequested(ChannelHandlerContext ctx, MessageEvent e) throws Exc
...
+		}
+		if (matches.size() > 1) {
+			logger.warn("Ambiguous HTTP path, multiple matches " + matches + " for path '" + uri + "'");
+		}
+		return result;
+	}
+
+	@SuppressWarnings("unchecked")
+	private static Map<String, Object> getModuleState() {
+		/* bit hacky, store module state in system properties */
```
garyrussell: Yeah - we need a better solution than storing this in system properties.

Perhaps we can add a general `<util:map id="moduleState"/>` to the global context, allowing modules to use it as a general bucket for shared state. Or, perhaps it needs to go in zookeeper, but that might be overkill - I'll discuss it with the rest of the XD team.

extensions/spring-xd-extension-http/src/main/java/org/springframework/integration/x/http/NettyHttpInboundChannelAdapter.java
```diff
@@ -109,6 +114,13 @@
```
garyrussell: Add your name to the `author` list for each changed Java source file.

extensions/spring-xd-extension-http/src/main/java/org/springframework/integration/x/http/NettyHttpInboundChannelAdapter.java
whummer: Thanks for the helpful comments. How can we proceed with this? Did you have a chance to discuss with the team? Some sort of built-in support between shared state between the modules would be ideal.

garyrussell: Unfortunately I have been tied up on other stuff; I'll try to get to this again this week.

markpollack:
Hi.  We have had several discussions about this and don't want to store state to support this feature.  ATM would recommend putting something like nginx in front of XD to split out the two paths.  Going forward I think you will like what we have in store for spring xd 2.0, as it would allow you to have the full Spring MVC infrastructure as part of a module.  A POC of to explore this space is here - https://github.com/scratches/spring-bus-demo/blob/master/voter/src/main/java/demo/VoterApplication.java

PR #1539, opened 2015-04-03
[Sprint: 47] [master][backport]XD-2879 Customizable Kafka message bus partition setting

mbogoevici:
Add `minKafkaPartitions` property that can be used on a Kafka Bus producer for (optionally) overpartitioning the bus. In case the module/partition count and concurrency are higher than this value, it will be overriden.

### DIFF
--- a/config/servers.yml
+++ b/config/servers.yml
@@ -91,6 +91,7 @@
 #        autoCommitEnabled:         true
 #        queueSize:                 1000
 #        fetchSize:                 1048576
+#        minPartitionCount:         1
 
 #Disable batch database initialization
 #spring:
--- a/spring-xd-dirt/src/main/resources/application.yml
+++ b/spring-xd-dirt/src/main/resources/application.yml
@@ -169,6 +169,7 @@ xd:
         autoCommitEnabled:         true
         queueSize:                 1000
         fetchSize:                 1048576
+        minPartitionCount:         1
   security:
     authorization:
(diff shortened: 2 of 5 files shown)


config/servers.yml
```diff
@@ -91,6 +91,7 @@
 #        autoCommitEnabled:         true
 #        queueSize:                 1000
 #        fetchSize:                 1048576
+#        minPartitionCount:         1
```
ilayaperumalg: Does this `minPartitionCount` need to be deployment property so that it can be specified for specific stream deployment rather than message bus specific property?
Also, lets say if someone needs to have higher `minPartitionCount` for a given stream but a very low value for other, then having this at the message bus level may not fit well. what do you think?

mbogoevici: Actually, it is a deployment property, set via `module.x.producer.minPartitionCount` (and wildcarded equivalents - note how it is used in the tests). What's in `servers.yml` is the default, which is normally 1. We could hardcode it, too, but I guess it doesn't hurt to make that configurable.

ilayaperumalg: yeah, right. I missed that one.

ilayaperumalg:
LGTM, merged into master a2b0107bb2efab1bc43dca98fedeb94dac80a7ed as well as 1.1.x ea8611046c2687382b4b394521dc4cc2eb07d8c6

PR #1540, opened 2015-04-03
XD-2893: Prettify&escape module info defaults

ericbottard:


markpollack:
The patch is causing not all the lines to be colored correctly (first line is in green ,the rest are in white) and the tab spacing is broken.  Is the only purpose of this to make <none> appear in yellow?   Here is a before/after screenshot 

https://github.com/markpollack/spring-xd-images/blob/master/module-info-color-pr.png

ericbottard:
This is not only about colors, it is mainly about replacing a literal `\n` (that would cause an actual carriage return) with a `\\n` (hence displayed as `\n`). The coloration was so that it is not mistaken for backslash + n but it proves to be more complicated that expected at the moment (there is no "go back to where you were" in ANSI codes, just "reset everything".

I'll rework the PR to at least fix the coloring reset, pending a better solution.

ericbottard:
Rebased and removed color

markpollack:
Tested looking at gpfdist sink output

````
  dbUser        database user                                          gpadmin    String
  delimiter     data line delimiter                                    \n         String
````

merged as 82fe3fc82b7e1cc615c61290cbc8ecba2fe64b8a

PR #1543, opened 2015-04-06
XD-2815: Refactored jdbchdfs job to not use Tuple data structure

mminella:
Removing the use of a `Tuple` in the processing of this job and just formatting the output in the `RowMapper` used in the reader provides significant performance benefits.

### DIFF
--- a/extensions/spring-xd-extension-jdbc/src/main/java/org/springframework/xd/jdbc/NamedColumnJdbcItemReaderFactory.java
+++ b/extensions/spring-xd-extension-jdbc/src/main/java/org/springframework/xd/jdbc/NamedColumnJdbcItemReaderFactory.java
@@ -59,6 +59,8 @@ public class NamedColumnJdbcItemReaderFactory implements FactoryBean<NamedColumn
 
 	private NamedColumnJdbcItemReader reader;
 
+	private String delimiter;
+
 	@Override
 	public NamedColumnJdbcItemReader getObject() throws Exception {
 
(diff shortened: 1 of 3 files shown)


extensions/spring-xd-extension-jdbc/src/main/java/org/springframework/xd/jdbc/NamedColumnJdbcItemReaderFactory.java
```diff
@@ -59,6 +59,8 @@
 
 	private NamedColumnJdbcItemReader reader;
 
+	private String delimiter;
```
markfisher: should there be a default?

mminella: I thought the correct way to handle defaults was in the option classes.  If that isn't correct, I can fix...

trisberg: I think the way it is handled is correct. We might want to enclose string values in quotes to avoid issues with values containing the delimiter character. Then we have the problem with strings containing new line and tab characters. I guess that would be a different JIRA though since that would be new functionality.

trisberg:
LGTM, merging ...

trisberg:
merged

PR #1548, opened 2015-04-08
[Sprint: 49] [master + backport] XD-2924 Tap spark processor output

ilayaperumalg:
- Add support to tap spark streaming processor module's output
   - Add wiretap channel to processor module output channel
   - Bind wiretap channel to messagebus pub/sub producer
   - Unbind/close wiretap channel upon module stop

### DIFF
--- a/spring-xd-dirt/src/main/java/org/springframework/xd/dirt/plugins/AbstractMessageBusBinderPlugin.java
+++ b/spring-xd-dirt/src/main/java/org/springframework/xd/dirt/plugins/AbstractMessageBusBinderPlugin.java
@@ -382,10 +382,9 @@ public abstract class AbstractMessageBusBinderPlugin extends AbstractPlugin {
 	/**
 	 * Event handler for tap additions.
 	 *
-	 * @param client curator client
 	 * @param data module data
 	 */
-	private void onTapAdded(CuratorFramework client, ChildData data) {
+	private void onTapAdded(ChildData data) {
 		String tapChannelName = buildTapChannelNameFromPath(data.getPath());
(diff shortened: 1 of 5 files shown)


ilayaperumalg:
Please note I haven't added the logic to use TapListeners to update ZK path in case of taps created on spark streaming processor's output channel. Since the `MessageBusSender` (where the output channel is bound to the messagebus) is running on spark cluster, building/binding the tap is done at the MessageBusSender itself.

markfisher:
I understand not wanting to connect to ZK from the module running on the Spark cluster just for the functionality where we enable publishing to a tap channel only when at least 1 subscriber is active. However, it's not very likely that a tap will be used at that location, definitely not the "normal" case. I wonder if we should instead support a boolean option for Spark Streaming modules that would explicitly enable the publishing to a tap channel. It would be FALSE by default, but if someone deploying the module knows they are going to create taps off that pipe, the could set that option value to TRUE.

ilayaperumalg:
@markfisher yeah, that's a good idea. I will work on adding an explicit option to create tap.

ilayaperumalg:
@markfisher Added support to enable tap on the spark processor output via a module option named `enableTap`. Please note that since this support uses module option, one can specify it for the sink module as well (with no consequences). I have created a JIRA to differentiate the module option classes for spark processor/sink modules: https://jira.spring.io/browse/XD-3002

markfisher:
LGTM

ilayaperumalg:
Squashed and merged as 32bda437ffdccdd0828804583377acd82c4a7f53. Also, back-ported to 1.1.x as 39539aeaac3bdbd12a5eb38710c6dfb6569543ff

PR #1553, opened 2015-04-10
[Sprint: 47] XD-2911 Improve performance of Tuple construction

mminella:
The Tuple interface requires that a Tuple be able to do a collection of
type conversions and formatting options as part of it's core
functionality.  The previous means for accomplishing this functionality
was via an extension of the DefaultFormattingConversionService.  This
commit changes that to a custom implementation of
ConfigurableConversionService.  This change is done to limit the number
of Converter instances that are created for each Tuple instance.

Other minor enhancements have been made around List size allocations and
lazy creation of some other instances (ObjectMapper, etc).

While this PR addresses some of the performance concerns with building Tuples, a design discussion should occur to better address the way the conversion behavior is provided for Tuples by Spring XD.


mminella:
Closing for a different approach

PR #1554, opened 2015-04-11
XD-2942 - add ftp source module

fmarchand:
I have signed and agree to the terms of the SpringSource Individual
Contributor License Agreement.

Tell me if I forgot something. Thx

PR #1557, opened 2015-04-15
XD-2942: Add ftp source.

ericbottard:
Based off work submitted by fmarchand (https://github.com/spring-projects/spring-xd/pull/1554)
Added e2e test, refactored module options some more


spring-xd-dirt/src/main/java/org/springframework/xd/dirt/modules/metadata/FtpSourceOptionsMetadata.java
```diff
@@ -0,0 +1,170 @@
...
+    public void setTmpFileSuffix(String tmpFileSuffix) {
+        this.tmpFileSuffix = tmpFileSuffix;
+    }
+
+    @Min(0)
+    public int getFixedRate() {
+        return fixedRate;
+    }
+
+    @ModuleOption("fixed delay in SECONDS to poll the remote directory")
```
fmarchand: I think fixed-rate for polling consumer is in milliseconds by default. isn't it ?

modules/source/ftp/config/ftp.xml
```diff
@@ -0,0 +1,53 @@
...
+        	<beans:property name="password" value="${password:#{null}}"/>
+        	<beans:property name="clientMode" value="${clientMode}"/>
+	</beans:bean>
+
+	<int-ftp:inbound-channel-adapter auto-startup="false" channel="files" session-factory="ftpSessionFactory"
+		filename-pattern="${filenamePattern}" remote-directory="${remoteDir}" local-directory="${localDir}"
+		remote-file-separator="${remoteFileSeparator}" preserve-timestamp="${preserveTimestamp}"
+		auto-create-local-directory="${autoCreateLocalDir}"
+		temporary-file-suffix="${tmpFileSuffix}" delete-remote-files="${deleteRemoteFiles}">
+		<poller fixed-rate="${fixedRate}" />
```
garyrussell: New pollers should use `PeriodicTriggerMixin` - see https://jira.spring.io/browse/XD-2846

spring-xd-dirt/src/main/java/org/springframework/xd/dirt/modules/metadata/FtpSourceOptionsMetadata.java
garyrussell: True, but see my other comment about `PeriodicTriggerMixin`.

ericbottard:
Rebased, added PeriodicTriggerMixin

modules/source/ftp/config/ftp.xml
```diff
@@ -0,0 +1,59 @@
...
+
+	<channel id="output" />
+
+	<channel id="files" />
+
+	<beans:beans profile="use-ref">
+		<bridge input-channel="files" output-channel="output"/>
+	</beans:beans>
+
+	<beans:beans profile="use-contents">
```
garyrussell: Hmmm - why did we add these profiles? 

I am not saying they are bad but it seems like gold plating and we should really have the `sftp` and `ftp` sources similarly configured - I think that means we should add these options to SFTP (or at least open a story for that).

spring-xd-shell/src/test/java/org/springframework/xd/shell/command/AbstractStreamIntegrationTest.java
```diff
@@ -28,17 +28,7 @@
...
-import org.springframework.xd.test.fixtures.FileSource;
-import org.springframework.xd.test.fixtures.JdbcSink;
-import org.springframework.xd.test.fixtures.JdbcSource;
-import org.springframework.xd.test.fixtures.MailSink;
-import org.springframework.xd.test.fixtures.NonPollingImapSource;
-import org.springframework.xd.test.fixtures.PollingMailSource;
-import org.springframework.xd.test.fixtures.TailSource;
-import org.springframework.xd.test.fixtures.TcpSink;
-import org.springframework.xd.test.fixtures.TcpSource;
+import org.springframework.xd.test.fixtures.*;
```
garyrussell: Have we switched back to wildcard imports?

ericbottard: Nope, but I switched laptops :) Will fix

Warning, one of your favorite discussions ahead: Why aren't the formatting options into VCS again?

modules/source/ftp/config/ftp.xml
ericbottard: Are your arguing about the use of profiles, or the ref/contents feature itself? While testing this, it quickly became obvious to me that having just the ref is hard to deal with, so I added this (which is on par with the `file` source).
So if you're discussing the latter, then yes I agree every source should behave the same

garyrussell: Yes, it's the latter - I realize it came from the `file` source, but it's not consistent with the implementation there (although I prefer this over the router there - and it's __completely missing__ from sftp :smile: )

ericbottard:
Created https://jira.spring.io/browse/XD-3013 to capture the `--ref=true|false` option for stfp

ericbottard:
Rebased, fixed imports.

garyrussell:
Merged as ed27e32103205c149b420b593ad06309cfd41071 with minor polish spaces->tabs etc.

PR #1558, opened 2015-04-17
[Sprint: 47] XD-2955 MessageBus performance improvements

dturanski:

PR #1560, opened 2015-04-20
XD-2911 - Improved performance of Tuple Creation

mminella:
**NOTE -** This is a breaking change.

This commit refactors how the conversion service used within a Tuple is
created to a more performant option.  Prior to this commit, for each
instance of TupleBuilder created (and under typical usage, each Tuple)
an instance of DefaultTupleConversionService was created.  This updated
version changes the default to be a static instance and changes the
recommended approach for customization to be a
ConfigurableConversionService be configured as a singleton within your
module's context and then used across all Tuple instances.  This
provides better performance in the creation of Tuples.


dturanski:
I will merge as is with the performance tests but I would like to move them into spring-xd-benchmark as soon as https://github.com/spring-projects/spring-xd/pull/1558 which introduces that project is merged. This is intended to provide similar tests and separate them from the normal build.

dturanski:
rebased and merged

PR #1562, opened 2015-04-20
XD-2872 Able to bypass authorization checks by appending ".json" or ".xml"

ghillert:


ericbottard:
Not sure we want to do this as it may catch false positives.
What about having two rules each time, eg
/streams
/streams.*

?

ghillert:
@ericbottard Rebased and changed the authorizations. This raises the question - Some controllers don't support returning XML responses which makes the API a bit erratic. Shall we just drop XML renderings altogether? @markpollack @sabbyanandan

ericbottard:
I'd rather think that we'll fix those couple ressource renderings one day.
Moreover, there is no problem denying access to a resource that does not
exist

spring-xd-dirt/src/main/resources/application.yml
```diff
@@ -215,35 +218,44 @@ xd:
         - GET    /jobs/executions/*/steps/*      => hasRole('ROLE_VIEW')
         - GET    /jobs/executions/*/steps/*/progress => hasRole('ROLE_VIEW')
         # Batch Job Instances
-        - GET    /jobs/instances?jobname=*       => hasRole('ROLE_VIEW')
+        - GET    /jobs/instances*                => hasRole('ROLE_VIEW')
```
ericbottard: Don't we want to explicitly catch
```
/jobs/instances?jobname=*
/jobs/instances.*?jobname=*
```
?

spring-xd-dirt/src/main/resources/application.yml
```diff
@@ -179,8 +179,11 @@ xd:
         # Streams
         - GET    /streams/definitions            => hasRole('ROLE_VIEW')
         - DELETE /streams/definitions            => hasRole('ROLE_CREATE')
+        - GET    /streams/definitions.*          => hasRole('ROLE_VIEW')
```
ericbottard: Would be better to have the dot version just below the non-dot, can fix on merge

spring-xd-dirt/src/test/java/org/springframework/xd/dirt/security/SingleNodeApplicationWithUsersFileTest.java
```diff
@@ -88,7 +90,716 @@ public void testWrongPasswordFails() throws Exception {
...
+								basicAuthorizationHeader("cartman", "cartmanpwd")))
+				.andDo(print())
+				.andExpect(status().isForbidden());
+	}
+
+	@Test
+	public void testAuthenticatedButUnauthorizedAccessToDefinitionsEndpointWithXmlExtension() throws Exception {
+		springXdResource.getMockMvc()
+				.perform(
+						get("/streams/definitions.xml").header("Authorization",
```
ericbottard: We should use JUnit @Parameterized to cover all 3 cases automatically, or this may quickly go beyond madness.

ghillert:
@ericbottard Rebased and converted to `@RunWith(Parameterized.class)`. Size of the test class was reduced to 225 lines of code. Thanks for the pointer, hadn't used parameterized Junit tests before.

ericbottard:
My pleasure :)

I wasn't expecting that approach though, was more thinking about using it for extensions only. Given that for each extension (including extensionless) we expect the same output, we should at least be able to divide the number of explicitly written data lines by 3.


Also, we could use the name() parameter of the `@Parameters` annotation to give each test a meaningful name (derived from the tested URL, etc.)

spring-xd-dirt/src/main/resources/application.yml
ghillert: Looks like query strings are stripped when using ant paths in Spring Security. See: http://docs.spring.io/spring-security/site/docs/current/reference/htmlsingle/#request-matching

You can switch to regular expression matching for more complicated matches but I don't know if that level of sophistication is really needed here.

ghillert:
@ericbottard Just pushed another commit - so that I ensure proper naming of tests. Hope this is finally getting mergeable.

ericbottard:
Merged on master (w/o squashing, as I am an idiot) as 9815144e440c9d4721e672ff2623460fe4926b7c and later commits

PR #1565, opened 2015-04-22
[Sprint: 47] XD-2572 Fixing configuration handling for SqoopTasklet

trisberg:
- adding hadoop.properties settings to the configuration used by SqoopRunner

  - adding config properties needed for kerberos secured cluster to SqoopRunner

  - adding hadoop-mapreduce-client-app.jar to build for each Hadoop distro (needed for secured cluster jobhistory)

  - removing default connection properties when --options-file is used


extensions/spring-xd-extension-sqoop/src/main/java/org/springframework/xd/sqoop/SqoopTasklet.java
```diff
@@ -74,6 +86,9 @@ protected boolean isStoppable() {
 				}
 			}
 		}
+		for (Map.Entry e : hadoopProperties.entrySet()) {
+			command.add(SPRING_HADOOP_CONFIG_PREFIX + "." + e.getKey() + "=" + e.getValue());
```
jvalkeal: You need to make this clear in docs that below doesn't work(like it works with shell and sinks):
```
cat ../config/hadoop.properties 
# Use servers.yml to change URI for namenode
# You can add additional properties in this file
spring.hadoop.security.authMethod=kerberos
spring.hadoop.security.namenodePrincipal=hdfs/neo@LOCALDOMAIN
spring.hadoop.security.rmManagerPrincipal=yarn/neo@LOCALDOMAIN
spring.hadoop.security.userKeytab=/usr/local/hadoops/jvalkealahti.keytab
spring.hadoop.security.userPrincipal=jvalkealahti/neo
mapreduce.jobhistory.principal=mapred/neo@LOCALDOMAIN
mapreduce.framework.name=yarn
```
but yml is fine:
```
# Hadoop properties
spring:
  hadoop:
    fsUri: hdfs://localhost:8020
    resourceManagerHost: localhost
    resourceManagerPort: 8032
    security:
      authMethod: kerberos
      namenodePrincipal: hdfs/neo@LOCALDOMAIN
      rmManagerPrincipal: yarn/neo@LOCALDOMAIN
      userKeytab: /usr/local/hadoops/jvalkealahti.keytab
      userPrincipal: jvalkealahti/neo
    config:
      mapreduce.jobhistory.principal: mapred/neo@LOCALDOMAIN
```
in a perfect world same hadoop.properties format should also work.

jvalkeal:
Other than that played using listings and imports with kerberized and non-kerberized hadoop and it was all right. Merging...

jvalkeal:
Merged, closing.

extensions/spring-xd-extension-sqoop/src/main/java/org/springframework/xd/sqoop/SqoopTasklet.java
trisberg: That should be pretty easy to fix. However, it is a bit confusing to the end user. Maybe we should switch the properties in the docs to be the corresponding Hadoop properties instead of the Spring ones. It is a hadoop.properties file and not a spring-hadoop.properties file. I was expecting to put Hadoop style properties in this file and not Spring style ones.

PR #1566, opened 2015-04-22
xd-2972: ensure eclipse metadata files are correct java level

aclement:
Ensures the source/target are set to 1.7 for created metadata. Without this the eclipse gradle plugin will set them to the level of the JDK you are running the gradle command on...

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -203,6 +203,10 @@ configure(javaProjects) { subproject ->
 		}
 
 		eclipse {
+			jdt {
+				sourceCompatibility=1.7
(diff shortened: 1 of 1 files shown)


ilayaperumalg:
LGTM, verified the eclipse metadata file `.settings/org.eclipse.jdt.core.prefs` under the sub projects have the correct java level and could import the projects into STS. Rebased and merged as  273086177f70e8569dddb3fd712d73a3c970298e. Also pushed to 1.1.x 8f45d55e99ea5e9def942bab5c0e88924f3728f7.

PR #1569, opened 2015-04-23
[XD-2920] Dynamic router should allow to discard messages

kdowbecki:
Router now allows to return null to discard the message.

Current tests would pass on both old and new `router` module implementation which is bad. The only difference can be observed in the logs as the `MessageDeliveryException` is not thrown. 

Is there a way to register a `MessageHandler` on router step `errorChannel` to intercept the `MessageDeliveryException` from `AbstractMessageRouter` so I can add some more test assertions? Is there a global Spring XD `errorChannel` if all steps within a stream are running in separate contexts?

Calling `getErrorChannel().subscribe(MessageHandler)` in my test is not working. Most likely because test is running synchronously using local transport and the `errorChannel` property is not set on `http` source which is the stream entry point.

### DIFF
--- a/modules/sink/router/config/router.xml
+++ b/modules/sink/router/config/router.xml
@@ -1,4 +1,4 @@
-<?xml version="1.0" encoding="UTF-8"?>
+ <?xml version="1.0" encoding="UTF-8"?>
 <beans:beans xmlns="http://www.springframework.org/schema/integration"
 	xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
 	xmlns:beans="http://www.springframework.org/schema/beans"
@@ -15,12 +15,13 @@
 
 
 	<beans:beans profile="use-expression">
-		<router input-channel="input" expression="${expression}"/>
+		<router input-channel="input" expression="${expression}" resolution-required="false"
+						default-output-channel="nullChannel"/>
 	</beans:beans>
 
 	<beans:beans profile="use-script">
 		<beans:import resource="../../../common/script-variable-generator.xml"/>
-		<router input-channel="input">
+		<router input-channel="input" resolution-required="false" default-output-channel="nullChannel">
 			<int-groovy:script location="${script}" script-variable-generator="variableGenerator" refresh-check-delay="60"/>
 		</router>
 	</beans:beans>
--- a/modules/sink/router/config/router.xml
+++ b/modules/sink/router/config/router.xml
@@ -1,4 +1,4 @@
- <?xml version="1.0" encoding="UTF-8"?>
+<?xml version="1.0" encoding="UTF-8"?>
 <beans:beans xmlns="http://www.springframework.org/schema/integration"
 	xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
 	xmlns:beans="http://www.springframework.org/schema/beans"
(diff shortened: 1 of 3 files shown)


ilayaperumalg:
> Calling getErrorChannel().subscribe(MessageHandler) in my test is not working. Most likely because test is running synchronously using local transport and the errorChannel property is not set on http source which is the stream entry point.

@kdowbecki  yeah, since this is synchronous with the local transport the `NettyHttpInboundChannelAdapter` that the `http` source uses throws the `internal_server_error` upon `MessagingException`. 

To have a global errorChannel sounds like the feature we plan to address here: https://jira.spring.io/browse/XD-2135

ilayaperumalg:
> Current tests would pass on both old and new router module implementation which is bad. The only difference can be observed in the logs as the MessageDeliveryException is not thrown.

I think we can use a different approach to test this one. Rather than using the `http` source, we could use a different one. Something like this:
```
public class RouterSinkTests extends AbstractStreamIntegrationTest {

	@Test
	public void testUsingExpressionThatThrowsMessagingException() throws Exception {
		FileSink fileSink = newFileSink().binary(true);
		String queue1 = generateQueueName();

		stream().create(generateStreamName(), "%s > transform --expression=payload+'-foo' | %s", queue1, fileSink);
		stream().create(generateStreamName(),
				"time | router --expression=payload.contains('a')?'" + queue1 + "':'" + "null" + "'");
		final CountDownLatch latch = new CountDownLatch(1);
		final List<Message<?>> messageList = new ArrayList<Message<?>>();
		MessageHandler handler = new MessageHandler() {

			@Override
			public void handleMessage(Message<?> message) throws MessagingException {
				messageList.add(message);
				latch.countDown();
			}
		};
		getErrorChannel().subscribe(handler);
		latch.await(5, TimeUnit.SECONDS);
		assertTrue(!messageList.isEmpty());
		Message<?> error = messageList.get(0);
		assertNotNull("expected an error message", error);
		Object payload = error.getPayload();
		System.out.println(payload);
		assertTrue("payload should be a MessagingException", payload instanceof MessagingException);
		getErrorChannel().unsubscribe(handler);
	}
```

modules/sink/router/config/router.xml
```diff
@@ -15,12 +15,13 @@
...
 	<beans:beans profile="use-expression">
-		<router input-channel="input" expression="${expression}"/>
+		<router input-channel="input" expression="${expression}" resolution-required="false"
+						default-output-channel="nullChannel"/>
 	</beans:beans>
 
 	<beans:beans profile="use-script">
 		<beans:import resource="../../../common/script-variable-generator.xml"/>
-		<router input-channel="input">
+		<router input-channel="input" resolution-required="false" default-output-channel="nullChannel">
```
ilayaperumalg: I think it would be nice to have a configuration option to specify the default output channel based on whether the messages should be discarded via `nullChannel` or put into `errorChannel`. Once we have a stream specific errorChannel (the feature specified in https://jira.spring.io/browse/XD-2135) then, it would all make sense.

ilayaperumalg: Created https://jira.spring.io/browse/XD-3068 to allow configuration option to specify default output channel.

ilayaperumalg:
@kdowbecki Thanks for the contribution.
LGTM. Squashed, rebased and merged as 240ba5546eca3f247a17f5fb74b571dc4007cbe9.

modules/sink/router/config/router.xml
garyrussell: `resolution-required` should be configurable too.

garyrussell: @ilayaperumalg 3068 doesn't need 2135 - setting the discard channel to, say `queue:discards` will send the discards to that named channel.

ilayaperumalg: @garyrussell sure, that's a nice idea.

PR #1572, opened 2015-04-24
[Sprint 47] XD-2984 Move hadoopDistro option to CommonOptions

ilayaperumalg:
- Since both the `admin` and `container` use `--hadoopDistro` option
we can have this option moved to `CommonOptions`
 - Fix XD config logger to log `hadooDistro` option for all the runtime JVMs.

### DIFF
--- a/scripts/xd/xd-admin
+++ b/scripts/xd/xd-admin
@@ -65,44 +65,29 @@ cd "`dirname \"$PRG\"`/.." >&-
 APP_HOME="`pwd -P`"
 cd "$SAVED" >&-
 
-APP_HOME_LIB=$APP_HOME/lib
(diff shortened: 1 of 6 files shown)


trisberg:
Shouldn't the handling of the --hadoopDistro option in the xd-admin and xd-container be the same? I would think that what is in xd-container is the best code to also use on xd-admin script rather than enumerate the hadoop versions supported in the script as well. This is already handled by the HadoopDistroOptionHandler.

ilayaperumalg:
@trisberg I see your point. I will modify the xd-admin script and update the PR.

ilayaperumalg:
@trisberg Updated the PR after fixing the admin scripts to use the OptionsHandler to validate the hadoop distro option.

trisberg:
LGTM, merging ...

PR #1576, opened 2015-04-27
XD-2995 Fix group name assignment

ilayaperumalg:
- For JMX MBean naming strategy, the mbean-exporter was using `xd.stream.name`
to uniquely identify each module. This broke the job module naming.
   - This fix uses a key `xd.group.name` to assign both the job/stream name
for the module being exported. That way we would avoid having Stream/Job specific
key to identify the module.


ilayaperumalg:
Also, cherry-picked and pushed to 1.1.x:
c6c2fa969793b857096dff65e626514a3711edab

PR #1577, opened 2015-04-27
[Sprint: 49] XD-2424, XD-2956, XD-2998, XD-3000 tuple functional and performance enhancements

dturanski:
This PR includes the following:

* moved codecs and tests out of dirt and into spring-xd-codecs.  TupleCodec is in spring-xd-tuple.
* Removed tuple id and timestamp creation by default. Use optional methods TupleBuilder.tuple.addId(), and addTimestamp()
* Avoids unnecessary creation of expensive objects such as ConversionService and DateFormat in serialization
* Additional serialization performance enhancements
* Added TupleCodecBenchmarkTests


ericbottard:
Do you have numbers regarding the perf enhancement (we always love numbers :) )

dturanski:
Yes and no. TupleCodecBenchmarkTests in spring-xd-benchmark measures
current performance. But not easy to reproduce the previous numbers for
comparison. Anecdotally it is an order of magnitude faster.  ./gradlew
spring-xd-benchmark:test -Drun_benchmark_tests=true.

dturanski:
@ericbottard - Most of the serialization costs were from restoreConversionService() in previous versions of TupleCodec.deserialize(). Additionally, as mentioned in XD-2424, TupleBuilder included a DateFormat field which has been removed. The cost of creating these classes is measured in TupleCreationBenchmarkTests.  The codec enhancements avoid this unless needed. Using a non-default ConversionService or DateFormat will perform as before.  Also, note that the cost of TupleCodec is O(n) relative to number of fields (assuming primitives) as shown [here](https://docs.google.com/spreadsheets/d/1HSl4N8Ze0GdHXcluhwU-23Y6P3I8Kub456yj8KNDIkI/edit?usp=sharing). Including nested Tuples, or POJO field values will decrease performance as well. TupleCodecBenchmarkTests includes numbers for both cases.

ericbottard:
@dturanski Got it. An order of magnitude faster is pretty cool. And I suspect most usecases deal with the default conversion service, so that's pretty good too!

markpollack:
I wonder if "TupleBuilder.setFormats() and TupleBuilder.setConversionService() settings are preserved when tuples are serialized." is needed....

dturanski:
@markpollack - explain please.

markpollack:
Yea sorry.  I mean the format is more about presentation than actual data values, so orthogonal to serialization.  The conversion service is maybe even somewhat more problematic as the user could create a tuplebuilder with all sorts of custom converters which may/maynot be able to be reconstituted on the receiving end.

dturanski:
Format, etc.  in theory is orthogonal but currently it is required since we need to invoke TupleBuilder to rehydrate a tuple. This info is state held by TB to create the Tuple so I don't see how you get an identical tuple back without addressing this.

ericbottard:
I'm not too keen on the first item:
> moved codecs and tests out of dirt and into spring-xd-spi-common. TupleCodec is in spring-xd-tuple. 

Can you elaborate why this was done?

dturanski:
Originally, I introduced a circular dependency that required moving the codecs. This dependency is no longer there and the repackaging is not necessary, but I included it here because it's an improvement IMHO. Codecs are coupled to the MB and not a concern of DIRT. In general, streamlining DIRT is a good thing.

ericbottard:
Got it. But while I agree to the latter, I don't think they belong in the
"spi" project. Implementors of busses should not need compile time
visibility of codecs.
So the question is, are we courageous enough to introduce a new
intermediate project?

dturanski:
I'll move it to a new project

spring-xd-codec/src/main/java/org/springframework/xd/dirt/integration/bus/serializer/CompositeCodec.java
```diff
@@ -24,6 +24,7 @@
 
 import org.springframework.integration.util.ClassUtils;
 import org.springframework.util.Assert;
+import org.springframework.xd.dirt.integration.bus.serializer.MultiTypeCodec;
```
markpollack: not used, will remove on merge

markpollack:
Fixed a small test error, added some javadoc and removed a few TODO comments in TupleBuilder,DefaultTuple. Merged as 5e6b7bc61c727db5baede2265f2534349a134265

PR #1578, opened 2015-04-28
XD-3004: Metadata resolution perf enhancement

ericbottard:
This is an obvious change, not re-creating an environment over and over again.
On my laptop, calling the REST endoint 100x times, went from ~650ms per unit
to approx. 200ms.

Other improvements are certainly possible, using caching but may be more involved.


ilayaperumalg:
LGTM, rebased and merged into master 2f9031a9fc6d6eaa6f51fe494b151061ddb4faeb & 1.1.x b20f0f54305ce155414a45bfbf9d781decce6706

PR #1583, opened 2015-04-29
XD-2953: Move code from SparkStreamingDriverModule to plugin

ericbottard:
This is the first step in an ongoing refactoring to remove Spark (and hadoop) from 
the main classpath, as discussed with Mark & Ilaya in Philly.

The next step will get rid of the custom module class entirely (and shouldBind())
The aim being to allow plugins to be autodetected, including from classes that
are only visible to the module classloader.


ilayaperumalg:
LGTM

ilayaperumalg:
Squashed, rebased and merged into master 95f71f78e379e6779297b6a65dfbca4c842fd83c

PR #1586, opened 2015-04-30
[Sprint: 49] XD-2993 Flo Specific parse endpoint

ericbottard:


spring-xd-dirt/src/main/java/org/springframework/xd/dirt/stream/XDParser.java
```diff
@@ -43,4 +45,29 @@
 	 *         described by the DSL.
 	 */
 	List<ModuleDescriptor> parse(String name, String config, ParsingContext type);
+
+	public static class DefinitionOrException {
```
markfisher: This class should not exist. The FloController should simply handle an Exception being thrown from the repository.

ericbottard: The problem is we want exceptions **per line** (_i.e._ in a multiline document such as
```
1: foo = http | log
2: bar = tap:stream:foo2 > log
```
we want to report an error at line 2 but still have a parse result for line 1)

markfisher: It seem like the loop should move up to the FloController then. If that's the only place it's used, we don't want to add complexity at this level.

ericbottard: I would disagree, as the role of the controller should be merely to interface between the HTTP world and business layer. If we were to switch transports, nothing should have to be rewritten.

Moving the loop upwards would require to move the transient repository thing there as well, clearly leaking business logic in the controller IMO.

I only see two ways to convey a mixed result: either via the `return` keyword, as the current state of this PR does, or via the `throw` keyword (we could have e.g. a PartialParsingFailureException, but then there would be code duplication with the 100% happy case)

markfisher: What does this have to do with transport changes? Perhaps the transient repository should not exist either (isn't it just a map of what's already been retrieved to avoid hitting the real repository again?). The reason I'm concerned is that while you are making these changes, we are also trying to simplify the parser where far too much responsibility has accumulated. At the moment, the changes you are pushing down to the parser appear to be only relevant for the use-case that hits FloController, so it feels like a responsibility leak in the other direction.

In any case, the classes with OR in the name are a red flag to me. Either in the parser or the controller, if such classes remain, I'd prefer to see something like ParseResult with hasErrors(), getErrors(), etc.

ericbottard: When I said "transport changes", I meant something other than http for the API (theoretical consideration, obviously). My rule of thumb being that we should be able _e.g._ to accept commands over let's say raw telnet and the underlying business logic should not change. Sorry for the use of (yet another) overloaded term.

The transient repository is here not just to avoid hitting the real repository, but also to span the whole method call. In the example I gave above, it stores the first successful stream "creation" (although we don't save it for real) so that _e.g._ a `tap:stream:foo` would succeed. This is also why it's not just about moving the loop to the controller (or the javascript client, for that matter).
I couldn't agree more with you about the parser, but this extension merely acknowledges the fact that we now want to support the 
```
name = DSL
name = DSL
name = DSL
```
in addition to the simple
```
DSL
```
construct. 

About the class names, those are definitely non final and the aim of this PR is exactly what we're doing now, _i.e._ discuss the approach.

markfisher: Justifying architecture decisions based on "theoretical considerations" that are 99% likely NOT to ever happen is the #1 source of technical debt. The parser should parse a single stream definition IMO. Anything else needs to be a layer above that. If only used by FloController, we should not be adding yet more complexity to the parser itself. I would even prefer to see a decorated version of the parser that is used by FloController, and for now that could even be an inner class of FloController.

Thanks for the "tap:stream:foo" example since that clarifies the rationale for more than a cache/Map. However, that also aligns with a concern I have as we're trying to refactor the parser. I think we should have a separation between syntactic parsing and validation based on things like existing streams.

ericbottard: I won't try to argue with you, especially in written form, I know where this leads :)

I like the decorated approach.

About the separation of syntactic / semantic split, please note that we already have two levels of parsers. BUT, the inner one had to know too much (mostly because of our previous approach to composed modules) and I would argue that the inner layer should not know about either the repository of existing streams (it does currently) nor things like module options (it does not atm). Those should be responsibilities of the outer layer.

For the record, I also think that the limitation we have where a stream can't be named after one of its module is now void and could be lifted.

markfisher: Agree 100% with your thoughts on where the "inner" parser should go.

It sounds like the best compromise for now would be to take the decoration path for the multi-stream "dry run" parser, and while at it maybe change the return value from a List of instances of classes that include "OR" in the name to something more like a MultiStreamParseResult or whatever.

aclement:
> I think we should have a separation between syntactic parsing and validation based on things like existing streams.

Completely agree on that.

If you guys want to push back on the multi stream parse, I'd be ok with having a discussion on that too. I feel like it can be helpful to craft your taps alongside your mainstream or if you have a set of streams connected by named channels, to be able to work on them as a group. But if you don't agree, we should be discussing it.

ericbottard:
Pushed some changes.

I think we should discuss what should get serialized in the case of success (tying the REST API to ModuleDescriptor is a big no-no IMO)

aclement:
StreamNode/ModuleNode appears to be what I used to build the structure I need in the current 'patched' XD.

spring-xd-dirt/src/main/java/org/springframework/xd/dirt/stream/XDStreamParser.java
```diff
@@ -322,4 +333,98 @@ private ModuleType resolveModuleType(String moduleName, ModuleType... candidates
...
+	/**
+	 * A wrapper around the canonical parser that supports multiline "document" parsing.
+	 *
+	 * <p>While XDStreamParser parses a single definition, this class is able to parse multiple, reporting
+	 * errors or accepting constructs like taps against definitions that appear earlier in the document (without
+	 * actually saving them in the main repository).</p>
+	 *
+	 * @author Eric Bottard
+	 */
+	public static class FloParser {
```
aclement: I'd rather nothing we put into XD was called 'Flo' specific. I mean the content assist isn't FloContentAssist is it - it is a generally usable endpoint - shell and Flo both use it.

spring-xd-dirt/src/main/java/org/springframework/xd/dirt/rest/FloController.java
```diff
@@ -0,0 +1,83 @@
...
+
+/**
+ * A controller that can accept a whole list of definitions and report whether
+ * they are valid as a whole or not.
+ *
+ * @author Eric Bottard
+ */
+@RestController
+@RequestMapping("/flo")
+public class FloController {
```
aclement: Similar comment to below, I don't really want flo specifics in XD. I want generally reusable endpoints that others may also want to use. I think a general parse endpoint makes a lot of sense. What seems to make things trickier here is that the syntax processing is mixed up a bit with the semantic processing in the parser. So if I want to parse two related streams, I have to use the TransientDefinitionRepository. If the syntax parsing was entirely separate from any semantic concerns (as I know we'd all like to do), I could simply call a parse endpoint that took a single DSL string. And it wouldn't matter if it referred to another stream that didn't exist yet. But because of those checks in the parser I have to pass in multiple strings so the verification checks can pass.  Is it worth trying to do that work now I wonder: pull out the parser semantic comprehension (may not be a lot of it), then you can just give me a single DSL string parsing endpoint.

ericbottard:
The "Flo" part in naming has always been a placeholder, until we find a better name (suggestions welcome).

Going back to the more architectural discussion, I'd like to restate what I understand from what we're all saying, just to make sure (I think I have issues with the very last bit):
- ideally, we'd like to have 2 layers in the parser:
  1. the inner layer only knows about AST parsing; This is ~StreamConfigParser without stream lookup functionality (taps)
  2. the outer layer would have semantic knowledge, including the tap verification, module option verification, etc. This is ~XDStreamParser. Ideally, I think this should return StreamDefinition (if this contained List\<ModuleDescriptor\>, as I think it should)
- seems like you're saying you want the result of the __inner__ parser (which could then be used to parse single units)

The problem I see with this is logic duplication on the client side

aclement:
I was thinking perhaps an endpoint around '/tools' or some such.  /tools/parse now,  /tools/whatever added in the future.

Ideally I would love all the semantic checking done server side and a nice list of warnings/errors alongside my parse results. I was only suggesting I'd take the result of the syntax parse (inner parser as you call it) because right now I have nothing at all and I'll take whatever I can get. If we can't agree on semantic checking stuff right now, I'd take single stream parsing (if we also turned off tap verification at the point where it is currently done in the parser).

If I have to temporarily duplicate the semantic checking client side, it is not ideal, but so be it. Just an observation that I also can see different kinds of semantic checking - some can be done on single streams (like module option checking) whilst some involves multiple streams (tap verification checking).

If you can get agreement on semantic checking with a multiple streams parsing endpoint, awesome, I'll take it :)

ericbottard:
At this point, I think it'd be easier to give you semantic parsing (then we can change our internals without impacting you). What I have a hard time understanding though is what you need in return. Do you really need the level of detail that the syntactic parsing is currently giving (e.g. token start / end, type, etc.)?

What about what is returned by the current version of this PR (in terms of content, not form):
`List<ModuleDescriptor>` gives module name, label, options etc in case of success
`VndError` gives a message that contains the first error _per stream_ (which currently may contain a caret pointer into the raw DSL string pointing to the error)

Is that ok?

aclement:
positions would be nice but are not necessary for my use case - *if* the warning/error messages associated with any problems include correct positional information. First error per stream is probably fine for now, but I will need that pointer to the start of the error (I can guess the end position if you aren't giving it to me). If the first error is a syntax error, that is fine. But if there are no syntax errors but a number of semantic errors, you could give me all of those rather than only the first one. But given the deadline, I'll settle for first error of whatever kind for now.

I thought I wasn't getting `ModuleDescriptor`s because above you mentioned "tying the REST API to ModuleDescriptor is a big no-no IMO". Those *should* be sufficient but honestly until i rewrite all my parser-to-graph processing stuff based on the new endpoint, I won't be 100% sure.

ericbottard:
Currently, we only expose the first error. But I guess if we were to catch several validation errors, you would benefit from it too.
I think we're closing to an agreement here, with the only questions pending being about _how_ to serialize:

We use that `VndError` thoughout until now, which only contains a textual representation. You'd currently have the caret position by parsing the error message (which is two lines), which is not ideal:
```
tap:stream:foobar > log
           ^ unknown stream 'foobar'
```
We'd need to go away from simple `VndError` to have a more structured error representation, which I'm totally fine with.

When I said I did not want to tie the REST API to ModuleDescriptor, I meant I did not want to blindly let Jackson serialize that class to you (in case it changes). But the bits of it, I'd give you through a dedicated DTO for the REST layer.

Lastly, do you prefer to have this structure:
```
{
  "error": null,
  "success": {something}
} /
{
  "error": {something},
  "success": null
}
```

or this :

```
{
   "my_field_that_is_only_there_in_case_of_error": {something},
   "my_other_field_that_is_only_there_in_case_of_error": {something}
} /
{
   "my_field_that_is_only_there_in_case_of_success": {something},
   "my_other_field_that_is_only_there_in_case_of_success": {something}
} 
```

I guess the former is better, correct?

aclement:
I believe at the moment we are parsing error text, which I agree is nasty. Something a little more structured would be great.

On the structure choice, I'd be ok with the former.

ericbottard:
I updated the PR with what we agreed on.
I chose to have several potential errors for now (although currently there can only be one).

ericbottard:
```
curl http://localhost:9393/tools/parse\?definitions\=foo\=time\|l:%20log%20--name\=foo%0Abar\=%20tap:stream:foo2\>log | json_pp
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100   418    0   418    0     0  12935      0 --:--:-- --:--:-- --:--:-- 13062
{
   "lines" : [
      {
         "success" : [
            {
               "sinkChannelName" : null,
               "sourceChannelName" : null,
               "options" : {},
               "type" : "source",
               "label" : "time",
               "name" : "time"
            },
            {
               "name" : "log",
               "options" : {
                  "name" : "foo"
               },
               "label" : "l",
               "type" : "sink",
               "sourceChannelName" : null,
               "sinkChannelName" : null
            }
         ],
         "errors" : null
      },
      {
         "errors" : [
            {
               "position" : 16,
               "message" : "XD116E:(pos 16): unrecognized stream reference 'foo2'\nbar= tap:stream:foo2>log\n                ^\n"
            }
         ],
         "success" : null
      }
   ]
}
```

aclement:
Thanks Eric - I think that output looks good to me.

spring-xd-dirt/src/main/java/org/springframework/xd/dirt/stream/XDStreamParser.java
```diff
@@ -322,4 +333,98 @@ private ModuleType resolveModuleType(String moduleName, ModuleType... candidates
...
+		}
+
+		public DocumentParseResult parse(String[] document) {
+			DocumentParseResult result = new DocumentParseResult(document.length);
+			CrudRepository<BaseDefinition, String> transientRepository = new TransientDefinitionRepository();
+			StreamConfigParser parser = new StreamConfigParser(transientRepository);
+			for (String nameAndDefinitionPair : document) {
+				try {
+
+					StreamNode stream = parser.parse(nameAndDefinitionPair);
```
ilayaperumalg: Are we planning to add support for parsing Job definition in a separate PR?

ilayaperumalg: or, does the scope for this story only deal with stream definition only?

ericbottard: Even though the class is named StreamNode, it's not Stream specific. I tried to rename things to be stream agnostic I my latest commit

ilayaperumalg: hmm, I see the ParsingContext uses `stream`. I think we need to be using `job` to parse the Job definition.

ericbottard: That's correct, that's the last bit that ties this PR to streams if I'm not mistaken. I discussed with @aclement and he's only interested in streams at the moment.

ilayaperumalg: ok. I have created https://jira.spring.io/browse/XD-3059 to track the support to parse multi line content with both stream and job definitions.

sabbyanandan: Flo currently only supports create and parsing of Streams; hence, the scope of this PR is good enough. Hopefully, we will be addressing the batch workflow creation (from Flo) after the Flo BETA release. Thanks for the JIRA.

spring-xd-dirt/src/main/java/org/springframework/xd/dirt/stream/XDStreamParser.java
```diff
@@ -322,4 +333,98 @@ private ModuleType resolveModuleType(String moduleName, ModuleType... candidates
...
+			}
+
+			@Override
+			public Iterable<BaseDefinition> findAll() {
+				return Iterables.concat(delegate.repository.findAll(), store.values());
+			}
+
+			@Override
+			public long count() {
+				return store.size() + delegate.repository.count();
```
ilayaperumalg: This could contain duplicate count as the stream definitions being parsed are always stored in this store. In any case, I don't see any of these methods (except `save` and `findOne`) being used. It doesn't hurt.

ilayaperumalg:
LGTM, merging

ilayaperumalg:
Squashed, rebased and merged as bee088be4914dc4b28ab25367b0a4e6da0bc6284 after adding some commit description :-)

PR #1592, opened 2015-05-01
[Sprint: 49] XD-3024: Extra metric endpoints for getting all metric resources for counters, gauges and rich-gauges.

BoykoAlex:
Counters supports paging. Gauges and rich-gauges I couldn't get paging support implementation thus it's just gets all resources. (Couldn't figure out the right implementation for Redis based repository class)

See https://jira.spring.io/browse/XD-3024

### DIFF
--- a/spring-xd-dirt/src/main/java/org/springframework/xd/dirt/rest/metrics/AbstractMetricsController.java
+++ b/spring-xd-dirt/src/main/java/org/springframework/xd/dirt/rest/metrics/AbstractMetricsController.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2002-2013 the original author or authors.
+ * Copyright 2002-2015 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -25,6 +25,7 @@ import org.springframework.data.web.PagedResourcesAssembler;
 import org.springframework.data.web.PagedResourcesAssemblerArgumentResolver;
 import org.springframework.hateoas.PagedResources;
 import org.springframework.hateoas.ResourceAssembler;
+import org.springframework.hateoas.mvc.ResourceAssemblerSupport;
 import org.springframework.http.HttpStatus;
 import org.springframework.web.bind.annotation.PathVariable;
 import org.springframework.web.bind.annotation.RequestMapping;
@@ -38,7 +39,7 @@ import org.springframework.xd.rest.domain.metrics.MetricResource;
 /**
  * Base class for controllers that expose metrics related resources. Subclasses are meant to provide the root
  * {@link RequestMapping} uri.
- * 
+ *
  * @author Eric Bottard
  * @author Ilayaperumal Gopinathan
  */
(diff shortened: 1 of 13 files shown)


ericbottard:
I'm not sure I understand the purpose of this PR. Seems this functionality is already supported (_e.g._ http://localhost:9393/metrics/counters )

Maybe what you want is an endpoint that returns the _values_ of the metrics (what you'd get when querying a particular metric) in addition to their name?). We've been adding such functionality in other areas, and what we did is to add a special request parameter (`detailed=true`). The implementation trick then is to dynamically select the ResourceAssembler used.

Such an approach may not be practically feasible for some metrics though (esp. aggregate-counters).

The proposed solution has the problem that it uses `/all` as an endpoint, but this could be the name of a metric.

BoykoAlex:
Ok, I'll rework the code to use `detailed` parameter instead of `/all`

BoykoAlex:
Can you give me an example REST endpoint where you use the `detailed` parameter that i could use as a reference implementation?

BoykoAlex:
Eric, I've corrected the PR. I adjusted the endpoints to have optional `detailed` parameter. The parameter  would only work for the following endpoints
- `/metrics/counters`
- `/metrics/gauges`
- `/metrics/rich-gauges`

spring-xd-dirt/src/main/java/org/springframework/xd/dirt/rest/metrics/AbstractMetricsController.java
```diff
@@ -66,8 +67,28 @@ public AbstractMetricsController(R repository) {
...
+	 * @return
+	 */
+	protected PagedResources<? extends MetricResource> list(Pageable pageable,
+			PagedResourcesAssembler<M> pagedAssembler,
+			ResourceAssemblerSupport<M, ? extends MetricResource> resourceAssembler) {
+		/* Page */Iterable<M> metrics = repository.findAll(/* pageable */);
+
+		// Ok for now until we use PagingAndSortingRepo as we know we have lists
+		Page<M> page = new PageImpl<M>((List<M>) metrics);
+		return pagedAssembler.toResource(page,
```
ericbottard: I'd change this to always require a third argument and have the subclass dynamically pass the correct assembler (`shallowResourceAssembler` will need to become protected then)

ericbottard: To clarify, the aim would be to have only 1 version of the list method (3 args)

ericbottard:
This is starting to look good. I guess you don't need a pointer to an implementation anymore, but here is one: f79eab425018d04fe0f2f9c768744ddc62e7741b

BoykoAlex:
Thanks Eric!-) I've updated the PR as per your comment.

BoykoAlex:
Shall we merge this PR in? Or there is a few  more tidy ups to do Eric?

ericbottard:
I'll be working on it asap (working on other stuff atm).
I'd like to have a consistent experience for all metrics, for one thing.

BoykoAlex:
I have updated FVC controller to have support for `detailed` parameter. 
Not sure how to deal with agg-counters. I suppose if i get all of them and then iterate over them and get the data for the last hour (or minute) and then send that as a response won't be a good scalable solution, right? (N + 1 requests to the repository-DB)

ericbottard:
Yes, that is true, but we can't do any better at the time anyway. We can always document this shortcoming. By the way, we'd need to amend [the documentation](https://github.com/spring-projects/spring-xd/blob/master/src/docs/asciidoc/REST-API.asciidoc) to reflect the possibility of adding `detailed=true` as well as adding some tests ([CountersControllerIntegrationTests](https://github.com/spring-projects/spring-xd/blob/master/spring-xd-dirt/src/test/java/org/springframework/xd/dirt/rest/metrics/CountersControllerIntegrationTests.java) and friends)

BoykoAlex:
I've added naive support for `detailed` request parameter for aggregate-counters (hope it looks alright). Updates REST API docs for `detailed` parameter and added Unit tests to cover `detailed` parameter addition.
Eric, let me know if there is anything else that needs to be done.

ericbottard:
Alex,

I merged this as 2198067 by adding the other parameters (`to`, `from`, `resolution`) to aggregate-counters controller. Indeed, this is not very effective in practice, but at least it's consistent and there to use for small numbers of counters.

Thanks again for all the hard work on this PR.

BoykoAlex:
Thanks Eric!-) Much appreciated :-)

PR #1594, opened 2015-05-03
[Sprint: 48] Update to Reactor 2.0.1.RELEASE

smaldini:
Upgrade reactor-ip and reactor-syslog
Exclude Reactor 1.x (org.projectreactor)
add explicit dependency to netty


build.gradle
```diff
@@ -255,6 +255,7 @@ configure(javaProjects) { subproject ->
 		configurations.all {
 			resolutionStrategy.cacheChangingModulesFor 60, 'minutes'
 			exclude group: 'ch.qos.logback', module: 'logback-classic'
+			exclude group: 'org.projectreactor', module: 'reactor-core'
```
artembilan: Unfortunately we can do this.
`GatewayProxyFactoryBean#setReactorEnvironment(Environment reactorEnvironment)`
And SF tries to do some type extraction from similar public methods on bean definition phase.

To overcome that and really allow to have `Reactor` as `optional` we did this in SI-4.2:
```
setReactorEnvironment(Object reactorEnvironment)
```

smaldini: Do you konw if 4.2 is in target for Spring XD 1.2 ? @markpollack  ?

markpollack:
No, 4.2 is not the target for XD 1.2.  I suspect likely 1.3.  Yep, SI 4.1.x depends on reactor 1, which is why we had that whole discussion a while ago about changing package/artifact names so they can both live on the same classpath.  Our CI integration tests on EC2 can catch this issue.  I'll look to merge w/o excluding the reactor 1 jars.

extensions/spring-xd-extension-reactor/src/main/java/org/springframework/xd/integration/reactor/config/SyslogInboundChannelAdapterParser.java
```diff
@@ -21,28 +21,33 @@
...
 		BeanDefinitionBuilder builder = ReactorNamespaceUtils.createBeanDefinitionBuilder(
-				SyslogInboundChannelAdapter.class,
+				SyslogInboundChannelAdapterConfiguration.class,
 				element
 				);
 
 		IntegrationNamespaceUtils.setValueIfAttributeDefined(builder, element, "host", "host");
 		IntegrationNamespaceUtils.setValueIfAttributeDefined(builder, element, "port", "port");
 
-		IntegrationNamespaceUtils.setReferenceIfAttributeDefined(builder, element, "channel", "outputChannel");
```
markpollack: was this a bug?  the xsd didn't have outputChannel before.

artembilan: @markpollack , correct: `outputChannel` is a setter in the `SyslogInboundChannelAdapter`.
The XSD has `channel` attribute.
But I agree: I don't see `setOutput()` on the `SyslogInboundChannelAdapterConfiguration` hierarchy...

extensions/spring-xd-extension-reactor/src/main/java/org/springframework/xd/integration/reactor/net/ReactorPeerFactoryBean.java
```diff
@@ -0,0 +1,280 @@
...
+import java.net.URI;
+import java.util.HashMap;
+import java.util.Map;
+
+
+/**
+ * @author Stephane Maldini
+ */
+@SuppressWarnings({"unchecked", "rawtypes"})
+public class ReactorPeerFactoryBean<IN, OUT> implements FactoryBean<ReactorPeer<IN, OUT, ?>> {
```
markpollack: This is a pretty big new addition - what does this do at a high level, can you add javadoc?

modules/source/reactor-ip/config/reactor-ip.properties
```diff
@@ -1,2 +1,2 @@
 info.shortDescription = Reactor based source that allows a remote client to submit data over a raw TCP or UDP socket.
-options_class = org.springframework.xd.integration.reactor.net.NetServerSourceOptionsMetadata
+options_class = org.springframework.xd.integration.reactor.net.ReactorPeerSourceOptionsMetadata
```
markpollack: guess this should be ReactorNetSourceOptionsMetadata as I get the error 

```
Exception in thread "main" java.lang.IllegalStateException: Unable to load class used by ModuleOptionsMetadata: org.springframework.xd.integration.reactor.net.ReactorPeerSourceOptionsMetadata
```
when compiling (or it wasn't added)

extensions/spring-xd-extension-reactor/src/main/java/org/springframework/xd/integration/reactor/net/ReactorPeerFactoryBean.java
```diff
@@ -0,0 +1,280 @@
...
+		} else if (UDP.equals(transport)) {
+			return NetStreams.udpServer(NettyDatagramServer.class, new NetStreams.UdpServerFactory<IN, OUT>() {
+				@Override
+				public Spec.DatagramServerSpec<IN, OUT> apply(Spec.DatagramServerSpec<IN, OUT> spec) {
+					commonSpecProperties(spec);
+					//Add specifics to UDP
+					return spec;
+				}
+			});
+		} else if (HTTP.equals(transport)) {
```
markpollack: This is a udp/tcp module - why is http functionality added here?

extensions/spring-xd-extension-reactor/src/main/java/org/springframework/xd/integration/reactor/net/ReactorPeerFactoryBean.java
```diff
@@ -0,0 +1,280 @@
...
+		this(env, "tcp", codecs);
+	}
+
+	public ReactorPeerFactoryBean(Environment env, String transport, Map<String, Codec> codecs) {
+		this.codecs = (null == codecs ? DEFAULT_CODECS : codecs);
+		this.transport = transport;
+		this.environment = (null == env ? Environment.initializeIfEmpty().assignErrorJournal() : env);
+	}
+
+	public ReactorPeerFactoryBean<IN, OUT> configure(URI uri) {
```
markpollack: This method is never used

extensions/spring-xd-extension-reactor/src/main/java/org/springframework/xd/integration/reactor/net/ReactorPeerFactoryBean.java
```diff
@@ -0,0 +1,280 @@
...
+
+	private int port;
+
+	private String framing;
+
+	private int lengthFieldLength;
+
+	private Codec delegateCodec;
+
+	public ReactorPeerFactoryBean(Environment env) {
```
markpollack: The first 3 ctors aren't used... get rid of them?

extensions/spring-xd-extension-reactor/src/main/java/org/springframework/xd/integration/reactor/net/ReactorPeerFactoryBean.java
```diff
@@ -0,0 +1,280 @@
...
+	 * @param dispatcher dispatcher name
+	 * @return {@literal this}
+	 */
+	public ReactorPeerFactoryBean<IN, OUT> setDispatcher(String dispatcher) {
+		Assert.notNull(environment, "Cannot lookup dispatcher from unassigned environment");
+		this.dispatcher = environment.getDispatcher(dispatcher);
+		return this;
+	}
+
+	public ReactorPeerFactoryBean<IN, OUT> setDispatcher(Dispatcher dispatcher) {
```
markpollack: Why the fluent API for the setters?  I don't see it used anywhere.

smaldini: I just ported the current code without any assumption on the feature but trying to make it work in the tests provided. But yeah, agree a lot of this is not used (was it ?) but I did want to just port 1-1 what you had, there is no addition everything has just been renamed and adapted.

extensions/spring-xd-extension-reactor/src/main/java/org/springframework/xd/integration/reactor/net/ReactorPeerFactoryBean.java
smaldini: good question, although you can stream over http (which is itself based on tcp).

extensions/spring-xd-extension-reactor/src/main/java/org/springframework/xd/integration/reactor/net/ReactorPeerFactoryBean.java
smaldini: yes left it as it was here wasn't sure about what it was for.

extensions/spring-xd-extension-reactor/src/main/java/org/springframework/xd/integration/reactor/net/ReactorPeerFactoryBean.java
smaldini: was here, not sure why it was tho but left it.

extensions/spring-xd-extension-reactor/src/main/java/org/springframework/xd/integration/reactor/config/SyslogInboundChannelAdapterParser.java
smaldini: yeah I left a lot of things as it was but some stuff wasn't seem to work when I ported so that one for instance.

build.gradle
smaldini: @markpollack you can still then have a 1.1.x depedency only on core to satisfy the Environment object and exclude everything else (disruptor, sl4j, gs collection).

smaldini: Side note, is this not possible to have a 4.1.4 with optional dependency on reactor already and document that. It doesn't seem to be breaking any functionality but the arcanes of maintaining popular libraries is a difficulty to overcome you have the experience with :)

smaldini:
So this just tries to port simply what happens, obviously IMO it would just be better off a new couple modules with appropriate options and refreshed internal that matches the latest current standards from SXD.
Also it's about the value they provide you, so far unbounded consuming of IO request is good but not the end goal. Also we provide clients and server should be able to reply too (replyTo header?). When it comes to that, reactor provides way (mainly documented here https://github.com/reactive-ipc/reactive-ipc-jvm) to address flushing and close separately, they are not exposed neither right now. Also when using Stream in combination with the writing method we detect its "capacity" and adapt to a batch flush strategy.

All of that to say this should be the first step to work to provide immediate drop of the older dependency but we will work on a better solution.

extensions/spring-xd-extension-reactor/src/main/java/org/springframework/xd/integration/reactor/config/SyslogInboundChannelAdapterParser.java
markpollack: I think this needs to be put back to `outputChannel`... testing with a stream definition so we test the xsd.  the current tests don't make use of the xsd.

markpollack: The whole thing blows up even earlier since `ReactorNamespaceUtils.createBeanDefinitionBuilder` is trying to set the reactor environment on a ctor for `SyslogInboundChannelAdapterConfiguration` which doesn't exist.

smaldini: Right just put it back then, I'll a bit around this afternoon ping me if you want.

markpollack:
reworked and merged as f6c2bd737e0b564ce00fab3fe26c32c5de315a16

smaldini:
Awesome we'll progress from that over the next iterations.

PR #1601, opened 2015-05-07
[Sprint 49] XD-2992 Support multiple topics for kafka source

ilayaperumalg:
- Allow comma separated list of topics as module option to kafka source module
 - Restrict initialOffset and explicit partition list only for single topic
   - Add validation that checks these options are valid only if single topic is configured
 - Modify `KafkaPartitionAllocator`
   - Remove ZooKeeper dependency to persist the partitions for the kafka source
   - Evenly distribute the kafka partitions (all or specific) to the requested module instances
 - Update test

### DIFF
--- a/extensions/spring-xd-extension-kafka/src/main/java/org/springframework/integration/x/kafka/KafkaSourceModuleOptionsMetadata.java
+++ b/extensions/spring-xd-extension-kafka/src/main/java/org/springframework/integration/x/kafka/KafkaSourceModuleOptionsMetadata.java
@@ -36,7 +36,9 @@ import org.springframework.xd.module.options.spi.ProfileNamesProvider;
 @Mixin({KafkaZKOptionMixin.class, KafkaConsumerOptionsMixin.class, KafkaOffsetTopicOptionsMixin.class})
 public class KafkaSourceModuleOptionsMetadata implements ProfileNamesProvider {
 
-	private String topics = ModulePlaceholders.XD_STREAM_NAME;
+	private String topic = "";
+
+	private String topics = "";
 
 	private String partitions = "";
 
@@ -58,9 +60,18 @@ public class KafkaSourceModuleOptionsMetadata implements ProfileNamesProvider {
 
 	private int queueSize = 1000;
 
+	@ModuleOption("single topic name")
+	public void setTopic(String topic) {
+		this.topic = topic;
+	}
+
+	public String getTopic() {
+		return this.topic;
+	}
+
 	@ModuleOption("comma separated kafka topic names")
-	public void setTopics(String topic) {
-		this.topics = topic;
+	public void setTopics(String topics) {
+		this.topics = topics;
 	}
 
 	public String getTopics() {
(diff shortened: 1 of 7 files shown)


ilayaperumalg:
Created https://jira.spring.io/browse/XD-3035 to handle the deletion of offsets upon stream destroy.

spring-xd-shell/src/test/java/org/springframework/xd/shell/command/KafkaSourceSinkTests.java
```diff
@@ -81,16 +94,21 @@ public void deleteTopic() {
...
 		// create stream with kafka source
 		final CounterSink counter = metrics().newCounterSink();
-		stream().create(generateStreamName(), "kafka --topic='%s' --zkconnect=%s --outputType=text/plain | " +
+		stream().create(generateStreamName(), "kafka --topics='%s,%s' --zkconnect=%s --outputType=text/plain | " +
 				"filter --expression=payload.toString().contains('%s') | %s",
-				topicToUse, kafkaTestSupport.getZkConnectString(), stringToPost, counter );
-		httpSource.ensureReady().postData(stringToPost);
+				topic1ToUse, topic2ToUse, kafkaTestSupport.getZkConnectString(), stringToPost, counter );
+		httpSource1.ensureReady().postData(stringToPost);
+		httpSource1.ensureReady().postData(stringToPost);
```
mbogoevici: the test posts data to the same http source - should be `httpSource2` instead

extensions/spring-xd-extension-kafka/src/main/java/org/springframework/integration/x/kafka/KafkaSourceModuleOptionsMetadata.java
```diff
@@ -31,7 +36,7 @@
 @Mixin({KafkaZKOptionMixin.class, KafkaConsumerOptionsMixin.class, KafkaOffsetTopicOptionsMixin.class})
 public class KafkaSourceModuleOptionsMetadata implements ProfileNamesProvider {
 
-	private String topic = ModulePlaceholders.XD_STREAM_NAME;
+	private String topics = ModulePlaceholders.XD_STREAM_NAME;
```
mbogoevici: I'm wondering if we shouldn't support both options (`topic`) and (`topics`) and allow only one to be set - for backwards compatibility - e.g. for existing deployments, etc. Granted, this adds some complexity ...

spring-xd-shell/src/test/java/org/springframework/xd/shell/command/KafkaSourceSinkTests.java
mbogoevici: this is sort of minor, can be fixed on merge.

extensions/spring-xd-extension-kafka/src/main/java/org/springframework/integration/x/kafka/KafkaSourceModuleOptionsMetadata.java
ilayaperumalg: We can have both options `topic` (at least for backward compatibility) and `topics` with a validation condition saying only one can be set at a given time.

ilayaperumalg:
Pushed the changes that makes both `topic` and `topics` available as module options while keeping them  mutually exclusive.

mbogoevici:
Rebased, merged w/https://github.com/spring-projects/spring-xd/commit/2ffeb96367fb78981ac0ceedd64495c6e4c79ec5

PR #1604, opened 2015-05-10
[Sprint 49]XD-2691 Upgraded to gradle 2.2 and ported to maven-publish plugin

dturanski:
* Upgraded to gradle 2.2 (There is an issue with the boot gradle plugin and 2.3 - https://github.com/spring-projects/spring-boot/issues/2679)
* Ported maven pom publishing to use maven-publish plugin
* Aliased publishPomToLocalMavenRepo -> install for backward compatibility. This required removing 'propdeps-maven' plugin which doesn't appear to do anything useful for spring-xd anyway, and includes the replaced 'maven' plugin.
* Removed dependencyManagement from spring-xd-*.pom since it is not needed and basically duplicates every dependency
* Added dependencyManagement to spring-xd-module-parent.pom - since module projects can use this to ensure consistent versions
* Moved developer info to a global property in build.gradle to make it more visible and reusable
* Removed unnecessary setting $XD_HOME from mvn and gradle test configuration for custom modules
* Tested spring-xd-samples with new spring-xd-module-parent.pom and spring-xd-gradle plugin. The built module jars appear to be identical.

NOTE that the generated pom files require manual validation. I visually compared several poms and carefully inspected spring-xd-dirt and spring-xd-module-parent to verify dependency scope, exclusions, etc.

### DIFF
--- a/gradle/build-gradle-plugins.gradle
+++ b/gradle/build-gradle-plugins.gradle
@@ -23,9 +23,6 @@ propertiesFile.withInputStream {
 	versions.load(it)
 }
 
-//apply from: 'publish-maven.gradle'
-//task install(dependsOn:'publishToMavenLocal'){}
-
 project('spring-xd-module-plugin') {
 	startScripts.setEnabled(false)
     description = 'Gradle plugin to build a Spring XD Module'
@@ -37,8 +34,8 @@ project('spring-xd-module-plugin') {
  		runtime "org.springframework.build.gradle:propdeps-plugin:${versions['org.springframework.build.gradle:propdeps-plugin']}"
  	}
  	compileGroovy {
-		sourceCompatibility = 1.6
-		targetCompatibility = 1.6
+		sourceCompatibility = project.compileJava.sourceCompatibility
+		targetCompatibility = project.compileJava.targetCompatibility
 	}
 }
 
--- a/gradle/build-gradle-plugins.gradle
+++ b/gradle/build-gradle-plugins.gradle
@@ -23,8 +23,11 @@ propertiesFile.withInputStream {
 	versions.load(it)
 }
 
+//apply from: 'publish-maven.gradle'
+//task install(dependsOn:'publishToMavenLocal'){}
+
 project('spring-xd-module-plugin') {
-	apply plugin: 'maven'
+	startScripts.setEnabled(false)
     description = 'Gradle plugin to build a Spring XD Module'
 	dependencies {
  		compile gradleApi()
--- a/publish-maven.gradle
+++ b/publish-maven.gradle
@@ -1,7 +1,7 @@
 apply plugin: 'maven-publish'
 
 //alias install -> publishToMavenLocal
-task install(dependsOn:'publishToMavenLocal'){}
+task install(dependsOn: 'publishToMavenLocal') {}
 
 // exclude hadoop install sub projects
 def hadoopInstalls = ['cdh5', 'hadoop26', 'hdp22', 'phd21', 'phd30']
@@ -9,14 +9,7 @@ if (hadoopInstalls.contains(project.name)) {
 	project.tasks.findByPath("artifactoryPublish")?.enabled = false
 }
 
-
-
-// load versions
-def versions = new Properties()
-def propertiesFile = new File('dependencies.properties')
-propertiesFile.withInputStream {
-	versions.load(it)
-}
+def nonPublishedProjects = hadoopInstalls + 'documentation-toolchain'
 
 def customizePom = {
 	description project.description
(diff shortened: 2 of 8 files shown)


gradle/build-gradle-plugins.gradle
```diff
@@ -23,8 +23,11 @@ propertiesFile.withInputStream {
 	versions.load(it)
 }
 
+//apply from: 'publish-maven.gradle'
+//task install(dependsOn:'publishToMavenLocal'){}
```
ericbottard: I this meant to be commented out?
Seems like it's in `publish-maven.gradle` now?

publish-maven.gradle
```diff
@@ -1,172 +1,152 @@
...
+		groupId dep.group
+		version dep.version
+		scope provided
+	}
+}
+
+providedDep.resolveStrategy = Closure.DELEGATE_ONLY
+
+//Note setting enabled=false on publishing tasks doesn't work as expected
+if (!(hadoopInstalls + 'documentation-toolchain').contains(project.name)) {
```
ericbottard: Should we rather have a `non-published-projects` variable or something?

publish-maven.gradle
```diff
@@ -1,172 +1,152 @@
...
 
-		// eliminate test-scoped dependencies (no need in maven central poms)
-		generatedPom.dependencies.removeAll { dep ->
-			dep.scope == 'test'
-		}
+// load versions
+def versions = new Properties()
+def propertiesFile = new File('dependencies.properties')
+propertiesFile.withInputStream {
+	versions.load(it)
```
ericbottard: This is not new with this PR, but I think instead of being file-centric for the versions resolution, we should somehow introspect the gradle model and look up what gradle ended up using. This is to avoid surprises the day we finally drop that properties file, what do you think?

publish-maven.gradle
```diff
@@ -1,172 +1,152 @@
...
+
+					asNode().dependencies.'*'.each {
+						// Add version as the 3rd element
+						it.version ||
+							it.children().add(2,
+								new Node(null, 'version', versions["${it.groupId.text()}:${it.artifactId.text()}"]))
+					}
+
+					project.configurations.optional.allDependencies.each { dep ->
+						def mvnDep = asNode().dependencies.'*'.find { it.artifactId.text() == dep.name }
```
ericbottard: May not happen often, but aren't we worried to match false positives if looking at the artifactId only? Same thing throughout this method.

publish-maven.gradle
```diff
@@ -1,172 +1,152 @@
...
+						new Node(mvnDep, 'optional', 'true')
+					}
+					project.configurations.provided.allDependencies.each { dep ->
+						println "provided: ${dep.name}"
```
ericbottard: "debug" statement to be removed

publish-maven.gradle
```diff
@@ -1,172 +1,152 @@
...
+						def mvnDep = asNode().dependencies.'*'.find {
+							it.artifactId.text() == dep.name && it.scope?.text() == 'runtime'
```
ericbottard: I'm not sure I understand all these checks. In particular, why catch misaligned scopes **only** if it was runtime. Seems like if we want `provided`, we want `provided` no matter what was there before, no?

spring-xd-starters/spring-xd-module-parent/publish-maven.gradle
```diff
@@ -25,153 +24,169 @@ def getAllDependentProjects(project) {
...
-						}
-						exclusion {
-							groupId 'org.springframework.xd'
-							artifactId 'spring-xd-spark-streaming'
+	inceptionYear '2014'
+	packaging 'pom'
+	description project.description
+	properties {
+		'project.build.sourceEncoding' 'UTF-8'
+		'java.source.version' '1.7'
```
ericbottard: Do you think we can read that version from elsewhere in our main gradle build? I foresee hard to spot issues the day we move to java 8 (like, in 2022)

gradle/build-gradle-plugins.gradle
dturanski: Hmmm... probably should be removed

publish-maven.gradle
dturanski: I'll take a look if that's possible. I agree that would be better. Also will avoid rework.

publish-maven.gradle
dturanski: Makes sense. I'll look into it

spring-xd-starters/spring-xd-module-parent/publish-maven.gradle
dturanski: ok

ericbottard:
Rebased, squashed and merged as 6ddcec8

dturanski:
./gradlew clean dist works for me. Did not verify the artifacts though

ericbottard:
Merged as 358114f

PR #1605, opened 2015-05-11
[Sprint: 49] XD-3030 Integrate RabbitMQ queue cleaning API

twoseat:
Previously Spring XD required the user to access a stand-alone REST API to delete RabbitMQ queues that had been created by Spring XD. This change adds support for the API call to SpringXDTemplate.

### DIFF
--- a/spring-xd-rest-client/src/main/java/org/springframework/xd/rest/client/ResourceOperations.java
+++ b/spring-xd-rest-client/src/main/java/org/springframework/xd/rest/client/ResourceOperations.java
@@ -50,4 +50,10 @@ public interface ResourceOperations {
 	 */
 	public void destroyAll();
 
+	/**
+	 * Delete Queues and DLQs for a stream, and any exchanges created for taps.
+	 */
+	public void cleanBusResources(String name);
+
+
 }
--- a/spring-xd-rest-client/src/main/java/org/springframework/xd/rest/client/impl/StreamTemplate.java
+++ b/spring-xd-rest-client/src/main/java/org/springframework/xd/rest/client/impl/StreamTemplate.java
@@ -16,14 +16,10 @@
 
 package org.springframework.xd.rest.client.impl;
 
-import java.util.Collections;
-import java.util.Map;
-
 import org.springframework.util.LinkedMultiValueMap;
 import org.springframework.util.MultiValueMap;
 import org.springframework.xd.rest.client.StreamOperations;
 import org.springframework.xd.rest.domain.StreamDefinitionResource;
-import org.springframework.xd.rest.domain.support.DeploymentPropertiesFormat;
 
 /**
  * Implementation of the Stream-related part of the API.
@@ -31,10 +27,10 @@ import org.springframework.xd.rest.domain.support.DeploymentPropertiesFormat;
  * @author Eric Bottard
  * @author Ilayaperumal Gopinathan
  */
-public class StreamTemplate extends AbstractTemplate implements StreamOperations {
+public class StreamTemplate extends AbstractResourceTemplate implements StreamOperations {
 
-	StreamTemplate(AbstractTemplate source) {
-		super(source);
+	StreamTemplate(AbstractTemplate source, String adminUri, String password, String username, String vhost) {
+		super(source, adminUri, password, "streams", username, vhost);
 	}
 
 	@Override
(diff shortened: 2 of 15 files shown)


spring-xd-rest-client/src/main/java/org/springframework/xd/rest/client/ResourceOperations.java
```diff
@@ -50,4 +50,10 @@
 	 */
 	public void destroyAll();
 
+	/**
+	 * Delete a RabbitMQ queue.
+	 */
+	public void deleteQueue(String name, Map<String, String> properties);
```
garyrussell: `deleteQueue` is not really all it does; it deletes all the queues (and DLQs) for the stream as well as any exchanges that were created for taps.

Suggest `clean`, `cleanBusResources` or similar.

spring-xd-rest-client/src/main/java/org/springframework/xd/rest/client/impl/StreamTemplate.java
```diff
@@ -95,4 +97,17 @@ public void destroyAll() {
...
 	}
 
+	@Override
+	public void deleteQueue(String name, Map<String, String> properties) {
+		String rootUri = resources.get("streams/clean/rabbit").toString() + "/{name}";
+		UriComponentsBuilder builder = UriComponentsBuilder.fromUriString(rootUri);
+
+		for(Map.Entry<String, String> entry: properties.entrySet()) {
+			builder.queryParam(entry.getKey(), entry.getValue());
+		}
```
garyrussell: It would be nice if the template could have some (maybe) all of these properties wired in as properties since they are likely to be the same for each call.

Ditto for the job implementation - however, it would be better to pull the guts of this method up to `AbstractTemplate` since the only difference between the two templates is the `rootUri`.

spring-xd-rest-client/src/test/java/org/springframework/xd/rest/client/impl/JobTemplateTest.java
```diff
@@ -0,0 +1,52 @@
...
+                Collections.singletonMap("test-property-key", "test-property-value"));
+
+		this.mockServer.verify();
+	}
+
+	@Controller
+	private static final class StubController {
+	}
+
+}
```
garyrussell: Nitpick - all files need a final newline.

spring-xd-rest-client/src/main/java/org/springframework/xd/rest/client/ResourceOperations.java
twoseat: Good point, I'll change to cleanBusResources, thanks

spring-xd-rest-client/src/test/java/org/springframework/xd/rest/client/impl/JobTemplateTest.java
twoseat: My mistake, will fix

spring-xd-rest-client/src/main/java/org/springframework/xd/rest/client/impl/StreamTemplate.java
twoseat: 1. Happy to do this, but it seems odd to require all users to provide their RabbitMQ details given that they may not want this feature (and presumably so far few or none have used it). Let me know if you want me to go ahead with the change anyway (for stream and job).

2. Similarly happy to do this - I was trying to stick with the pattern set by the two classes, which keep their methods separate despite such commonality (e.g. deploy/undeploy/destroy etc.) Can you confirm that you want me to break that pattern?

ericbottard: Will need to introduce an intermediary class, as `AbstractTemplate` is really just the base class for REST support.

garyrussell: 1. I am not sure what you mean; I meant add the ability to optionally configure the rabbitmq details once, during template creation (with a bunch of setters), and then merge the runtime properties with the configuration properties.

2. Fine with me (intermediate class); I simply abhor seeing essentially identical code in sibling classes - it's a maintenance nightmare.

nebhale: > Fine with me (intermediate class); I simply abhor seeing essentially identical code in sibling classes - it's a maintenance nightmare.

Just to confirm with you @garyrussell, you'd like to see us break [the pattern set](https://github.com/spring-projects/spring-xd/blob/master/spring-xd-rest-client/src/main/java/org/springframework/xd/rest/client/impl/StreamTemplate.java#L72-L96) by the [rest of the code](https://github.com/spring-projects/spring-xd/blob/master/spring-xd-rest-client/src/main/java/org/springframework/xd/rest/client/impl/JobTemplate.java#L111-L133) in the templates?

garyrussell: Ideally I'd like to see the other violations corrected too but we can do that (and perhaps this one) via a technical debt JIRA issue, given that you have gone out of your way to provide this PR.

twoseat:
@garyrussell Because we are one Pivotal, we've done the refactoring you requested :smile:.  We've pulled the shared Resources code up and created a hierarchy for the Metrics implementations as well.

spring-xd-rest-client/src/main/java/org/springframework/xd/rest/client/ResourceOperations.java
```diff
@@ -50,4 +50,10 @@
 	 */
 	public void destroyAll();
 
+	/**
+	 * Delete Queues and DLQs for a stream, and any exchanges created for taps.
```
garyrussell: We should make is bus-agnostic - will fix during merge; something like:

"Remove any resources that cannot automatically be cleaned up during undeploy"

We might need to extend it to other bus implementations in future.

garyrussell:
Thanks!! @twoseat @nebhale 

LGTM; minor polish here: https://github.com/garyrussell/spring-xd/commit/a90aef0bf334eaead2d11e5436d561f1eea35888

(mainly copyrights etc, but also a few minor code changes and a bunch of whitespace cleanup).

Ready to merge after you approve the polishing.

I suggest you review with `?w=1` to avoid most of the whitespace noise:

https://github.com/garyrussell/spring-xd/commit/a90aef0bf334eaead2d11e5436d561f1eea35888?w=1

spring-xd-rest-client/src/main/java/org/springframework/xd/rest/client/ResourceOperations.java
mbogoevici: Interesting - maybe we'll find a way to leverage that for Kafka topics too. When, you know, deletion works or something.

twoseat:
Thanks for doing all that polishing @garyrussell, it looks really good except for one thing - the formatting on [SpringXDTemplate](https://github.com/garyrussell/spring-xd/blob/a90aef0bf334eaead2d11e5436d561f1eea35888/spring-xd-rest-client/src/main/java/org/springframework/xd/rest/client/impl/SpringXDTemplate.java) looks odd in GitHub. If that's just a web quirk then no worries, I'm happy for this to go ahead.

garyrussell:
Thanks; can you clarify "odd" ?

All I did with that file is add the extra constructor arg (`busPrefix`) to the "full" constructor and move the minimal arg ctors up above the "full" ctor (which is our standard).

twoseat:
Actually I can't, because it's not happening any more. I guess it really was a web rendering thing!

garyrussell:
Merged; thanks.

PR #1609, opened 2015-05-12
[Sprint: 49] XD-3019: Document HDFS module registry

ericbottard:


src/docs/asciidoc/Sources.asciidoc
```diff
@@ -1054,14 +1054,26 @@ The **$$kafka$$** $$source$$ has the following options:
 
 $$autoOffsetReset$$:: $$strategy to reset the offset when there is no initial offset in ZK or if an offset is out of range$$ *($$AutoOffsetResetStrategy$$, default: `smallest`, possible values: `smallest,largest`)*
```
cppwfs: So do we need the Sources.asciidoc as a part of the PR?

ericbottard: huh, no. I fixed that already. If you rebase, it should disappear

src/docs/asciidoc/Modules.asciidoc
```diff
@@ -264,6 +265,22 @@ In cases where you want to start e.g. a single-node runtime with a custom module
 `XD_CUSTOMMODULE_HOME=file\:/path/to/custom-modules bin/xd-singlenode`
 ====
 
+[[replicating-module-registry]]
+==== Replicating Module Registry
+When running in distributed mode, an alternative to using a shared file system for custom modules is to use the _replicating module registry_.
+
+If the value of `xd.customModule.home` does not use the `file:` protocol, then Spring XD will automatically set up a replicating registry that proxies that remote registry (at the end of the day, all modules need to be read from local filesystem anyway). This is all done transparently and by default, files are copied down from the central repository only if their contents has changed.
```
cppwfs: We can remove "(at the end of the day, all modules need to be read from local filesystem anyway)".  This is covered below.

src/docs/asciidoc/Modules.asciidoc
```diff
@@ -264,6 +265,22 @@ In cases where you want to start e.g. a single-node runtime with a custom module
...
+At the time of writing, only the `hdfs:` protocol is supported. Setting this up is straightforward:
+
+[source,yml]
+----
+xd:
+  customModule:
+    home: hdfs://somehost/root/path/of/registry
+----
+
+Files will be replicated on the local filesystem in a temporary directory, on demand and loaded from there. Obviously, the XD Admin process will need to have write access to that shared HDFS directory. Intermediary paths (`/root/path/of/registry` in the example above) are created at startup if they don't exist yet.
```
cppwfs: "Obviously, the"  We can remove the word Obviously.

src/docs/asciidoc/Modules.asciidoc
markfisher: We should probably avoid the figure of speech "at the end of the day" in general, as it sounds like we're describing an old school batch process ;)

ericbottard: Can I tweet that? :)

More to Glenn's point, even if the form may be changed, I think we need to explain *why* this registry is synchronizing (and why it needs to)

src/docs/asciidoc/Modules.asciidoc
ericbottard: You're right, even more so because at this time, the connection details to HDFS are very basic, so it's not even "correct" (as in there is no authentication against the unix user running the process)

ericbottard:
Rebased and changed per review comments

cppwfs:
LGTM

cppwfs:
Rebased,squashed, Merged

PR #1614, opened 2015-05-13
[Sprint 49] XD-3054 Validate when processing deployment message

ilayaperumalg:
- Add deployment validation when the deployment message is consumed
by the deployment message consumer. Previously this check was only when
the deployment message is published.

### DIFF
--- a/spring-xd-dirt/src/main/java/org/springframework/xd/dirt/stream/AbstractDeployer.java
+++ b/spring-xd-dirt/src/main/java/org/springframework/xd/dirt/stream/AbstractDeployer.java
@@ -89,11 +89,13 @@ public abstract class AbstractDeployer<D extends BaseDefinition> implements Reso
 	@Override
 	public D save(D definition) {
 		Assert.notNull(definition, "Definition may not be null");
+		String name = definition.getName();
+		String def = definition.getDefinition();
+		validateBeforeSave(name, def);
 		if (repository.findOne(definition.getName()) != null) {
 			throwDefinitionAlreadyExistsException(definition);
 		}
-		List<ModuleDescriptor> moduleDescriptors = parser.parse(definition.getName(),
-				definition.getDefinition(), definitionKind);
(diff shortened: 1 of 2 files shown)


spring-xd-dirt/src/main/java/org/springframework/xd/dirt/stream/AbstractDeployer.java
```diff
@@ -89,11 +89,13 @@ protected AbstractDeployer(ZooKeeperConnection zkConnection, PagingAndSortingRep
 	@Override
 	public D save(D definition) {
 		Assert.notNull(definition, "Definition may not be null");
+		String name = definition.getName();
+		String def = definition.getDefinition();
+		validateBeforeSave(name, def);
 		if (repository.findOne(definition.getName()) != null) {
```
pperalta: Doesn't `validateBeforeSave` do the same thing as:
```
if (repository.findOne(definition.getName()) != null) {
	throwDefinitionAlreadyExistsException(definition);
}
```

spring-xd-dirt/src/main/java/org/springframework/xd/dirt/stream/AbstractInstancePersistingDeployer.java
```diff
@@ -67,6 +68,7 @@ protected void beforeDelete(D definition) {
 	@Override
 	public void undeploy(String name) {
 		Assert.hasText(name, "name cannot be blank or null");
+		validateBeforeUndeploy(name);
```
pperalta: Same here, it seems that `validateBeforeUndeploy` is doing the exact same thing as the code below.

ilayaperumalg:
@pperalta Pushed the changes that remove the duplicate validation.

pperalta:
LGTM; squashed, rebased, and merged as 9161736a78a59ed082227da661fc7e5c375a63ec

PR #1598, opened 2015-05-05
[Sprint: 49] XD-2899: Rabbit HA - Connect to Queue Node

garyrussell:
JIRA: https://jira.spring.io/browse/XD-2899

Enable configuration such that the bus will connect to the rabbitmq node that hosts a queue.

PR #1616, opened 2015-05-14
[Sprint: 49; AFTER #1598] XD-3048: Fix/Support Wild Card Rabbit Bus Cleanup 

garyrussell:
Previously, the `RabbitBusCleaner` incorrectly removed all queues starting with
the requested stream name (although it would not remove in-use queues).

Only remove resources pertaining to the requested stream/job.
    
Also add support for a simple wild card `*` at the end of the name, to delete
resources for all streams/jobs that match that pattern.

__NOTE: DO NOT MERGE UNTIL #1598 IS MERGED__

This PR has conflicts with that PR and contains its commits. We will need to rebase/cherry-pick after that PR is merged.


mbogoevici:
Merged w/https://github.com/spring-projects/spring-xd/commit/52fd38fdaddb73f011c243138dfe929a6047343c

PR #1622, opened 2015-05-15
[Sprint: 49] XD-3052 Move gpfdist sink from spring-xd-modules to spring-xd

markpollack:


jvalkeal:
Ran my usual tests etc, looks good, merged.

PR #1623, opened 2015-05-15
[Sprint: 49] XD-2309: Added incremental import to JDBCHDFS job

mminella:
This update adds a new partitioner that keeps track of the last value in
the checkColumn and uses that as the minimum value for the next import.
This commit serves as the minimum implementation of this functionality.
Current limitations include:

1. Incremental imports are only supported for job definitions that use
the column/table notation (providing SQL is not supported).
2. The checkColumn value must be numeric.


trisberg:
I tried using the overrideCheckColumnValue but it is inconvenient since I have to un-deploy the original job, create a new one using the same name and parameters adding the --overrideCheckColumnValue. Also, since the override is now in the job definition, I would have to do this each time I wanted to run the job again, updating the override value or it will start off at the previously defined override value. It would be better to provide the option of overriding the check column value using an optional job parameter when the job is launched.

trisberg:
If I run the job without adding any rows to the database table I get an NPE - 

```
java.lang.NullPointerException
	at org.springframework.batch.integration.x.IncrementalColumnRangePartitioner.beforeStep(IncrementalColumnRangePartitioner.java:224)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAccessorImpl.invoke(NativeMethodAccessorImpl.java:57)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:606)
	at org.springframework.aop.support.AopUtils.invokeJoinpointUsingReflection(AopUtils.java:317)
	at org.springframework.aop.framework.ReflectiveMethodInvocation.invokeJoinpoint(ReflectiveMethodInvocation.java:190)
	at org.springframework.aop.framework.ReflectiveMethodInvocation.proceed(ReflectiveMethodInvocation.java:157)
	at org.springframework.aop.support.DelegatingIntroductionInterceptor.doProceed(DelegatingIntroductionInterceptor.java:133)
	at org.springframework.aop.support.DelegatingIntroductionInterceptor.invoke(DelegatingIntroductionInterceptor.java:121)
	at org.springframework.aop.framework.ReflectiveMethodInvocation.proceed(ReflectiveMethodInvocation.java:179)
	at org.springframework.aop.framework.JdkDynamicAopProxy.invoke(JdkDynamicAopProxy.java:207)
	at com.sun.proxy.$Proxy133.beforeStep(Unknown Source)
	at org.springframework.batch.core.listener.CompositeStepExecutionListener.beforeStep(CompositeStepExecutionListener.java:77)
	at org.springframework.batch.core.step.AbstractStep.execute(AbstractStep.java:194)
	at org.springframework.batch.core.job.SimpleStepHandler.handleStep(SimpleStepHandler.java:148)
	at org.springframework.batch.core.job.flow.JobFlowExecutor.executeStep(JobFlowExecutor.java:64)
	at org.springframework.batch.core.job.flow.support.state.StepState.handle(StepState.java:67)
	at org.springframework.batch.core.job.flow.support.SimpleFlow.resume(SimpleFlow.java:165)
	at org.springframework.batch.core.job.flow.support.SimpleFlow.start(SimpleFlow.java:144)
	at org.springframework.batch.core.job.flow.FlowJob.doExecute(FlowJob.java:134)
	at org.springframework.batch.core.job.AbstractJob.execute(AbstractJob.java:304)
	at org.springframework.batch.core.launch.support.SimpleJobLauncher$1.run(SimpleJobLauncher.java:135)
	at org.springframework.core.task.SyncTaskExecutor.execute(SyncTaskExecutor.java:50)
	at org.springframework.batch.core.launch.support.SimpleJobLauncher.run(SimpleJobLauncher.java:128)
	at sun.reflect.NativeMethodAccessorImpl.invoke0(Native Method)
	at sun.reflect.NativeMethodAcc
```

mminella:
The NPE..do you mean no records in the source table?

trisberg:
No new records with check value >  batch.incremental.maxId

trisberg:
Also had one strange run where with 3 partitions, the order they ran in was 2, 1, 0 and the 0 partitioned failed complaining the output directory already existed:

329	step1-master	105525	104525	106	1	738 ms	FAILED	
330	step1-master:partition2	52262	52262	53	0	635 ms	COMPLETED	
331	step1-master:partition1	52263	52263	53	0	654 ms	COMPLETED	
332	step1-master:partition0	1000	0	0	1	53 ms	FAILED

Step Execution Details - Step Execution ID: 332

Exit Description

```
java.lang.IllegalStateException: Cannot create directory /xd/jdbc2/: File exists
	at org.springframework.data.hadoop.fs.FsShell.mkdir(FsShell.java:631)
	at org.springframework.data.hadoop.fs.FsShell.mkdir(FsShell.java:619)
	at org.springframework.xd.batch.item.hadoop.AbstractHdfsItemWriter.initializeCounterIfNecessary(AbstractHdfsItemWriter.java:87)
	at org.springframework.xd.batch.item.hadoop.HdfsTextItemWriter.write(HdfsTextItemWriter.java:60)
	at sun.reflect.GeneratedMethodAccessor81.invoke(Unknown Source)
	at sun.reflect.DelegatingMethodAccessorImpl.invoke(DelegatingMethodAccessorImpl.java:43)
	at java.lang.reflect.Method.invoke(Method.java:606)
	at org.springframework.aop.support.AopUtils.invokeJoinpointUsingReflection(AopUtils.java:317)
	at org.springframework.aop.framework.ReflectiveMethodInvocation.invokeJoinpoint(ReflectiveMethodInvocation.java:190)
	at org.springframework.aop.framework.ReflectiveMethodInvocation.proceed(ReflectiveMethodInvocation.java:157)
	at org.springframework.aop.support.DelegatingIntroductionInterceptor.doProceed(DelegatingIntroductionInterceptor.java:133)
	at org.springframework.aop.support.DelegatingIntroductionInterceptor.invoke(DelegatingIntroductionInterceptor.java:121)
	at org.springframework.aop.framework.ReflectiveMethodInvocation.proceed(ReflectiveMethodInvocation.java:179)
	at org.springframework.aop.framework.JdkDynamicAopProxy.invoke(JdkDynamicAopProxy.java:207)
	at com.sun.proxy.$Proxy128.write(Unknown Source)
	at org.springframework.batch.core.step.item.SimpleChunkProcessor.writeItems(SimpleChunkProcessor.java:175)
	at org.springframework.batch.core.step.item.SimpleChunkProcessor.doWrite(SimpleChunkProcessor.java:151)
	at org.springframework.batch.core.step.item.SimpleChunkProcessor.write(SimpleChunkProcessor.java:274)
	at org.springframework.batch.core.step.item.SimpleChunkProcessor.process(SimpleChunkProcessor.java:199)
	at org.springframework.batch.core.step.item.ChunkOrientedTasklet.execute(ChunkOrientedTasklet.java:75)
	at org.springframework.batch.core.step.tasklet.TaskletStep$ChunkTransactionCallback.doInTransaction(TaskletStep.java:406)
	at org.springframework.batch.core.step.tasklet.TaskletStep$ChunkTransactionCallback.doInTransaction(TaskletStep.java:330)
	at org.springframework.transaction.support.TransactionTemplate.execute(TransactionTemplate.java:133)
	at org.springframework.batch.core.step.tasklet.TaskletStep$2.doInChunkContext(TaskletStep.java:271)
	at org.springframework.batch.core.scope.contex
```

trisberg:
Any thought about using Long instead of Integer for partitioning and check columns? Other than that LGTM.

trisberg:
Also thought I had some additional comments about removing the `ColumnRangePartitioner`, but I don't see them any more. Where did they go?

mminella:
I have no issues with making the change from long to integer.  I left the ColumnRangePartitioner there because I didn't do an exhaustive search to see if it was used elsewhere.  If it isn't, yes, it should be removed (I don't know where the comments about that went).

trisberg:
It's not used anywhere else, so we should just remove it and change the test to test the new partitioner.

mminella:
One question about the `ColumnRangePartitioner` vs the `IncrementalColumnRangePartitioner`.  The former is in spring-xd-batch and the latter is in extensions/spring-xd-extension-batch.  I'm not sure what should go in each.  I put the new one where it is because there were other partitioners there and this seems like an extension and not something needed for core batch functionality for XD but I could be wrong.  Thoughts?

trisberg:
I think having the partitioner in extensions is actually better

mminella:
Ok...I moved the tests to extensions instead of moving my partitioner :)

trisberg:
Last two issues:

Line 207 of `IncrementalColumnRangePartitioner` should be getLong and not getInt :)
this.incrementalMin = lastExecution.getExecutionContext().getLong(BATCH_INCREMENTAL_MAX_ID);

When the job/steps FAIL the max id isn't set in the job execution context so we loose a previous max value. It is written to the step execution context of the partition step but not the job execution context.

trisberg:
The step execution context has the new max value, so I guess we should put the old value in the job execution context when we start and then update it when the job is successful, what do you think?

mminella:
With regards to job failures, my thinking is that the value shouldn't be promoted until the job succeeds...If you want to override that, you can use the override function.  If we promote it on a failed job, we'd end up skipping the records (the next run would pick up where the previous job *should* have left off...not where it actually did) that did not get imported due to the failure which isn't correct IMHO.

trisberg:
I think we should start with the max value of the last successful job. The way it works now is that if a job fails we ignore any history since we only check the last execution. The job before the failed one could have been successful and could contain a max id that we want to start from. 

I'm experimenting with checking the last execution:

 lastExecution.getExitStatus().equals(ExitStatus.COMPLETED)

and skipping any that aren't COMPLETED

mminella:
How would that work with a job that imported half of the records before failure?  For example:

* **run1:** min = 0, max = 100 - COMPLETE
* **run 2:** min = 100, max = 200 - FAIL after 50
* **run 3**
    * my way: You either override or must restart run 2
    * your way: min = 200, max = 300 - We end up missing 150-200

Am I missing something?

trisberg:
my way we would ignore run2 and start from 100 or the user could override and start from 150 if they verified that that was the last id

mminella:
How do you handle the already imported 100-150?

trisberg:
you have to do it manually AFAICT - the current implementation starts from 0 unless I override, my implementation starts from 100 unless I override

mminella:
What about adding a strict mode flag that allows the user to skip the restart of run 2 and use the last successful value (strict = false) or fail indicating that there are records that may not have been imported (strict = true).  Either flag setting would be ignored when using the override option.  Thoughts?

trisberg:
how would I restart a failed run? Don't see that option.

mminella:
The same way you restart a failed job now...

`xd:> job execution restart --id <failed_id>`

trisberg:
org.springframework.batch.core.repository.JobRestartException: JobInstance already exists and is not restartable

mminella:
Ah...restartable is a job option that must be passed in and defaults to false...that seems to be a general XD thing (I'm not sure I agree with that...)

trisberg:
LGTM, fixed `IncrementalColumnRangePartitioner` to use getLong, squashed and merged

(if we want to tweak the handling of failed runs we can open a new JIRA)

PR #1624, opened 2015-05-15
[Sprint: 49] XD-2850 Improve file source to efficiently read files

ghillert:
* replace `ref` option with a `mode` option
* `mode` can accept
  - `ref` (same behavior as the former `ref=true`)
  - `fileAsBytes` (same behavior as the former `ref=false`)
  - textLine (New option, will split files line-by-line and emit a new message for each line)
* Add tests
* Refactor several existing tests
* Update documentation

### DIFF
--- a/modules/common/file-source-common-context.xml
+++ b/modules/common/file-source-common-context.xml
@@ -1,0 +1,50 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<beans:beans xmlns="http://www.springframework.org/schema/integration"
+	xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+	xmlns:beans="http://www.springframework.org/schema/beans"
+	xmlns:file="http://www.springframework.org/schema/integration/file"
+	xsi:schemaLocation="http://www.springframework.org/schema/beans
+		http://www.springframework.org/schema/beans/spring-beans.xsd
+		http://www.springframework.org/schema/integration
+		http://www.springframework.org/schema/integration/spring-integration.xsd
+		http://www.springframework.org/schema/integration/file
+		http://www.springframework.org/schema/integration/file/spring-integration-file.xsd">
+
+	<channel id="output" />
+
+	<channel id="files" />
+
+	<beans:bean id="trigger" class="org.springframework.scheduling.support.PeriodicTrigger">
+		<beans:constructor-arg value="${fixedDelay}" />
+		<beans:constructor-arg value="${timeUnit}" />
+		<beans:property name="initialDelay" value="${initialDelay} "/>
+	</beans:bean>
+
+	<beans:beans profile="use-ref">
+		<bridge input-channel="files" output-channel="output"/>
+	</beans:beans>
+
+	<beans:beans profile="use-contents">
+		<chain id="extractContents" input-channel="files" output-channel="output">
+			<header-enricher>
+				<header name="contentType" value="application/octet-stream"/>
+			</header-enricher>
+			<file:file-to-bytes-transformer />
+		</chain>
+	</beans:beans>
+
+	<beans:beans profile="use-contents-with-split">
+		<chain id="extractContentsAndSplit" input-channel="files" output-channel="output">
+			<header-enricher>
+				<header name="contentType" value="text/plain"/>
+				<header name="file_name"   expression="payload.name"/>
+			</header-enricher>
+			<splitter>
+				<beans:bean class="org.springframework.integration.file.splitter.FileSplitter">
+					<beans:constructor-arg value="false"/>
+				</beans:bean>
+			</splitter>
+		</chain>
+	</beans:beans>
+
+</beans:beans>
--- a/modules/common/file-source-common-context.xml
+++ b/modules/common/file-source-common-context.xml
@@ -40,9 +40,7 @@
 				<header name="file_name"   expression="payload.name"/>
 			</header-enricher>
 			<splitter>
-				<beans:bean class="org.springframework.integration.file.splitter.FileSplitter">
-					<beans:constructor-arg value="false"/>
-				</beans:bean>
+				<beans:bean class="org.springframework.integration.file.splitter.FileSplitter"/>
 			</splitter>
 		</chain>
 	</beans:beans>
--- a/modules/source/file/config/file.xml
+++ b/modules/source/file/config/file.xml
@@ -39,7 +39,20 @@
 			</header-enricher>
 			<file:file-to-bytes-transformer />
 		</chain>
+	</beans:beans>
 
+	<beans:beans profile="use-contents-with-split">
+		<chain id="extractContentsAndSplit" input-channel="files" output-channel="output">
+			<header-enricher>
+				<header name="contentType" value="text/plain"/>
+				<header name="file_name"   expression="payload.name"/>
+			</header-enricher>
+			<splitter>
+				<beans:bean class="org.springframework.integration.file.splitter.FileSplitter">
+					<beans:constructor-arg value="false"/>
+				</beans:bean>
+			</splitter>
+		</chain>
 	</beans:beans>
 
 </beans:beans>
--- a/modules/source/file/config/file.xml
+++ b/modules/source/file/config/file.xml
@@ -18,41 +18,6 @@
 		<poller trigger="trigger" />
 	</file:inbound-channel-adapter>
 
-	<channel id="output"/>
-
-	<channel id="files" />
-
-	<beans:bean id="trigger" class="org.springframework.scheduling.support.PeriodicTrigger">
-		<beans:constructor-arg value="${fixedDelay}" />
-		<beans:constructor-arg value="${timeUnit}" />
-		<beans:property name="initialDelay" value="${initialDelay} "/>
-	</beans:bean>
-
-	<beans:beans profile="use-ref">
-		<bridge input-channel="files" output-channel="output"/>
-	</beans:beans>
-
-	<beans:beans profile="use-contents">
-		<chain id="extractContents" input-channel="files" output-channel="output">
-			<header-enricher>
-				<header name="contentType" value="application/octet-stream"/>
-			</header-enricher>
-			<file:file-to-bytes-transformer />
-		</chain>
-	</beans:beans>
-
-	<beans:beans profile="use-contents-with-split">
-		<chain id="extractContentsAndSplit" input-channel="files" output-channel="output">
-			<header-enricher>
-				<header name="contentType" value="text/plain"/>
-				<header name="file_name"   expression="payload.name"/>
-			</header-enricher>
-			<splitter>
-				<beans:bean class="org.springframework.integration.file.splitter.FileSplitter">
-					<beans:constructor-arg value="false"/>
-				</beans:bean>
-			</splitter>
-		</chain>
-	</beans:beans>
+	<beans:import resource="../../../common/file-source-common-context.xml"/>
 
 </beans:beans>
--- a/spring-xd-dirt/src/main/java/org/springframework/xd/dirt/modules/metadata/FileReadingMode.java
+++ b/spring-xd-dirt/src/main/java/org/springframework/xd/dirt/modules/metadata/FileReadingMode.java
@@ -16,13 +16,6 @@
 
 package org.springframework.xd.dirt.modules.metadata;
 
-import java.util.Arrays;
-
-import org.springframework.util.Assert;
-import org.springframework.util.StringUtils;
-
-import com.google.common.base.Function;
-import com.google.common.collect.Collections2;
 
 /**
  *
@@ -31,12 +24,10 @@ import com.google.common.collect.Collections2;
  */
 
 public enum FileReadingMode {
-
-	REF("ref", "use-ref"),
-	TEXT_LINE("textLine", "use-contents-with-split"),
-	FILE_AS_BYTES("fileAsBytes", "use-contents");
-
-	private String key;
+	//ref, textLine, fileAsBytes
+	ref("use-ref"),
+	lines("use-contents-with-split"),
+	contents("use-contents");
 
 	private String profile;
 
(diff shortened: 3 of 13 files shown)


modules/source/file/config/file.xml
```diff
@@ -39,7 +39,20 @@
 			</header-enricher>
 			<file:file-to-bytes-transformer />
 		</chain>
+	</beans:beans>
 
+	<beans:beans profile="use-contents-with-split">
+		<chain id="extractContentsAndSplit" input-channel="files" output-channel="output">
+			<header-enricher>
+				<header name="contentType" value="text/plain"/>
```
ericbottard: We don't know for sure that lines are text/plain.

spring-xd-dirt/src/main/java/org/springframework/xd/dirt/modules/metadata/FileReadingMode.java
```diff
@@ -0,0 +1,85 @@
...
+
+/**
+ *
+ * @author Gunnar Hillert
+ * @since 1.2
+ */
+
+public enum FileReadingMode {
+
+	REF("ref", "use-ref"),
```
ericbottard: Better to not have that first constructor arg, use lowercase (or camelcase here) enum names and have the option be of type FileReadingMode instead of String. That way, we get validation and documentation of values for free

spring-xd-dirt/src/main/java/org/springframework/xd/dirt/modules/metadata/FileReadingMode.java
```diff
@@ -0,0 +1,85 @@
...
+ */
+
+package org.springframework.xd.dirt.modules.metadata;
+
+import java.util.Arrays;
+
+import org.springframework.util.Assert;
+import org.springframework.util.StringUtils;
+
+import com.google.common.base.Function;
```
jvalkeal: I think we should really avoid using guava stuff in our own classes. It's one of a worst libs for backward compatibility and source of most of a trouble in hadoop world.

modules/source/file/config/file.xml
jvalkeal: I guess not but it anyway goes to `bufferedReader.readLine()` method in SI so do we really have any other option?

ericbottard:
After second thought, I'm not sure I understand why we are doing this overall.
To me splitting by lines is orthogonal to the content type.
Also, and even though we have been doing this for other modules, it seems it violates the single responsibility principle. This strikes me even more as I am reviewing the FTP sink PR. To be symmetrical, should we add an aggregator on the file & ftp sink? What's wrong with letting the user introduce the splitter if they need to?

To clarify, I'm not 100% against this, but want to raise awareness that this seems to violate our "unix pipes" model a bit.

jvalkeal:
Simply because file source as it's done now is reading whole file in memory. Doesn't really work if you need to process very large files(10's GBs).

ericbottard:
Then maybe I'm missing something, but how does adding a splitter downwards prevent this?

garyrussell:
Correct; you can simply compose a module containing the file source (with ref) and a file splitter.

modules/source/file/config/file.xml
garyrussell: Right; the FileSplitter only supports text files.

jvalkeal:
I did some perf testing with gpfdist sink. There's something fishy here if we compare perf rates for one large file vs. split into smaller files vs. small file. We definitely do have some perf issue. Everything works thought which is a really good thing. Don't know if there's a delay between reading individual files or what's going on.

# 10M in 1 file
stream create --name filestogpfdist1 --definition "file --dir=/mnt/data2/tmp/xdfiles1 --mode=textLine |gpfdist --dbHost=mdw --table=xdsink --batchTimeout=2 --batchCount=1000 --batchPeriod=0 --flushCount=200 --flushTime=2 --rateInterval=1000000" --deploy

# 10M in 100 files
stream create --name filestogpfdist2 --definition "file --dir=/mnt/data2/tmp/xdfiles2 --mode=textLine |gpfdist --dbHost=mdw --table=xdsink --batchTimeout=2 --batchCount=1000 --batchPeriod=0 --flushCount=200 --flushTime=2 --rateInterval=1000000" --deploy

# 1M in 1 file
stream create --name filestogpfdist3 --definition "file --dir=/mnt/data2/tmp/xdfiles3 --mode=textLine |gpfdist --dbHost=mdw --table=xdsink --batchTimeout=2 --batchCount=1000 --batchPeriod=0 --flushCount=200 --flushTime=2 --rateInterval=1000000" --deploy

# filestogpfdist1
2015-05-15 11:55:32,265 1.2.0.SNAP  INFO DeploymentSupervisor-0 zk.ZKStreamDeploymentHandler - Deployment status for stream 'filestogpfdist1': DeploymentStatus{state=deployed}
2015-05-15 11:55:38,012 1.2.0.SNAP  INFO task-scheduler-6 gpfdist.GPFDistMessageHandler - METER: 1 minute rate = 0.0 mean rate = 162670.47593811943
2015-05-15 11:55:44,901 1.2.0.SNAP  INFO task-scheduler-6 gpfdist.GPFDistMessageHandler - METER: 1 minute rate = 15991.117074135342 mean rate = 153408.44077588603
2015-05-15 11:55:48,655 1.2.0.SNAP  INFO task-scheduler-6 gpfdist.GPFDistMessageHandler - METER: 1 minute rate = 30703.655021877174 mean rate = 178671.38038328252
2015-05-15 11:55:52,504 1.2.0.SNAP  INFO task-scheduler-6 gpfdist.GPFDistMessageHandler - METER: 1 minute rate = 44239.843385719 mean rate = 193802.312550707
2015-05-15 11:55:56,143 1.2.0.SNAP  INFO task-scheduler-6 gpfdist.GPFDistMessageHandler - METER: 1 minute rate = 44239.843385719 mean rate = 205939.92411929733
2015-05-15 11:55:59,628 1.2.0.SNAP  INFO task-scheduler-6 gpfdist.GPFDistMessageHandler - METER: 1 minute rate = 72684.85495937747 mean rate = 216107.17162910756
2015-05-15 11:56:03,147 1.2.0.SNAP  INFO task-scheduler-6 gpfdist.GPFDistMessageHandler - METER: 1 minute rate = 82864.41190765305 mean rate = 223766.74425689018
2015-05-15 11:56:06,461 1.2.0.SNAP  INFO task-scheduler-6 gpfdist.GPFDistMessageHandler - METER: 1 minute rate = 82864.41190765305 mean rate = 231232.27744470848
2015-05-15 11:56:09,849 1.2.0.SNAP  INFO task-scheduler-6 gpfdist.GPFDistMessageHandler - METER: 1 minute rate = 108221.17349545046 mean rate = 236934.94865924914
2015-05-15 11:56:13,198 1.2.0.SNAP  INFO task-scheduler-6 gpfdist.GPFDistMessageHandler - METER: 1 minute rate = 115559.4032932555 mean rate = 241935.3204787784

# filestogpfdist2
2015-05-15 11:58:02,568 1.2.0.SNAP  INFO DeploymentSupervisor-0 zk.ZKStreamDeploymentHandler - Deployment status for stream 'filestogpfdist2': DeploymentStatus{state=deployed}
2015-05-15 11:58:02,580 1.2.0.SNAP  INFO task-scheduler-3 gpfdist.GPFDistMessageHandler - METER: 1 minute rate = 0.0 mean rate = 4128358.390652287
2015-05-15 11:58:56,935 1.2.0.SNAP  INFO task-scheduler-5 gpfdist.GPFDistMessageHandler - METER: 1 minute rate = 94473.31054820298 mean rate = 36631.744361704106
2015-05-15 11:59:51,414 1.2.0.SNAP  INFO task-scheduler-5 gpfdist.GPFDistMessageHandler - METER: 1 minute rate = 44724.83139995858 mean rate = 27503.615630353095
2015-05-15 12:00:45,766 1.2.0.SNAP  INFO task-scheduler-10 gpfdist.GPFDistMessageHandler - METER: 1 minute rate = 24832.919208351188 mean rate = 24475.576024532136
(didn't bother to wait more)

# filestogpfdist3
2015-05-15 12:02:11,097 1.2.0.SNAP  INFO DeploymentSupervisor-0 zk.ZKStreamDeploymentHandler - Deployment status for stream 'filestogpfdist3': DeploymentStatus{state=deployed}
2015-05-15 12:02:11,180 1.2.0.SNAP  INFO task-scheduler-10 gpfdist.GPFDistMessageHandler - METER: 1 minute rate = 0.0 mean rate = 3115439.310636276

markpollack:
I don't think it is at all intuitive to expect the user to add a splitter in order to read the contents of a file,
So having it part of the file module itself provides that simple ease of use.

markpollack:
Janne, I've noticed that codahale metrics seem crazy in accurate over small time periods.  Maybe that is a possible reason for the results in the third stream?

garyrussell:
True, but in the sprit of DRY, it's unfortunate the config is repeated 3 times; maybe it's time that these profiles should be moved to a common `<include/>` file.

jvalkeal:
Mark, that's why codahale is only updated once per rateInterval. It was a first thing I tried with it that how it performs if you hit it with every message.

markfisher:
If we go ahead with this PR we at least need to simplify the "mode" values. For example, "fileAsBytes" is redundant considering this is the "file" source. I'd suggest these names for the 3 options: ref, content, lines (content is "bytes" upon reading, but ultimately the "outputType" option can change that, so I think there's no need to include bytes in the name for that one either).

That said, I tend to agree with Eric. Also, I'm concerned that many use-cases will require something more sophisticated than simple line-reading... the types of concerns that would be handled by delegating to Spring Batch components, such as specifying patterns for multi-line split points and ignoring headers (as well as the ability to use headers to drive the names of a Tuple, a very common requirements for CSV data). Perhaps a compromise would be a new source module that provides that level of sophistication, with a default of emitting simple line-by-line messages.

markpollack:
Fwiw in the discussion with Gunnar on this approach, it feels like we may need to revisit the approach as we will need to add support for reading binary formats (Avro, parquet, Orc) and have the ability to specify a registered (somehow) deserializer instance.   Removing the repeated config would be nice.

markpollack:
The fileAsBytes name is meant to indicate that all of he file goes into memory.  I felt just using bytes would not be appropriate in terms of some sort of future proofing form when we support binary formats.  Perhaps mode=inMemory?

markfisher:
My suggestion was "content". I think it's pretty clear that means the file content goes into a Message payload and therefore memory. Of course it could be documented clearly as well. It's pretty common when reading files/http-responses/etc whether to read all content in one go or use an InputStream. We could even provide a content-length header for the downstream processors to consider.

garyrussell:
@jvalkeal There's a 5 second delay between each file (by default). We need to expose `maxMessagesPerPoll` to avoid delays between polls; in the meantime, try setting the `fixedDelay=0' (but it will run away with a CPU when there are no files left).

garyrussell:
Or change the source thus:

		<poller trigger="trigger" max-messages-per-poll="100" />

markpollack:
Content as the value is fine by me

garyrussell:
https://jira.spring.io/browse/XD-3063

jvalkeal:
@garyrussell maybe we could have a poll algorithm which would gradually slow down if new files are not found. That would handle pressure from cpu and take new files as quickly as possible?

jvalkeal:
@garyrussell ah, did see your second comment until now.

spring-xd-dirt/src/main/java/org/springframework/xd/dirt/modules/metadata/FileReadingMode.java
ghillert: Will do. Also created: https://jira.spring.io/browse/XD-3066

garyrussell:
@ghillert Are you going to do the changes suggested here (new values for mode and pull the common config into an include file?)

If so, I will re-review then.

Thanks

markpollack:
Also should we exclude the contentType in the header, could be a line of json or XML?

spring-xd-dirt/src/main/java/org/springframework/xd/dirt/modules/metadata/FileReadingMode.java
ghillert: @ericbottard For validation purposes - with your approach, the error response is something like:

````
'org.springframework.xd.dirt.modules.metadata.FileReadingMode' for property 'mode'; nested exception is java.lang.IllegalStateException: Cannot convert value of type [java.lang.String] to required type [org.springframework.xd.dirt.modules.metadata.FileReadingMode] for property 'mode': no matching editors or conversion strategy found
````

Wouldn't it be nice (better UX) to also return a list of all the permissible enum options?

**Comment by markpollack** [comment]
Also, is it possible to not use guava?  @jvalkeal We do have guava in xd/lib ATM, prob via some hadoop dep that will go away in 1.3.

**Comment by ghillert** [comment]
Pushed Changes: 

* Simplify `FileReadingMode` enum (This also removes Guava usage)
* Change mode options to `ref`, `lines` and `contents`
* Update tests

Outstanding issue: Move common context-XML into a template file. Will do that shortly and run one more build before pushing.

**Comment by ghillert** [comment]
@garyrussell @markpollack PR Should be good now:

* Moved common files source configuration to `file-source-common-context.xml`
* Rebased with MASTER

**Comment by jvalkeal** [comment]
@markpollack yes, guava is prob via some deps and that's why I'm so strong for not depend on it in our own stuff. I don't want to get into same gigantic cluster*uck where hadoop umbrella ended up with it.

File: src/docs/asciidoc/Sources.asciidoc [hunk 4]
```diff
@@ -238,7 +238,12 @@ Some platforms, such as linux, send status messages to `stderr`. The tail module
 [[file]]
 === File
 
-The file source provides the contents of a File as a byte array by default but may be configured to provide the file reference itself.
+The file source provides the contents of a File as a byte array by default. However, this can be
+customzied using the `--mode` option:
+
+- *ref* Provides a `java.io.File` reference
+- *textLine* Will split files line-by-line and emit a new message for each line
+- *fileAsBytes* The default. Provides the contents of a file as a byte array
```
garyrussell: Needs fixing for new values; also add to FTP, SFTP

spring-xd-dirt/src/main/java/org/springframework/xd/dirt/modules/metadata/FileReadingMode.java
```diff
@@ -0,0 +1,46 @@
...
+
+
+/**
+ *
+ * @author Gunnar Hillert
+ * @since 1.2
+ */
+
+public enum FileReadingMode {
+	//ref, textLine, fileAsBytes
```
garyrussell: The value of this comment is not clear; plus it's wrong.

modules/common/file-source-common-context.xml
```diff
@@ -0,0 +1,50 @@
...
+
+	<beans:beans profile="use-contents-with-split">
+		<chain id="extractContentsAndSplit" input-channel="files" output-channel="output">
+			<header-enricher>
+				<header name="contentType" value="text/plain"/>
+				<header name="file_name"   expression="payload.name"/>
+			</header-enricher>
+			<splitter>
+				<beans:bean class="org.springframework.integration.file.splitter.FileSplitter">
+					<beans:constructor-arg value="false"/>
```
garyrussell: Remove this ctor arg - we __WANT__ to use an an iterator (which is the default for `FileSplitter` with the defauilt ctor; setting this to `false` reverts to an in-memory splitter.

garyrussell: Here https://github.com/garyrussell/spring-xd/commit/ba637fefbe45c46baca66e85b3bed76de6d136ce is a test case that illustrates the problem.

ghillert:
@garyrussell Rebased and push raised issues.

garyrussell:
asciidoc?

garyrussell:
sorry - 2 commits :smile:

garyrussell:
Merged as 729f818db24601de37a0d395f074317acf3411d9

PR #1626, opened 2015-05-15
[Sprint: 49] XD-2958-3022 Upgrade to Kafka 0.8.2.1 + fix partition creation

mbogoevici:
- Upgrade to Spring Integration Kafka 1.2.0.M1
- Upgrade to Kafka 0.8.2.1
- Change Kafka Sink options to match new producer settings;
- Kafka sink accepts now a byte[] by default.
- Change Message Bus options (including servers.yml and application:.yml) to match new producer settings;
- Elliminate a race condition in reading metadata when creating Message Bus consumers;
- Add `spring-integration-kafka` dependency to Spark Streaming tests, to avoid failure due to class visibility;
- Fix partitoning calculations with respect to consumer module concurrency (XD-3022)


markpollack:
merged as efa33e9ff493782f9fb2737659d6c5f2811059ec

PR #1630, opened 2015-05-18
[Sprint: 49] XD-3061/XD-2031 Remove id&timestamp from tuple, add putAll to TupleBuilder

markpollack:


markpollack:
Also uses `TupleToJsonStringConverter` as the default converter for the `toString` method in `DefaultTuple`.  `TupleBuilder`, sets the converter so it will have minimal impact since that is the most common way to create a tuple vs. using the ctor of `DefaultTuple` directly.

dturanski:
LGTM

mbogoevici:
Rebased, merged https://github.com/spring-projects/spring-xd/commit/dce34a51aa2c0afd0cdf09c3e6cecc62d492304e

PR #1632, opened 2015-05-19
[Sprint: 49] XD-3077 Set HDFS as custom module registry for YARN deployments

trisberg:
- Update docs

### DIFF
--- a/spring-xd-yarn/site/config/servers.yml
+++ b/spring-xd-yarn/site/config/servers.yml
@@ -13,12 +13,21 @@ xd:
   containerMemory: 512M
   containerJavaOpts: -XX:MaxPermSize=128m
   containerLocality: false
-  container:
-    groups: yarn
 
 spring:
   yarn:
     applicationBaseDir: /xd/yarn/
+
+---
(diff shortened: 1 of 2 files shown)


spring-xd-yarn/site/config/servers.yml
```diff
@@ -63,6 +72,9 @@ spring:
 #  hadoop:
 #    fsUri: hdfs://localhost:8020
 #    resourceManagerHost: localhost
+#    resourceManagerPort: 8032
+#    resourceManagerSchedulerAddress: ${spring.hadoop.resourceManagerHost:8030}
+#    jobHistoryAddress: ${spring.hadoop.resourceManagerHost:10020}
```
jvalkeal: where is `spring.hadoop.jobHistoryAddress` used? We don't have that in shdp conf shortcuts.

trisberg: It's not used, it is needed for custom batch jobs so I though we should include it here

jvalkeal: ah, ok. I'll need to add that to shdp as we talked a while back.

jvalkeal:
ok, looks good. I'll merge this in a bit.

jvalkeal:
merged

PR #1634, opened 2015-05-19
[Sprint 49] XD-3083 Support running DSLs from cmd file

ilayaperumalg:
- Now the Streams/JobsController saves the stream/job definitions while all
the deployment requests are published into ZK distributed queue.
 - Remove `create` and `createAndDeploy` actions from DeploymentAction enum

### DIFF
--- a/spring-xd-dirt/src/main/java/org/springframework/xd/dirt/rest/JobsController.java
+++ b/spring-xd-dirt/src/main/java/org/springframework/xd/dirt/rest/JobsController.java
@@ -17,6 +17,7 @@
 package org.springframework.xd.dirt.rest;
 
 import java.util.ArrayList;
+import java.util.Collections;
 import java.util.List;
 import java.util.Map;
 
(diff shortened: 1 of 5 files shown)


pperalta:
The implementation looks good. However, is it possible this will cause a regression similar to https://jira.spring.io/browse/XD-2671? For instance, consider the following sequence:

1. Stream is created and deployed.
2. Stream is destroyed
3. Stream is created and deployed

Since destruction is asynchronous, it is possible in step 3 to create the stream before it is destroyed.

ilayaperumalg:
@pperalta that's right. That's the reason I didn't remove `delete` and `deleteAll` action from DeploymentAction. Please see my comments here: https://jira.spring.io/browse/XD-3083. Given the deployment/un-deployment always happen asynchronously, we don't expect someone having `deploy`/`undeploy` and `destroy`/`create` (since destroy also involves `undeploy`) on the same file.

ilayaperumalg:
After some discussion with @pperalta, we are going with a different approach. Hence, closing this PR.

PR #1635, opened 2015-05-20
[Sprint: 49] XD-3072 Parser endpoint

aclement:
Modifications to the parse endpoint to make it work better for tools. It now:

- attempts to continue processing in the case of an error.

- records a parsed structure in the response and all exceptions that occurred (not just the first one)

If there is something still lacking it is that the NoSuchModuleException does not get record the position in the DSL string of where the problem module is. But let's get this first bit done first.

With these changes, this will actually return a structure:

     curl http://localhost:9393/tools/parse\?definitions\=time

as will this:

     curl http://localhost:9393/tools/parse\?definitions\=foo\=time\|l:%20log%20--name2\=foo

### DIFF
--- a/spring-xd-dirt/src/main/java/org/springframework/xd/dirt/stream/XDStreamParser.java
+++ b/spring-xd-dirt/src/main/java/org/springframework/xd/dirt/stream/XDStreamParser.java
@@ -17,7 +17,7 @@
 package org.springframework.xd.dirt.stream;
 
 
-import static org.springframework.xd.dirt.stream.dsl.XDDSLMessages.*;
+import static org.springframework.xd.dirt.stream.dsl.XDDSLMessages.NAMED_CHANNELS_UNSUPPORTED_HERE;
 
 import java.util.ArrayList;
 import java.util.Deque;
@@ -25,8 +25,6 @@ import java.util.HashMap;
 import java.util.LinkedList;
 import java.util.List;
 import java.util.Map;
-
-import com.google.common.collect.Iterables;
 
 import org.springframework.data.repository.CrudRepository;
 import org.springframework.util.Assert;
@@ -51,6 +49,8 @@ import org.springframework.xd.module.core.CompositeModule;
 import org.springframework.xd.module.options.ModuleOptionsMetadata;
 import org.springframework.xd.module.options.ModuleOptionsMetadataResolver;
 import org.springframework.xd.store.AbstractRepository;
+
+import com.google.common.collect.Iterables;
 
 /**
  * Parser to convert a DSL string for a stream into a list of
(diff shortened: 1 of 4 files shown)


spring-xd-dirt/src/main/java/org/springframework/xd/dirt/stream/XDStreamParser.java
```diff
@@ -177,19 +181,44 @@ public XDStreamParser(ModuleRegistry moduleRegistry,
...
+			try {
+				moduleType = determineType(builder, builders.size() - 1, parsingContext);
+			}
+			catch (NoSuchModuleException nsme) {
+				if (errorAccumulator != null) {
+					errorAccumulator.add(nsme);
+					// 'processor' below effectively indicates 'unknown' - a caller passing an
+					// error accumulator should be aware that this can happen (the accumulator
+					// will contain the exception that indicates it did)
+					moduleType = ModuleType.processor;
```
aclement: This seems the simplest option instead of adding lots of null handling throughout the following methods that get called.

spring-xd-dirt/src/main/java/org/springframework/xd/dirt/stream/XDStreamParser.java
```diff
@@ -354,20 +385,27 @@ public DocumentParseResult parse(String[] document) {
...
 					String streamName = stream.getStreamName();
+					if (streamName == null) {
+						// Give the stream a placeholder name as none was supplied. Someone
+						// processing the parse result should know this can happen.
+						streamName = "UNKNOWN_" + line;
+					}
+					List<Exception> errorAccumulator = new ArrayList<Exception>();
 					List<ModuleDescriptor> moduleDescriptors = delegate.buildModuleDescriptors(streamName,
-							nameAndDefinitionPair, ParsingContext.stream, stream);
+							nameAndDefinitionPair, ParsingContext.partial_stream, stream, errorAccumulator);
```
ilayaperumalg: Since the ParsingContext is now `partial_stream`, there won't be any validation on the module options (Since `bindAndValidate` is false for `partial_stream`). This would result in success response for all the definitions with modules that have invalid module options.

aclement: You have a keen eye for spotting that Ilaya :) I found the lack of positional info in those exceptions related to invalid options meant I couldn't really do a whole lot with them on the client at the moment. So I decided to go with partial_stream - it means the option validation isn't occurring but puts us in the same position as we are now (we aren't losing anything because we have never had option validation). Basically this PR was about doing the minimum to get us to something that works like it does today against our old endpoint in patched XD - we can take then take it forward from there after 1.2 is out and maybe enhance the exceptions with positions (the option validation one and the no such module one).

spring-xd-dirt/src/main/java/org/springframework/xd/dirt/stream/XDStreamParser.java
```diff
@@ -354,20 +385,27 @@ public DocumentParseResult parse(String[] document) {
...
 					BaseDefinition streamDefinition = new StreamDefinition(streamName, nameAndDefinitionPair);
 					transientRepository.save(streamDefinition);
-					result.success(moduleDescriptors);
+					result.success(moduleDescriptors, errorAccumulator);
```
ilayaperumalg: If we accumulate the errors but still save them as `success` would mean we would have the module listed in the success response as well. We will also have a error response for the error that is accumulated. Is that what we want?

aclement: There are two kinds of problems that can occur when parsing a definition. A syntax error or a validation error. On a syntax error I'm not really expecting a structure back.  On a validation error there is really nothing wrong with the parsed structure and it can be used by the client to graphically show the stream and then we can highlight the pieces of the stream with any validation errors. If all the get is the validation errors we can't show the user any kind of up to date graph. This becomes a shame when a user is typing because we want to reduce the times we can't show a graph.  As a user starts typing his stream 'mail', when he presses 'm' we can immediately show a node in the graph marked with 'not a known module', as they continue typing to spell 'mail' the graph constantly updates and eventually is shown as a valid mail node with no validation errors. Suppose they then start extending it 'mail | tran' at this point, there is a validation error on 'tran' (they haven't finished typing transform') but unless you give me the parsed structure, should I blank out the graph? the 'mail' node would then disappear when we know that is OK.  Hope you get what I'm describing.

ilayaperumalg: ok, thank you for clarifying. Now I understand we only handle `NoSuchModuleException` and the syntax error. The module option validation is not in place as we need positional info for that.

ilayaperumalg:
LGTM, merging.

ilayaperumalg:
Squashed, rebased and merged as 66a2e51034014a0a85d5ec66bc4150f3bd5de176

PR #1636, opened 2015-05-20
[Sprint 49] XD-3076 Add JavaMail properties to mail source

ilayaperumalg:
- Import JavaMail properties from a centralized config file.
This would also help adding any more properties and making it easier to work with.
 - Update SSL configs on JavaMail properties


ilayaperumalg:
Added JavaMail properties into ${xd.config.home} so that maintaining the properties would be easier.

modules/source/mail/config/mail.xml
```diff
@@ -34,7 +38,7 @@
 			store-uri="${protocol}://${username:}:${password:}@${host}:${port}/${folder}"
```
ilayaperumalg: For some reason I couldn't get the `imap-idle-channel-adapter working when I set --usePolling=false (which is by default false). The Polling mail source works.

markpollack: Why not put the properties inside the source xml file and expose the two Boolean false values as config properties of the source?

ilayaperumalg: I noticed there are few more JavaMail properties that user would like to add based on the need. In that case, they would need to edit the module xml to add new properties. I thought having them in config home could help. Also, once we have mail sink module updated for the same, we can add `SMTP` config there as well.

markpollack: We should assume that users won't have access to edit the file and also different stream instances may want to use different properties.  I guess the location isn't as important - if we want to reduce duplication for shared config.  I guess we have a few examples of that already.  Adding module options for the current Boolean props and allowing to pass in additional props should be added.

markpollack: Also, shouldn't thesr properties only apply when the mode is imaps?  I thought perhaps a profile based off the protocol would be the mechanism to apply the props

ilayaperumalg: > We should assume that users won't have access to edit the file and also different stream instances may want to use different properties. I guess the location isn't as important - if we want to reduce duplication for shared config.

I agree. We can add these properties into module XML and drive them based on profile.
If someone wants to add any more JavaMail properties, then can we still assume the module XML is the place to add them?

markpollack: No since that means file access as well.  I believe we have modules already that accept a spring resource location for additional properties.  Not sure if we ever supported something inline to specify properties.

ilayaperumalg:
Updated the PR to have `--properties` and `--propertiesFile` as options to inject JavaMail properties.

ilayaperumalg:
@markpollack Updated the PR that sets default JavaMail properties if the protocol is set to use `imaps`. Also, these properties can be overridden by properties from any of the `--properties` or `--propertiesFile` options.

markpollack:
cool, this looks much better, will review today.b

markpollack:
LGTM, merged as 6fc4425606e2be1eed4e467941fb1da9d0417857

tx!

PR #1637, opened 2015-05-20
[Sprint: 49] XD-3086 Fix random Spark streaming test failure

ilayaperumalg:
- The `testTapSparkProcessor` has the test that checks the contents at the output of spark
streaming word count processor. It turns out that the order in which these messages are
processed are not always in order. Hence, remvoed the content verification and added
the counter that verifies the total number of mesages after spark computation.

### DIFF
--- a/spring-xd-spark-streaming-tests/src/test/java/org/springframework/xd/spark/streaming/AbstractSparkStreamingTests.java
+++ b/spring-xd-spark-streaming-tests/src/test/java/org/springframework/xd/spark/streaming/AbstractSparkStreamingTests.java
@@ -20,13 +20,14 @@ import static org.hamcrest.core.IsEqual.equalTo;
 import static org.hamcrest.core.StringContains.containsString;
 import static org.junit.Assert.assertThat;
 import static org.springframework.xd.shell.command.fixtures.XDMatchers.eventually;
+import static org.springframework.xd.shell.command.fixtures.XDMatchers.exists;
 import static org.springframework.xd.shell.command.fixtures.XDMatchers.fileContent;
 import static org.springframework.xd.shell.command.fixtures.XDMatchers.hasContentsThat;
+import static org.springframework.xd.shell.command.fixtures.XDMatchers.hasValue;
 
 import java.io.File;
 import java.util.Random;
 
-import org.hamcrest.core.StringContains;
 import org.junit.After;
 import org.junit.Before;
 import org.junit.Rule;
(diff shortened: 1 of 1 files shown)


garyrussell:
Hi @ilayaperumalg I am now getting...

    java.lang.NullPointerException
	at org.springframework.xd.shell.command.fixtures.MetricExistsMatcher.matches(MetricExistsMatcher.java:39)
	at org.springframework.xd.test.fixtures.EventuallyMatcher.matches(EventuallyMatcher.java:59)
	at org.hamcrest.DiagnosingMatcher.matches(DiagnosingMatcher.java:12)
	at org.hamcrest.MatcherAssert.assertThat(MatcherAssert.java:12)
	at org.junit.Assert.assertThat(Assert.java:956)
	at org.junit.Assert.assertThat(Assert.java:923)
	at org.springframework.xd.spark.streaming.AbstractSparkStreamingTests.testTapSparkProcessor(AbstractSparkStreamingTests.java:168)

for both the Local and Rabbit tests.

mbogoevici:
Seems like a race condition here:

https://github.com/spring-projects/spring-xd/blob/master/spring-xd-shell/src/test/java/org/springframework/xd/shell/command/fixtures/MetricExistsMatcher.java#L39

Most likely the counter doesn't exist yet and the `Table` instance is null.

mbogoevici:
Rebased after https://github.com/spring-projects/spring-xd/commit/34dcdcac5af73e79160fe3beb29f481a4f70ad82

Now getting:

````
java.lang.AssertionError: 
Expected: an existing metric, trying at most 20 times
     but: failed after 20*100=2000ms:
counter named 'random98413' did not exist
````

One possible reason is that the test sends data before the tap is fully initialized, and therefore the processing doesn't happen.

mbogoevici:
Ok so it turns out that the test requires Redis for analytics, and therefore the counters weren't created at all. Will move to in-memory analytics.

ilayaperumalg:
This specific test started requiring the analytics repo after the fix in this PR that checks for the counter. I think the original issue was related to message ordering. So, moving to verify counter, fixes on MetricsMatcher (the null check) and finally to the in-memory analytic repo helped :-)

mbogoevici:
@ilayaperumalg Right :)

mbogoevici:
Changed to in-memory, rebased, merged https://github.com/spring-projects/spring-xd/commit/44b9920fd4020a162ed8cdeda4d812dd63e0ffe9

garyrussell:
Thanks; confirmed this worked on my old MacBook Air.

Thankfully, I'm back on my MacBook Pro now :smile:

PR #1647, opened 2015-05-21
[Sprint: 49] XD-3092: Synchronous deployment/undeployments

pperalta:
REST API controller now blocks until deployment/undeployment requests have been fulfilled by the admin supervisor.

### DIFF
--- a/spring-xd-dirt/src/main/java/org/springframework/xd/dirt/server/admin/deployment/zk/DeploymentMessageConsumer.java
+++ b/spring-xd-dirt/src/main/java/org/springframework/xd/dirt/server/admin/deployment/zk/DeploymentMessageConsumer.java
@@ -18,13 +18,14 @@ package org.springframework.xd.dirt.server.admin.deployment.zk;
 
 import java.util.Collections;
 
-import org.slf4j.Logger;
-import org.slf4j.LoggerFactory;
 import org.apache.curator.framework.CuratorFramework;
 import org.apache.curator.framework.recipes.queue.QueueConsumer;
 import org.apache.curator.framework.state.ConnectionState;
+import org.slf4j.Logger;
+import org.slf4j.LoggerFactory;
 
 import org.springframework.beans.factory.annotation.Autowired;
+import org.springframework.util.StringUtils;
 import org.springframework.xd.dirt.core.ResourceDeployer;
 import org.springframework.xd.dirt.server.admin.deployment.DeploymentAction;
 import org.springframework.xd.dirt.server.admin.deployment.DeploymentMessage;
@@ -32,6 +33,9 @@ import org.springframework.xd.dirt.stream.JobDefinition;
 import org.springframework.xd.dirt.stream.JobDeployer;
 import org.springframework.xd.dirt.stream.StreamDefinition;
 import org.springframework.xd.dirt.stream.StreamDeployer;
+import org.springframework.xd.dirt.zookeeper.Paths;
+import org.springframework.xd.dirt.zookeeper.ZooKeeperConnection;
+import org.springframework.xd.dirt.zookeeper.ZooKeeperUtils;
 
 /**
  * Consumer for {@link org.springframework.xd.dirt.server.admin.deployment.DeploymentMessage}
@@ -50,7 +54,10 @@ public class DeploymentMessageConsumer implements QueueConsumer<DeploymentMessag
 	@Autowired
 	private JobDeployer jobDeployer;
 
-	// for testing only
+	@Autowired
+	private ZooKeeperConnection zkConnection;
+
+	// todo: for testing only; this will be removed eventually
 	public void consumeMessage(DeploymentMessage message, StreamDeployer streamDeployer, JobDeployer jobDeployer) throws Exception {
 		this.streamDeployer = streamDeployer;
 		this.jobDeployer = jobDeployer;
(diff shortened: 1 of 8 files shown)


spring-xd-dirt/src/main/java/org/springframework/xd/dirt/server/admin/deployment/zk/ZKDeploymentMessagePublisher.java
```diff
@@ -33,21 +83,164 @@
...
+	 * Return the amount of time in milliseconds to wait for the message
+	 * consumer to process the deployment message. This value is 3 times
+	 * the value of the configured module deployment timeout.
 	 *
-	 * @param deploymentMessage the deployment message
+	 * @see #deploymentTimeout
+	 * @return timeout for deployment message processing
 	 */
+	private long getTimeout() {
+		return deploymentTimeout * 3;
```
ilayaperumalg: I understand why you have chosen a longer timeout than the `deploymentTimeout`. In case when we deploy hundreds or thousands of modules (as many number of module counts), then we might timeout too soon. This also raises one other question if we really want the client to wait for that longer time. But it might makes sense for automated deployments. We may want to have a better strategy to decide the timeout here I believe.

pperalta: Yes the 3x was an arbitrary choice. In theory I suppose we could examine/parse the deployment to determine the number of modules and use that count, but I didn't want to parse just for that. Maybe there's some other way to determine the number of modules to be deployed?

ilayaperumalg: We can probably assign the timeout based on the `DeploymentAction`. For the `create`, `createAndDeploy` and `destroy` (without **undeploy**) actions, we can have a shorter timeout as they don't involve any deployment actions. If the `DeploymentAction` is deploy then we can read the count from `deploymentProperties`. If the  action is `undeploy` then I think we would need to know only from ZK.

ilayaperumalg: Since  we don't support `deploymentProperties` for the `createAndDeploy` action, we can have relatively shorter timeout for this action even though it does the deployment.

pperalta: Looking at the properties and deployment action is a good idea. I can take a look at this, or if you want to push a suggestion I'm happy to look at that too.

ilayaperumalg: sure, I can try this and push the suggestion.

ilayaperumalg: While I am at it, I see a caveat when using deployment properties to determine the module count.  If the count is specified as `module.*.count` then no way we know ho many number of modules were there. One rough solution would be to get the number of pipes from the definition string and calculate the number of modules. This would work even if we have composed module in the definition string since we are concerned about the number of `ModuleDescriptors` that will be used for the deployment. What do you think?

ilayaperumalg: @pperalta Added some logic that calculates timeout based on the modules in context. There are still a few items to address. Before that, I wanted to make sure this is in the right direction: https://github.com/ilayaperumalg/spring-xd/commit/40b0f43979d7a96daf37fad9dd4fba707759ce78

spring-xd-dirt/src/main/java/org/springframework/xd/dirt/server/admin/deployment/zk/ZKDeploymentMessagePublisher.java
```diff
@@ -33,21 +83,164 @@
...
+		}
+		catch (RuntimeException e) {
+			throw e;
 		}
 		catch (Exception e) {
 			throw new RuntimeException(e);
 		}
+		finally {
+			try {
+				client.delete().forPath(resultPath);
```
ilayaperumalg: This should be deleting the children path as well. Will fix while merging.

spring-xd-dirt/src/main/java/org/springframework/xd/dirt/server/admin/deployment/zk/DeploymentMessageConsumer.java
```diff
@@ -79,46 +86,84 @@ public void consumeMessage(DeploymentMessage deploymentMessage) throws Exception
...
+	 *
+	 * @param message    deployment message
+	 * @param errorDesc  error description if an error occurred processing
+	 *                   the request
+	 */
+	private void writeResponse(DeploymentMessage message, String errorDesc) {
+		try {
+			String requestId = message.getRequestId();
+			if (StringUtils.hasText(requestId)) {
+				logger.info("Processed deployment request " + requestId);
```
ilayaperumalg: I believe we can log this as `debug`. Also, the requestId doesn't need to exposed in the `info` level logs.

spring-xd-dirt/src/main/java/org/springframework/xd/dirt/server/admin/deployment/zk/DeploymentMessageConsumer.java
```diff
@@ -79,46 +86,84 @@ public void consumeMessage(DeploymentMessage deploymentMessage) throws Exception
...
+				String resultPath = Paths.build(Paths.DEPLOYMENTS, Paths.RESPONSES, requestId,
+						errorDesc == null
+								? ZKDeploymentMessagePublisher.SUCCESS
+								: ZKDeploymentMessagePublisher.ERROR);
+				logger.info("creating result path {}", resultPath);
```
ilayaperumalg: Same here, I think we can change the log level to `debug` here as well.

ilayaperumalg:
LGTM, rebased and merged as 9f2294ffee048bba87d76eb915ca77051e69a33e. Also, pushed the changes ea0f15e15b4ff641a10bb8f95aba0de6ce96617d that include:
 - Make sure to delete the children path when deleting the response node
 - Change the log level to `debug` at DeploymentMessageConsumer
 - Fix StreamPartitionTests to include `redis` and `rabbit` rule

spring-xd-dirt/src/main/java/org/springframework/xd/dirt/server/admin/deployment/zk/DeploymentMessageConsumer.java
pperalta: Agreed on making this debug; hopefully you kept the request id anyway - this can come in useful when debugging.

ilayaperumalg: Yeah, right! I kept it as is :-)

PR #1649, opened 2015-05-21
[Sprint: 50] XD-2717 Add nameExpression and directoryExpression properties to File Sink

fmarchand:

### DIFF
--- a/spring-xd-shell/src/test/java/org/springframework/xd/shell/command/FileSourceAndFileSinkTests.java
+++ b/spring-xd-shell/src/test/java/org/springframework/xd/shell/command/FileSourceAndFileSinkTests.java
@@ -16,6 +16,7 @@
 
 package org.springframework.xd.shell.command;
 
+import static org.hamcrest.CoreMatchers.is;
 import static org.hamcrest.Matchers.equalTo;
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertThat;
@@ -38,6 +39,7 @@ import org.springframework.xd.test.fixtures.FileSource;
  * 
  * @author Eric Bottard
  * @author David Turanski
+ * @author Franck Marchand
  */
 public class FileSourceAndFileSinkTests extends AbstractStreamIntegrationTest {
 
@@ -67,6 +69,11 @@ public class FileSourceAndFileSinkTests extends AbstractStreamIntegrationTest {
 	@Test
 	public void testCustomSuffix() throws Exception {
 		testFileSinkSource("." + TEXT_SINKFILE_SUFFIX, "--suffix='" + TEXT_SINKFILE_SUFFIX + "'");
+	}
+
+	@Test
+	public void testNameAndDirectoryExpression() throws Exception {
+		testFileSinkSource("--nameExpression=payload.trim()", "dir-", "-suffix");
 	}
 
 	@Test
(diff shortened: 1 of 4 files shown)


fmarchand:
I tried something using this PR.
Input payload is a json containing a field `date` :

   `... | file --nameExpression='#jsonPath(payload,''$.date'')'"`

Spring XD tells me that method jsonPath is null ... 

if I try :

`... | log --expression='#jsonPath(payload,''$.date'')'"`

it works ...

I don't see why ! Could someone help me to fix that. Otherwise this PR is incomplete I think.

fmarchand:
I'm adding dirExpression to this module don't merge now please I will have finished in 30 minutes ...

markpollack:
Can one option cover the full path as in the hdfs sink?

fmarchand:
I'm gonna check the hdfs sink to see if I can do that. It's a good idea ! That would be easier to use.

fmarchand:
I don't see the option you're talking about in HDFS sink ... I see `directory` but it's not a SpEL expression.

markpollack:
Look at http://docs.spring.io/spring-xd/docs/current/reference/html/#hadoop-hdfs and --partitionPath

markpollack:
Maybe it is in fact two distinct options after all...

fmarchand:
Oh .... I see ... because I use the same options from Spring Integration `file:outbound-adapter` I  kept the two distinct options. I did this pull request because I need it in one of my projects and keeping the two options is easier in my case and the stream is more readable when you display it with `stream list`. Don't you think ?

markpollack:
Yea, you are right.  Mapping to the underlying options in SI is better.  Sorry for the noise

fmarchand:
No pb :) That's always a pleasure to talk with you pivotal guys :) I always learn a lot from you. I'm gonna finish my PR. Hope it will be ok.

fmarchand:
Job done.

ghillert:
@fmarchand Thanks for the PR!! Could you change `DirExpression` to `DirectoryExpression` (properties, the module options etc.). I think this will be more consistent with the underlying Spring Integration.

fmarchand:
I've done the 'directory' thing

ghillert:
@fmarchand Sorry for my earlier confusion. I merged the PR. Only at second looks I realized that we were already using `dir`. In that context dirExpression actually makes sense. Thanks for the contribution!!!!

fmarchand:
You're welcome ! NP for the confusion. That happens :)

spring-xd-shell/src/test/java/org/springframework/xd/shell/command/FileSourceAndFileSinkTests.java
```diff
@@ -113,4 +120,38 @@ private void testFileSinkSource(String sinkSuffix, String sinkParam) throws Exce
...
+		try {
+			FileCopyUtils.copy("hello", new FileWriter(in));
+			String directoryExpression = " --directoryExpression='''/tmp/'' + ''" + dirPrefix + "'' + payload.trim() + ''"
+					+ dirSuffix + "'''";
+			stream().create(streamName, "in: file --outputType=text/plain | out: file " + nameExpressionParam +
+					directoryExpression);
+			Thread.sleep(1000);
+			String actual = FileCopyUtils.copyToString(new FileReader(out));
+			assertEquals("hello", actual.trim());
+			assertEquals(out.exists(), is(true));
```
garyrussell: @ghillert I am getting a failure here on master; this should be `assertThat()` (or simply `assertTrue`).

`true` can never equal the hamcrest matcher `Is`.

spring-xd-shell/src/test/java/org/springframework/xd/shell/command/FileSourceAndFileSinkTests.java
```diff
@@ -113,4 +120,38 @@ private void testFileSinkSource(String sinkSuffix, String sinkParam) throws Exce
...
+		File outDir = new File("/tmp/" + dirPrefix + "hello" + dirSuffix);
+		File out = new File(outDir, "hello");
+
+		try {
+			FileCopyUtils.copy("hello", new FileWriter(in));
+			String directoryExpression = " --directoryExpression='''/tmp/'' + ''" + dirPrefix + "'' + payload.trim() + ''"
+					+ dirSuffix + "'''";
+			stream().create(streamName, "in: file --outputType=text/plain | out: file " + nameExpressionParam +
+					directoryExpression);
+			Thread.sleep(1000);
```
garyrussell: Also, arbitrary sleeps like this are unreliable on the CI servers and can extend the overall test times; it's better to intelligently wait for some condition (e.g. loop 100 times for 100ms until the condition is true).

It provides a fast exit when the condition is generally true faster than the arbitrary sleep, while allowing for a longer time when running on a heavily loaded server.

spring-xd-shell/src/test/java/org/springframework/xd/shell/command/FileSourceAndFileSinkTests.java
dturanski: This test should use the same pattern as others in here, i.e., https://github.com/spring-projects/spring-xd/blob/9356b584b1765304e16a1c2db5450e28aebab38b/spring-xd-shell/src/test/java/org/springframework/xd/shell/command/FileSourceAndFileSinkTests.java#L87

fmarchand: @garyrussel : yes it should be assertThat ... and that's what I'm used to use ... my mistake ! But what bothers me the most : why I did not get exception when I ran the gradlew dist ? I missed something ...

PR #1651, opened 2015-05-22
Try increasing timeout ? local build fails to fail so have to give a …

smaldini:
…try to Bamboo


markpollack:
Applied changes in 46e2b385f3f7dcf37397c9cd728f65833595e0c0 as 4396aeb0c760a7a19d91ff3d2a82c25101e82e94 as the minimal changes to resolve failing test issue.  Changes wrt to orderly shutdown should be done in a separate PR.  

Note, also added additional exception message in `BroadcasterMessageHandler` to be in sync with behavior in `MultipleBroadcasterMessageHandler`

PR #1654, opened 2015-05-22
[Sprint 50] XD-3098 Message bus optimizations

mbogoevici:
- Optimize mimetype determination: avoid parsing for each incoming mesage
- Optmize embed headers: embed without creating new `Message` instances
- Consumer improvements: preserve existing functionality while minimizing `Message` instances creation
- Kafka Message Bus improvements : add 'mode' for header- and synchronization-less bus transport;
- Add wait time parameter for Kafka client
- Fix configurability of buffered bytes for Kafka message bus producer
- Add tests

### DIFF
--- a/config/servers.yml
+++ b/config/servers.yml
@@ -70,6 +70,7 @@
 #    kafka:
 #      brokers:                                 localhost:9092
 #      zkAddress:                               localhost:2181
+#      mode:                                    embeddedHeaders
 #      numOfKafkaPartitionsForCountEqualsZero:  10
 #      socketBufferSize:                        2097152
 #      offsetStoreTopic:                        SpringXdOffsets
@@ -84,7 +85,6 @@
 #      offsetUpdateCount:                       0
 #      offsetUpdateShutdownTimeout:             2000
 #      default:
-#        batchingEnabled:           false
 #        batchSize:                 16384
 #        batchTimeout:              0
 #        replicationFactor:         1
@@ -92,7 +92,8 @@
 #        requiredAcks:              1
 #        compressionCodec:          none
 #        autoCommitEnabled:         true
-#        queueSize:                 1000
+#        queueSize:                 1024
+#        maxWait:                   100
 #        fetchSize:                 1048576
 #        minPartitionCount:         1
 
--- a/spring-xd-dirt/src/test/java/org/springframework/xd/dirt/integration/bus/kafka/RawKafkaPartitionTestSupport.java
+++ b/spring-xd-dirt/src/test/java/org/springframework/xd/dirt/integration/bus/kafka/RawKafkaPartitionTestSupport.java
@@ -1,0 +1,40 @@
+/*
+ * Copyright 2014 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.xd.dirt.integration.bus.kafka;
+
+import org.springframework.messaging.Message;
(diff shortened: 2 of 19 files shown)


mbogoevici:
Documenting message bus configuration changes to be addressed as part of https://jira.spring.io/browse/XD-3016

config/servers.yml
```diff
@@ -92,7 +92,8 @@
 #        requiredAcks:              1
 #        compressionCodec:          none
 #        autoCommitEnabled:         true
-#        queueSize:                 1000
+#        queueSize:                 1024
```
mbogoevici: Document that it must be a power of 2

spring-xd-dirt/src/test/java/org/springframework/xd/dirt/integration/bus/kafka/RawKafkaPartitionTestSupport.java
```diff
@@ -0,0 +1,40 @@
...
+
+package org.springframework.xd.dirt.integration.bus.kafka;
+
+import org.springframework.messaging.Message;
+import org.springframework.xd.dirt.integration.bus.PartitionKeyExtractorStrategy;
+import org.springframework.xd.dirt.integration.bus.PartitionSelectorStrategy;
+
+
+/**
+ *
```
dturanski: Add a description

spring-xd-messagebus-spi/src/main/java/org/springframework/xd/dirt/integration/bus/MessageBusSupport.java
```diff
@@ -556,28 +556,20 @@ protected void stopBindings() {
...
-			if (originalContentType != null) {
-				messageValues.put(XdHeaders.XD_ORIGINAL_CONTENT_TYPE, originalContentType);
-			}
-			return messageValues;
-		}
-		else {
-			throw new IllegalArgumentException("'to' can only be 'ALL' or 'APPLICATION_OCTET_STREAM'");
+
+		//Pass content type as String since some transport adapters will exclude CONTENT_TYPE Header otherwise
+		Object contentType = JavaClassMimeTypeConversion.mimeTypeFromObject(originalPayload).toString();
```
dturanski: @markpollack - we talked about having this method return a String since we always convert the result to a string anyway.

spring-xd-messagebus-spi/src/main/java/org/springframework/xd/dirt/integration/bus/MessageBusSupport.java
```diff
@@ -1039,27 +1035,37 @@ public T lookupSharedChannel(String name) {
 	 */
 	abstract static class JavaClassMimeTypeConversion {
 
+		public static final MimeType APPLICATION_OCTET_STREAM_MIME_TYPE = MimeType.valueOf(APPLICATION_OCTET_STREAM_VALUE);
```
dturanski: I think we can accomplish this with only String values and possibly eliminate the need for the cache.

spring-xd-messagebus-spi/src/main/java/org/springframework/xd/dirt/integration/bus/MessageBusSupport.java
markpollack: Doesn't seem to hurt ATM.  There is a cascading effect on tests to then create the MimeType to pass into `classNameFromMimeType` and that method with it's current signature is also called via a chain of invocations from deserializePayload.  Given the # of changes in the PR and that it doesn't seem to impact perf, I suggest a 2nd iteration for anymore cleanup.

spring-xd-messagebus-spi/src/main/java/org/springframework/xd/dirt/integration/bus/MessageBusSupport.java
markpollack: There isn't a cache for invocations of `mimeTypeFromObject` that references APPLICATION_OCTET_STREAM_MIME_TYPE but there is a cache for invocations of `classNameFromMimeType.

In our perf tests we are relying on the fact that the object is an instance of byte[], so calls to MimeType.valueOf are bypassed.  It maybe relevant if the payload isn't of type byte[] or String since that is what leads to a call that does `MimeType.valueOf`.  

If we need this change, it should be part of another issue.

dturanski:
OK then. LGTM

dturanski:
rebased and merged

PR #1655, opened 2015-05-25
[Sprint: 50] XD-3090 Use Job Wait times for JdbcHdfs Acceptance Tests.

cppwfs:
Use Job Wait times for JdbcHdfs Acceptance Tests. This should resolve Acceptance tests failures.


mminella:
Merged.

PR #1657, opened 2015-05-26
[Sprint: 50] XD-3079 Hadoop kerberos config not used in jobs

jvalkeal:
- In XD-3079 user complained that kerberos tickets are
  not renewed. Due to missing kerberos config, these jobs
  doesn't even work with kerberized hdfs if job is only
  one accessing hdfs. Because hadoop sec config is a system
  wide setting in hadoop classes, jobs can access kerberized
  hdfs if i.e. hdfs sink first did a login.
- Having this proper sec config should allow jobs to work
  beyond kerberos tgt expire and renew times.
- Adding same sec config from hdfs sink to hdfsmongodb,
  hdfsjdbc, jdbchdfs, ftphdfs and filepollhdfs jobs.
- In hdfs/hdfs-dataset sinks we support security settings in both
  hadoop.properties and servers.yml, so property-placeholder
  is needed solely for hadoop.properties.
- In tests remove property-placeholder bean in favor of registering
  needed properties in context for everything to
  work with property-placeholder beans in job xml's.


trisberg:
Merged via commit e2deca6dbafb0e35923cd4e509df25419b71302d

PR #1660, opened 2015-05-26
XD-3103 Remove use of ResolvableType from SF and use custom helper class

markpollack:
- Use of SoftReference in ResolvableType was introducing a NPE
- Add Messaging exception in the case of mismatched output type for processor


markpollack:
Based on work in 46e2b385f3f7dcf37397c9cd728f65833595e0c0

PR #1661, opened 2015-05-26
[Sprint: 50] XD-3089: Documented incremental imports via jdbchdfs

mminella:


src/docs/asciidoc/Batch-Jobs.asciidoc
```diff
@@ -509,6 +510,25 @@ xd:> job create partitionedJob --definition "jdbchdfs --columns='id,col1,col2' -
...
+If you want to specify the value for the `checkColumn`, you can pass the override value in as a `JobParameter` named `overrideCheckColumnValue` as shown below:
+
+----
+xd:> job launch incrementalImportJob --params {"overrideCheckColumnValue" : 2}
+----
+
+There are two things to keep in mind when using incremental imports with this job:
+
+* When using incremental imports, the `sql` option is not available.  Use `tableName` and `columns` instead.
+* If an import fails, it must be rerun to completion before running the next import.  Without this, inconsistent data may result.
```
trisberg: Since HDFS isn't a transactional resource, should we add a note to check the exported data for completeness before restarting the job? Should we add the restartable option?

mminella: Good catch on the transactional piece.  What do you mean by the "restartable option"?

trisberg: in the example above add --restartable=true

mminella: Gotcha.

trisberg:
Merged via commit f2b842c3f8c87898dcc3f3f0cf197b5e07da97ac

PR #1662, opened 2015-05-26
[Sprint 50] XD-3107 build fixes

dturanski:
This fixes gradle 'dist' and 'distZip'. Also removes some extraneous output messages from the build. Prior to this, 'dist' appears to work correctly on CI, but doesn't do any work (thinks it's up to date) on the command line.  'distZip' resulted in 'startScripts' errors, but once those were corrected, created build/distributions/*-dist.zip for every project. The net effect was that spring-xd-dist.zip and spring-xd-yarn-dist.zip, created by 'dist', were huge due to including the module artifacts created by running 'distZip'.  So, this disables distZip for all subprojects and also disables startScripts.  Note that these issues are not observed in the CI Publish build which configures the Gradle artifactory plugin.

Added some logic to exclude hadoopDistros, 'documentation-toolchain' and the rootProject from publications('mavenCustom'). Also added logback.xml to the document-toolchain classpath to supress DEBUG and INFO log messages.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -182,6 +182,7 @@ configure(javaProjects) { subproject ->
 	dependencies {
 		versionManagement files(rootProject.file('dependencies.properties'))
 	}
+	startScripts.enabled = false
 
 	bootRepackage {
 		enabled = false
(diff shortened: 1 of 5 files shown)


markpollack:
squashed and merged as b7068504a63f2439368ad34b7715fc8e4dc4a6fb

PR #1663, opened 2015-05-27
[For Review] Refresh reactor module

smaldini:
A fresh perspective on Reactor Module,

Using @EnableReactorModule and source/sink/processor contract.


markpollack:
will revisit this at a later date.

PR #1665, opened 2015-05-27
XD-2294 Add header enricher module

fmarchand:


dturanski:
See https://github.com/spring-projects/spring-xd-modules/pull/18.  I would like to close this and go with that since this only supports a single header.

PR #1666, opened 2015-05-28
[Sprint: 50] XD-3112 Hide passwords used as module properties in streams

ghillert:
* Mask passwords when returning streams via the REST-API
* Add test

### DIFF
--- a/spring-xd-dirt/src/main/java/org/springframework/xd/dirt/rest/StreamsController.java
+++ b/spring-xd-dirt/src/main/java/org/springframework/xd/dirt/rest/StreamsController.java
@@ -16,6 +16,7 @@
 
 package org.springframework.xd.dirt.rest;
 
+import java.util.ArrayList;
 import java.util.List;
 import java.util.Map;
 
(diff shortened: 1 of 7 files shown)


spring-xd-dirt/src/main/java/org/springframework/xd/dirt/rest/StreamsController.java
```diff
@@ -67,7 +68,24 @@ public StreamsController(StreamDeployer streamDeployer) {
 	@ResponseStatus(HttpStatus.OK)
 	public PagedResources<StreamDefinitionResource> list(Pageable pageable,
 			PagedResourcesAssembler<StreamDefinition> assembler) {
-		return listValues(pageable, assembler);
+
+		PagedResources<StreamDefinitionResource> pagedResources = listValues(pageable, assembler);
```
ilayaperumalg: We could preferably do the masking of the `definition` at the `StreamDefinitionResourceAssembler` rather than doing it explicitly below.

ilayaperumalg:
We still need to mask the `module` properties. For example, when I run `runtime modules` I see the password content (if any) is displayed there.

ghillert:
Will fix.

ghillert:
@ilayaperumalg @sabbyanandan Added password masking for `/runtime/modules`. Also created https://jira.spring.io/browse/XD-3158 It would theoretically be nice if the module option type information was available somehow.

ilayaperumalg:
LGTM, merging.

ilayaperumalg:
Squashed, rebased and merged as 70866a5fed9c6abcdda33cb05e5ab7a09cfa40ad

PR #1677, opened 2015-06-01
XD-3130 Create BusCleanerUtils

ilayaperumalg:
- Move `buildRestTemplate` method from `BusUtils` to a new class `BusCleanerUtils`
 - Since `Spark-streaming` uses `BusUtils`, this change would make Spark streaming not
to depend on `httpClient` that is used by the bus cleaner util method to build rest template

### DIFF
--- a/spring-xd-dirt/src/main/java/org/springframework/xd/dirt/integration/bus/rabbit/RabbitBusCleaner.java
+++ b/spring-xd-dirt/src/main/java/org/springframework/xd/dirt/integration/bus/rabbit/RabbitBusCleaner.java
@@ -45,6 +45,7 @@ import org.apache.http.protocol.HttpContext;
 import org.springframework.web.client.RestTemplate;
 import org.springframework.web.util.UriComponentsBuilder;
 import org.springframework.xd.dirt.integration.bus.BusCleaner;
+import org.springframework.xd.dirt.integration.bus.BusCleanerUtils;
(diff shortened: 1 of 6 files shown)


garyrussell:
I don't think we should call it `BusCleanerUtils`.

The reason I moved it from `RabbitBusCleaner` is because it's now also used by the `LocalizedQueueMessageFactory`. So it's no longer just for "bus cleaning".

Perhaps `RabbitAdminUtils`?

garyrussell:
Otherwise; no issues.

ilayaperumalg:
+1 for RabbitManagementUtils, I also wonder why this is not in `spring-xd-messagebus-rabbit` but in `messagebus-spi`? I also see `RabbitAdminException` on the same package.

ilayaperumalg:
I see the reason why they exist in `spring-xd-messagebus-spi`. Pushed the change to rename the class.

garyrussell:
Right - the issue is really that, when the bus code was split out of dirt, the tests were not so we can't have dirt depend on bus implementations. We should revisit this in 1.3 (clean separation of bus implementations from dirt).

garyrussell:
Merged as 50e0f8298b3c6808316ead0a082eb34804ceb857

markpollack:
Merged in 1.2.0.RC1 branch as 958f8f57763a0cbf39d7048c7bbfe9781f8686a4

PR #1678, opened 2015-06-01
[Sprint: 50] XD-3131 Disable tap listener for spark streaming plugin

ilayaperumalg:
- This avoids the creation of the curator pathchildren cache thread for Taplistener when the
SparkStreamingPlugin is initialized

### DIFF
--- a/spring-xd-dirt/src/main/java/org/springframework/xd/dirt/plugins/spark/streaming/SparkStreamingPlugin.java
+++ b/spring-xd-dirt/src/main/java/org/springframework/xd/dirt/plugins/spark/streaming/SparkStreamingPlugin.java
@@ -107,8 +107,8 @@ public class SparkStreamingPlugin extends AbstractStreamPlugin {
 	private Map<Module, JavaStreamingContext> streamingContexts = new HashMap<>();
 
 	@Autowired
-	public SparkStreamingPlugin(MessageBus messageBus, ZooKeeperConnection zkConnection) {
-		super(messageBus, zkConnection);
+	public SparkStreamingPlugin(MessageBus messageBus) {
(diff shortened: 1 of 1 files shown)


spring-xd-dirt/src/main/java/org/springframework/xd/dirt/plugins/spark/streaming/SparkStreamingPlugin.java
```diff
@@ -107,8 +107,8 @@
 	private Map<Module, JavaStreamingContext> streamingContexts = new HashMap<>();
 
 	@Autowired
-	public SparkStreamingPlugin(MessageBus messageBus, ZooKeeperConnection zkConnection) {
-		super(messageBus, zkConnection);
+	public SparkStreamingPlugin(MessageBus messageBus) {
+		super(messageBus, null);
```
pperalta: I think there is a 1 parameter version of this constructor that you can use instead of passing `null`.

ilayaperumalg: yeah, fixing it.

ilayaperumalg:
Pushed the change.

pperalta:
Rebased, squashed & merged as 7d19adcf9302c2f58ec95c7ecacb70166387d88a

ilayaperumalg:
@pperalta This change needs to be cherry-picked and pushed to 1.1.x as well

pperalta:
Cherry-picked to 1.1.x as 1009a00764235d2a72682edc352d91f20d796ae8

ilayaperumalg:
@pperalta thank you!

PR #1684, opened 2015-06-03
[Sprint 50]XD-3127,XD-3140 Added Available Modules and Jobs page

dturanski:


ericbottard:
Need to change the gradle build that generates doc from options metadata, following the renaming from Batch-Jobs to Jobs.

dturanski:
https://github.com/spring-projects/spring-xd/pull/1684/files#diff-5f01d5115b576ef8df34e4a8de08e7d1R46

Actually, it wasn't a rename, Jobs was added. I'm not sure why git thinks it's a rename, both files are still there.

ericbottard:
my bad

trisberg:
This part looks dated:

==== Register a Job Module
 
 Register a *Job Module* with the _Module Registry_ by putting XML and/or jar files into the +$XD_HOME/modules/jobs+ directory.

Line 51 in Batch-Jobs.asciidoc
(didn't see a way to comment on a line in the source that's not part of the changes)

trisberg:
While we are changing this - it looks weird to me that we have Sources, Processors, Sinks and Jobs as top level chapters while "Counters and Gauges" is under the "Analytics" chapter. Should we move Sources, Processors, Sinks and Jobs to be sections within the "Modules" chapter?

dturanski:
I thought about that. I could add them as sections to the Available modules page. So the HTML TOC would look like

````
Available Modules
     Sources

     Sinks 

     Jobs
````

trisberg:
They can stay in the separate asciidoc files, just add another '=' to each heading to push them in one level. Looks better IMO.

dturanski:
I can try it. Currently we don't do that anywhere, so something else may break

trisberg:
On the other hand, this change would remove the individual modules from the ToC in the published docs since that only shows 2 levels.

dturanski:
Then I think we should promote C&G to a top level page.

trisberg:
LGTM now, going to merge this now, so any other changes can go in another PR

trisberg:
Merged as commit 892f4100542ef89f630bf56fbfb66c883389ac69

PR #1687, opened 2015-06-04
[Sprint 50]move counters and gauges to its own page

dturanski:
Improvement on previous doc changes


src/docs/asciidoc/Analytics.asciidoc
```diff
@@ -331,301 +331,11 @@ datasets$testset[,1:4][1,]
...
-Check the gauge:
-
-[source,bash]
-----
-xd:>rich-gauge display emcprice
-xd:>rich-gauge display vmwprice
-----
-
-==== Accessing Analytics Data over the RESTful API
+=== Accessing Analytics Data over the RESTful API
```
ericbottard: the REST API only pertains to counters & gauges, so one could argue it should be a child of said section. Maybe move it to the new file as well?

src/docs/asciidoc/Analytics.asciidoc
```diff
@@ -331,301 +331,11 @@ datasets$testset[,1:4][1,]
 [1] versicolor
 ----
 
-[[counters-and-gauges]]
 === Counters and Gauges
```
ericbottard: maybe we should slightly change this section title, to avoid autogenerated id issues

ericbottard:
Merged as de27c762de88a648f64c67fd08ab52ce63ac00bc

PR #1688, opened 2015-06-04
[Sprint: 50] XD-3141 Avoid ClasspathResource for module config files

ilayaperumalg:
- When the module configuration files (config/*.xml, config/*.properties) are loaded,
they are added as `ClasspathResource`. This doesn't work in case if the module was already
loaded once and the module definition is overridden by `module upload` command. This
requires server restart to re-initialize the classloader.
  - This PR suggests using `UrlResource` for all the `ClasspathResource` sources and thereby
any overridden module definitions are always re-loaded.

### DIFF
--- a/spring-xd-module/src/main/java/org/springframework/xd/module/core/ResourceConfiguredModule.java
+++ b/spring-xd-module/src/main/java/org/springframework/xd/module/core/ResourceConfiguredModule.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2014 the original author or authors.
+ * Copyright 2014-2015 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -16,7 +16,12 @@
 
 package org.springframework.xd.module.core;
 
+import java.io.IOException;
+
+import org.springframework.core.io.ClassPathResource;
 import org.springframework.core.io.Resource;
+import org.springframework.core.io.UrlResource;
+import org.springframework.util.Assert;
 import org.springframework.xd.module.ModuleDeploymentProperties;
 import org.springframework.xd.module.ModuleDescriptor;
 import org.springframework.xd.module.SimpleModuleDefinition;
@@ -29,6 +34,7 @@ import org.springframework.xd.module.options.ModuleUtils;
  *
  * @author David Turanski
  * @author Eric Bottard
+ * @author Ilayaperumal Gopinathan
  */
 public class ResourceConfiguredModule extends SimpleModule {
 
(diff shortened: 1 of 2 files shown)


ericbottard:
We should get rid of classpath: module loading entirely anyway, as currently, because
1. `$xd.home/config` is on the classpath
2. we support `classpath:/modules/` as a registry root

then module option override directories (such as `$xd.home/config/modules/sink/rabbit`) are mistaken as module definitions. I a user were to upload a custom 'foo' module and put defaults in `$xd.home/config/modules/sink/foo`, then he can't `--force` update it.

See https://jira.spring.io/browse/XD-2988

ilayaperumalg:
@ericbottard Since `module upload --force` will require restart of admin/container when the changes are in `ClassPathResource`,  I believe we need this fix at least for GA as we wouldn't plan to get rid of `classpath` loading entirely at least for GA. What do you think?

ericbottard:
I think we should consider dropping class path resource, given that it is causing the problem I described AND the one this PR is trying to cure. The only reason we have it still is for YARN if I'm not mistaken, but maybe @trisberg can come up with a different approach

trisberg:
Why would they put custom properties for foo in `$xd.home/config/modules/sink/foo`? Shouldn't that be packaged in the foo.jar? 

For YARN we are now using hdfs for xd.customModule.home

and for module config: xd.module.config.location: "file:./modules-config.zip/"

ericbottard:
`$xd.home/config/modules/sink/foo` is the location where the module _user_ can put defaults (as opposed to the module _author_). The reporter of https://jira.spring.io/browse/XD-2988 tried exactly that (even though he actually played the two roles)

trisberg:
I see, makes sense now. Any chance we could add hdfs support for xd.module.config.location?

ilayaperumalg:
For the module configurations, I don't think we need to drawback support for classpath resource. In some way, that makes sense. The reason is that, we don't want to have conflicts when there are existing modules that use the over-ridding configurations. Since the only place where we allow override is `module upload` by the option `--force`, I believe this PR would fix the underlying issue.

ilayaperumalg:
Removed the unused `ResourceConfigureModule` changes.

ilayaperumalg:
The **rootCause** for the problem is that the underlying module classloader has a weak reference to the jar that was loaded before. Anytime the module classloader is created **just for loading** the properties, then it needs to be closed. In other cases, the module undeploy would take care of closing this.

ilayaperumalg:
Pushed the change to fix the root cause.

ericbottard:
I created https://github.com/spring-projects/spring-xd/pull/1696 which IMHO addresses the real problem.

From a pure API contract point of view, we can't close the classloader as we don't know how the resource has been obtained and it might be used still.

ilayaperumalg:
> From a pure API contract point of view, we can't close the classloader as we don't know how the resource has been obtained and it might be used still.

At least in this case, we know how this resource is obtained and all we needed is the `properties`. Moreover, the classloader we create here is a throwaway one and the resources will `always` be re-loaded if it is a `ClassPathResource`.

Your suggestion on the other PR uses URL connection for the resource was similar to the original fix to return `URLResource` if the instance is of `ClassPathResource` though now it uses `URLConnection` for all the resources. If we choose this way, then we don't need to set the classloader for the resource when loading `properties` as we make URLConnection all the time. and, the caching there is never used as it only caches if the resource uses JNLP file.

I am fine with both ways, but requires some testing on the new PR :-)

ericbottard:
> Your suggestion on the other PR uses URL connection for the resource was similar to the original fix to return URLResource

not entirely. The `URLConnection` is a JDK API used anytime we access streams, while `URLResource` is our resource abstraction over a plain URL. `URLConnection` is used in all cases actually

ilayaperumalg:
@ericbottard `similar` != `equal` ;) what I meant there is that we retrieve the resource explicitly by URL rather than downloading from Classpath. So, we don't depend on the **classloader**

ericbottard:
Closing for now, as this PR moved more on the side of ClassLoader closing. Need to address this more globally, created https://jira.spring.io/browse/XD-3159 to that effect

PR #1691, opened 2015-06-05
[Sprint: 50] XD-3150 Fix filepollhdfs problem with multiple files

jvalkeal:
- Fixing regression where second file threw error
  because writer tried to flush a closed stream.
- Fix docs to use correct option for file sink
  in filepollhdfs job section
  "file --ref=true" vs. "file --mode=ref"


trisberg:
Merged via commit 0ff333f2ab087114192686d677df0fae006b6f8c

trisberg:
Merged into 1.1.x via commit ee7cee49179a403886a8e9865e73d0cc3e51a8da

PR #1692, opened 2015-06-05
[Sprint: 50] XD-3155 Use ClassLoaderUtils to load classes

philwebb:
Update MessageBusSupport to use ClassUtils.forName rather than
Class.forName to load classes. This ensure that classes are always
loaded from the context classloader.


markpollack:
Thanks Phil!

mbogoevici:
Minor changes to avoid parsing the contentType when it is text. Rebased, merged. https://github.com/spring-projects/spring-xd/commit/63995b23e517360c87e59a0714ddb541458067e7

PR #1694, opened 2015-06-08
[Sprint 50 ]XD-2996 Align transport native partitioning with Spring XD partitioning

mbogoevici:
- Eliminate the `partitionCount` property, which is currently just a copy of `count`, as well as its calculation;
- Propagate `nextModuleCount` and `nextModuleConcurrency` to all non-last modules;
- Make the target partition count an argument for the construction of `PartitioningMetadata`, and defer to the transport-specific implementation to decide on the value to be used: Redis and Rabbit will simply use `nextModuleCount`, and Kafka will use the partition count of the target topic;
- the number of the partitions used by the Kafka bus is the larger of `minPartitionCount` and `nextModuleConcurrency` * `nextModuleCount`, unless a topic exists and already has more partitions than either of the two values, case in which the number of existing partitions will be used;
- eliminate the additional calculation in the Kafka bus, which was required because the number of Spring XD partitions could have been, in principle, different from the number of underlying Kafka partitions;
- unit tests adjustments;

### DIFF
--- a/spring-xd-messagebus-kafka/src/main/java/org/springframework/xd/dirt/integration/kafka/KafkaMessageBus.java
+++ b/spring-xd-messagebus-kafka/src/main/java/org/springframework/xd/dirt/integration/kafka/KafkaMessageBus.java
@@ -137,7 +137,6 @@ public class KafkaMessageBus extends MessageBusSupport {
 
 	public static final int METADATA_VERIFICATION_MAX_INTERVAL = 1000;
 
-
 	public static final String FETCH_SIZE = "fetchSize";
 
 	public static final String QUEUE_SIZE = "fetchSize";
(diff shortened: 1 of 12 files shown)


spring-xd-messagebus-kafka/src/main/java/org/springframework/xd/dirt/integration/kafka/KafkaMessageBus.java
```diff
@@ -720,7 +718,7 @@ private KafkaMessageListenerContainer createMessageListenerContainer(KafkaProper
 	}
 
 	private KafkaMessageListenerContainer createMessageListenerContainer(KafkaPropertiesAccessor accessor,
-			String group, int numThreads, String topic,
+			String group, int maxConcurrency, String topic,
```
garyrussell: `s/numThreads/maxConcurrency/` in the other `createMessageListenerContainer()` methods?

garyrussell:
Otherwise LGTM; except maybe we should add a paragraph to "Stream Partitioning" about how it works with Kafka - also need to fix...

> NOTE
Partitioning is only allowed when using a RabbitMessageBus or a RedisMessageBus.

mbogoevici:
Will address documentation changes as part of XD-3016.

garyrussell:
Merged as 2180cdcf5537ba8cc763f71ab8245ef6be7fbe22

Doc fixes to come in a separate PR

PR #1696, opened 2015-06-08
[Sprint: 51] XD-3141: Prevent caching when using ClassPathResource

ericbottard:


ilayaperumalg:
Unfortunately, this change only fixes the re-loading of `config/*.properties` and `config/*.xml` file doesn't get re-loaded and still uses the already loaded one.

ericbottard:
I'm not sure I get it.
You say it fixes it.

I tried by changing the value of the options_class key for example to
something that does not exist: the change gets picked up (where it
previously didn't)

ilayaperumalg:
If you make a change in the xml file then that doesn't work.

ericbottard:
Ended up merging a fix similar to your first implementation Ilaya, as we can't prevent caching for the xml case easily. You won :) d892467

PR #1700, opened 2015-06-09
[Sprint: 51] XD-3160: Reorganize modules alphabetically

ericbottard:
Man, that was fun...

Also removed the "Introduction" section title, so that it does not appear as module name in the TOC.


dturanski:
For consistency, should we do the same for Jobs and Counters-and-Gauges?

ericbottard:
Arf, missed jobs. Silly me.
Looked at counters & gauges and thought it was ok. I've never been too good
with alphabet though :)
Will update the PR

ericbottard:
Updated jobs.
Leaving counters as is, as the narrative builds from simple to complex metrics and there are a lot of back references.

dturanski:
merged;

PR #1711, opened 2015-06-10
[Sprint: 51 (GA)] XD-3177: Make Rabbit Bus Properties Optional

garyrussell:
JIRA: https://jira.spring.io/browse/XD-3177

Fall back to boot defaults for rabbitmq properties if no
application.yml or servers.yml on classpath.

### DIFF
--- a/spring-xd-messagebus-rabbit/src/main/java/org/springframework/xd/dirt/integration/rabbit/ConnectionFactorySettings.java
+++ b/spring-xd-messagebus-rabbit/src/main/java/org/springframework/xd/dirt/integration/rabbit/ConnectionFactorySettings.java
@@ -38,10 +38,10 @@ import org.springframework.core.io.Resource;
 @Configuration
 public class ConnectionFactorySettings {
 
-	@Value("${spring.rabbitmq.useSSL}")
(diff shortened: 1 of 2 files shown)


ilayaperumalg:
I thnk this change will break rabbitmq modules that use `RabbitConnectionMixin` module option metadata. We need to update the defaults in the Mixin as well.

garyrussell:
Hmmmm - this simply makes passing the properties into the bus optional when using it outside of a Spring XD container (in spring-bus).

You are probably right if the mixin is used in a spring-bus module but it's not relevant for existing uses of the mixin within XD.

ilayaperumalg:
Yeah, I was just thinking about the protection for the module options when there are no application.yml/servers.yml config in CP. Inside XD we would always have application.yml protecting this.

ilayaperumalg:
It looks like defining empty default values for the placeholders at the `RabbitConnectionMixin` would violate the module option validation constraints.

ilayaperumalg:
LGTM, merging the changes as the above mentioned module options wouldn't have any impact within the context of XD.

ilayaperumalg:
Rebased and merged as 6de993eb21c99d9abc0184b511dc6a011b89b026

PR #1712, opened 2015-06-10
[Sprint: 51][1.2 GA] XD-3176 Adding Kerberos configuration for WritableResourceModuleRegistry

trisberg:
- this only applies when HDFS is used for custom modules

- pulling security properties from servers.yml first and if not found there then from hadoop.properties


jvalkeal:
this is great, it works. played with singlenode/kerberos and throughput sink. merging...

jvalkeal:
merged

PR #1720, opened 2015-06-16
[Sprint 51] XD-3164 Expose Kafka bus properties on producers/consumers

mbogoevici:
Exposes additional Kafka bus properties so that they can set on stream producers/consumers directly


ericbottard:
Rebased and merged as 6b9ac6f

PR #1721, opened 2015-06-17
XD-3186 Flag to set CL for ModuleOptions resolver

ilayaperumalg:
- Add `useParentClassLoader` flag to specify if the parent classloader needs to
be used when loading the module options metadata classes.
 - Remove unused constructor

### DIFF
--- a/spring-xd-module/src/main/java/org/springframework/xd/module/options/DefaultModuleOptionsMetadataResolver.java
+++ b/spring-xd-module/src/main/java/org/springframework/xd/module/options/DefaultModuleOptionsMetadataResolver.java
@@ -35,7 +35,6 @@ import org.springframework.core.io.Resource;
 import org.springframework.core.io.ResourceLoader;
 import org.springframework.core.io.support.PathMatchingResourcePatternResolver;
 import org.springframework.core.io.support.ResourcePatternResolver;
-import org.springframework.util.Assert;
 import org.springframework.xd.module.CompositeModuleDefinition;
 import org.springframework.xd.module.ModuleDefinition;
 import org.springframework.xd.module.SimpleModuleDefinition;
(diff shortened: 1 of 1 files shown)


spring-xd-module/src/main/java/org/springframework/xd/module/options/DefaultModuleOptionsMetadataResolver.java
```diff
@@ -101,15 +101,7 @@
...
-	/**
-	 * Construct a new {@link DefaultModuleOptionsMetadataResolver} that will use the provided conversion service when
-	 * converting from String to rich object (supported for {@link PojoModuleOptionsMetadata} only).
-	 *
-	 * @param conversionService Must not be null
-	 */
-	public DefaultModuleOptionsMetadataResolver(ConversionService conversionService) {
-		Assert.notNull(conversionService, "The provided conversionService must not be null.");
-	}
+	private boolean useParentClassLoader = false;
```
markfisher: This is slightly confusing given that the classloader only becomes a "parent" IF a child classloader is created. In other words, when this boolean is set to TRUE, then there actually is no parent classloader. Perhaps it would be clearer to flip the boolean (so it's true by default and would be set to false by the new modules) and use a name for the variable that indicates a new child classloader would be created if it's true. Does that make sense?

ilayaperumalg: Yeah, I agree. What do you think about having a flag named `hasModuleClassLoader` and set to `false` here in `DefaultModuleOptionsMetadataResolver` ? This way the module which has its classloader set already would set to `true`.

ericbottard:
Can you explain in which cases we want to use the parent classloader *first*?
I'm having a hard time understanding the requirement: for regular classes, there may be cases where a class with a single given name exists (in two different versions). But for metadata classes, this seems very artificial

markfisher:
@ericbottard for more context, see: https://github.com/markfisher/module-launcher/pull/2

specifically look at the MultiModuleLauncher (it's a work-in-progress, but the general idea should be self-explanatory)

the 3rd related PR is this one: https://github.com/spring-projects/spring-bus/pull/6

if you have suggestions for a better way to handle it, let us know

markfisher:
rebased, squashed and merged as https://github.com/spring-projects/spring-xd/commit/7f4a9505189c83c77e15d42757a2c2e1d9d20e83

PR #1722, opened 2015-06-17
XD-3187 AdminLeader cleanup deployment phase

ilayaperumalg:
- Cleanup stale container deployments only after the container path cache
is started


markpollack:
Seems like we should back port to 1.2.x branch?

ilayaperumalg:
Yeah, we need to back port this. Cherry-picked the changes and pushed ade19817f5718e35a9b83d8f184466f7b3ed9512

PR #1725, opened 2015-06-26
XD-2407: Add ability to upload a module from a dir

ericbottard:
Using this in my current tests for Spark refactorings, so might as well
make a PR for it.

### DIFF
--- a/spring-xd-shell/src/main/java/org/springframework/xd/shell/command/ModuleCommands.java
+++ b/spring-xd-shell/src/main/java/org/springframework/xd/shell/command/ModuleCommands.java
@@ -16,13 +16,27 @@
 
 package org.springframework.xd.shell.command;
 
+import java.io.BufferedInputStream;
 import java.io.File;
+import java.io.FileInputStream;
+import java.io.FileNotFoundException;
+import java.io.FilterInputStream;
 import java.io.IOException;
+import java.io.InputStream;
+import java.io.PipedInputStream;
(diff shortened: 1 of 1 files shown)


spring-xd-shell/src/main/java/org/springframework/xd/shell/command/ModuleCommands.java
```diff
@@ -141,10 +157,31 @@ public String uploadModule(
...
+				}
+
+				@Override
+				public long contentLength() throws IOException {
+					return -1; // and this is to prevent multiple stream reads, as
+					// this is now effectively a subclass of ISResource
+				}
+			};
+			final JarOutputStream jos = new JarOutputStream(pos);
+			new Thread(new DirectoryJarBuilder(file, jos)).start();
```
dturanski: Since this is async, if this fails, we get a success message back from the command anyway. Does it need to run in a thread?

ericbottard: It needs to run in  2 separate  threads because of piped  is/os.
Will need to see what happens in case of io error.
Le 10 juil. 2015 23:01, "David Turanski" <notifications@github.com> a
écrit :

dturanski:
Closing w/o merge per team consensus not to support this feature in 1.x (not applicable to 2.x, as file based module management will be obsolete)

PR #1717, opened 2015-06-16
[Sprint 52] 1.2.x XD-3125, XD-3126 refactor and enhance serialization configuration

dturanski:
This PR, motivated by XD-3125 and XD-3126, refactors and enhances Kryo serialization. Here is a summary of the changes

* The "codec" bean, CompositeCodec {TupleCodec, FileCodec, PojoCodec} is replaced by PojoCodec with custom serializers configured for Tuple and File.
* PojoCodec is wired with List<AbstractKryoRegistrar>. Previously, KryoRegistrar was a singleton that was configured only if a user wished to register domain objects with a unique ID.  Now, PojoCodec may be configured with multiple `AbstractKryoRegistrar`s which are merged and validated, (checked for conflicting registrations) at start up. Kryo Registrations are logged at the INFO level and any conflicts will result in a SerializationExeption at start up. XD provides `TupleKryoRegistrar` and `FileKryoRegistrar` and users may provide additional ones and register them in the SharedServerContext. Spring will "autowire" any `AbstractKryoRegistrar`s in a known base package or XML location.  Note that annotation based autowiring was tried but turned out to be less clean in this case.
* codec.xml replaced with Java config in consideration of support for potentially implementing an alternate "codec" without forcing xml config. Related to https://jira.spring.io/browse/SPR-7341. Currently "codec" is configured in `SharedServerContextConfiguration` and `@ConditionalOnMissingBean(name = "codec")`  Currently the message bus configuration requires a bean named "codec". That could be improved via autowiring as well, but looks like the DI will require additional refactoring in the class hierarchy. 
* Added `RegistrationKryoRegistrar` to support user defined Kryo Serializers. This takes a `List<Registration>`, where Registration is a native Kryo type associating (type, Serializer, ID). 
* The existing extension point for serializers is META-INF/spring-xd/bus/ext/*.xml. This is retained for backward compatibility. Component scanning is also enabled for "spring.xd.bus.ext". This package is hard coded currently in SharedServerContextConfiguration. It didn't seem worth making the base package configurable for this as we did for PluginContext extensions (it seems that feature has been used very little, if at all).  I would consider a different package name, but currently it mirrors, more or less, the xml path.  

[...]

PR #1726, opened 2015-06-29
[Sprint 52] 1.2.x only XD-3125, XD-3126 Refactor Codec configuration.

dturanski:
This is a redo of https://github.com/spring-projects/spring-xd/pull/1717 addressing @markpollack review comments.


spring-xd-codec/src/main/java/org/springframework/xd/dirt/integration/bus/serializer/kryo/AbstractKryoCodec.java
```diff
@@ -38,17 +38,6 @@
  */
 public abstract class AbstractKryoCodec<T> extends AbstractCodec<T> {
 
-	protected final static int FILE_REGISTRATION_ID = 40;
```
markpollack: The intention here is that these would be done by the user should they want to have IDs instead of strings for classnames for these types?

spring-xd-codec/src/main/java/org/springframework/xd/dirt/integration/bus/serializer/kryo/AbstractKryoCodec.java
```diff
@@ -38,17 +38,6 @@
  */
 public abstract class AbstractKryoCodec<T> extends AbstractCodec<T> {
```
markpollack: Related to the optimization work for serialization/deserialization that would go into another PR, AbstractCode always using a ByteArrayInputStream might not be a good idea.  Perhaps we should change this abstract base class to an interface.

markpollack: AbstractKryoCodec overrides deserialize(byte[] bytes) with the same exact implementation as the base class, will remove on merge.

markpollack:
Do we want this also for the 1.2.x branch?

dturanski:
Only for 1.2 branch

markpollack:
but also on master?

Take a look at https://github.com/markpollack/spring-xd/tree/dturanski-XD-3125b and the rework in AbstractKryoMultiTypeCodec to see if you are ok with it.

dturanski:
I thought master was for 2.0. All this will move to spring-bus

markpollack:
merged on 1.2.x as bc15d3861732a618d61c8e96508469e9afcd47ae

PR #1731, opened 2015-07-06
[Sprint: 53] XD-3218 Handle deployment status recalucation failures

ilayaperumalg:
- If the admin fails to load the stream/job (mostly if the module registry isn't in sync)
when recalculating the deployment status, then set the stream/job deployment unit status
to `unknown` and logs error message
  - This way the admin leader election would still happen but the stream status will
notify the user to act on.


pperalta:
Merged as 845636657d485773382c1f4b6ce76ea639692000

PR #1736, opened 2015-07-09
[Sprint: 53] XD-3232 Update Integration, Framework, AMQP, Batch

garyrussell:
JIRA: https://jira.spring.io/browse/XD-3232

* Update versions
* SI TCP serializers need to be flushed since 4.2
* Remove uses of deprecated APIs
* Some modules were relying on a transitive dependency on slf4j which is no longer there
* Work around for INT-3766 (reactor-code but not reactor-stream on CP)


dependencies.properties
```diff
@@ -288,34 +288,33 @@ org.slf4j\:slf4j-api=1.7.11
...
-org.springframework\:spring-orm=4.1.6.RELEASE
-org.springframework\:spring-oxm=4.1.6.RELEASE
-org.springframework\:spring-test=4.1.6.RELEASE
-org.springframework\:spring-tx=4.1.6.RELEASE
-org.springframework\:spring-web=4.1.6.RELEASE
-org.springframework\:spring-webmvc=4.1.6.RELEASE
-org.springframework\:spring-webmvc-portlet=4.1.6.RELEASE
-org.springframework\:spring-websocket=4.1.6.RELEASE
-org.springframework.amqp\:spring-amqp=1.4.5.RELEASE
-org.springframework.amqp\:spring-erlang=1.4.5.RELEASE
```
ericbottard: Is it normal that spring-erlang is gone in your version?

garyrussell: Yes, spring-erlang is no more (as of 1.5.0). It is replaced by the use of the rabbit REST API via the `RabbitManagementTemplate`.

ericbottard: Excellent, was just checking, as this jumped at me while merging.

garyrussell:
Rebased; pushed update to s-i-kafka to fix SpEL evaluation context initialization problem.

markpollack:
Squashed and merged as 94c44a3edd68a691ad658b0ddffb3742e1f7ea9f

PR #1742, opened 2015-07-10
[Sprint: 53] XD-3244 - Delete message bus related tests -XD 2.0 (master) only

dturanski:
Removes some test classes for components that will become obselete in XD-2.0


ericbottard:
Merged as 9425c20

PR #1745, opened 2015-07-14
[Sprint 53] 1.2.x only add kryo optimizations discovered in benchmarking and clean up

dturanski:
Addresses XD-3238 and XD-2825
Not the corresponding changes will be mirrored in s-c-s. 

* Output creation optimized in serialize methods
* Support kryo.setReferences(false) which is optimal for object graphs with no cycles, but required otherwise. The default is true for backward compatibility. In general, we must be able to serialize any Pojo. However if users know that that there are no cyclical references in their domain, and they need high throughput, can set the property _spring.xd.codec.kryo.references: false_ in servers.yml. 
*  XD-2825. After simplifying the class hierarchy, there  is no longer an implementation required of _protected T doDeserialize(Kryo kryo, Input input). 
* AbstractMultiTypeKryoCodec<T> has been removed and merged into AbstractKryoCodec. XD no longer implements an Codecs for specific types. Only PojoCodec is used. CompositeCodec is not used but preserved just in case.


spring-xd-dirt/src/main/java/org/springframework/xd/dirt/server/MessageBusExtensionsConfiguration.java
```diff
@@ -43,13 +44,16 @@
 public class MessageBusExtensionsConfiguration {
 	@Autowired
 	ApplicationContext applicationContext;
+	
+	@Value("${spring.xd.codec.kryo.references:true}")
```
ilayaperumalg: May be we need servers.yml entry for this? also, instead of specifying the default value here, can we define it in application.yml?

config/servers.yml
```diff
@@ -387,3 +387,14 @@ endpoints:
 #         backOffMaxInterval:        10000
 #         backOffMultiplier:         2.0
 #         maxAttempts:               5
+
+---
+# Kryo uses references by default. This feature must be enabled unless we know the payload types being serialized do not
+# contain circular refefences. Disabling this feature improves performance if it is not needed.
```
ilayaperumalg: typo: references. will fix while merging.

ilayaperumalg: Though commented, the default value is set to `false` (which makes it performant). This might confuse the user about what is mentioned in the description: `Disabling this feature improves performance if it is not needed.` -> whether to set to true or false to make it performant.

spring-xd-dirt/src/main/resources/application.yml
```diff
@@ -349,4 +349,13 @@ transport: local
 analytics: memory
 embeddedHsql: true
 
----
\ No newline at end of file
+---
+# Kryo uses references by default. This feature must be enabled unless we know the payload types being serialized do not
+# contain circular refefences. Disabling this feature improves performance if it is not needed.
```
ilayaperumalg: Instead of using `enabling` or `disabling`, can we specify what happens it this is set to `true` or `false`?  That way it will be very clear to the end user.

ilayaperumalg:
LGTM, Once we finalize on the description, I can merge.

ilayaperumalg:
Squashed, rebased and merged as 1fce81983dedb901998b221a7a0051cf8ffc56ac

PR #1747, opened 2015-07-15
XD-3251: Add a boot @ConfigurationProperties compatible MOMResolver.

ericbottard:
Upgrade to latest Boot SNAPSHOT (introduce an explicit task scheduler for "framework" tasks, different from the batch and SI schedulers), upgrade Spring Batch.

Add a boot @ConfigurationProperties compatible MOMResolver.

With this PR merged, one could drop a spring-cloud-stream module (such as the one resulting from https://github.com/spring-cloud/spring-cloud-stream/pull/19) and properly see its "module options". Obviously, running the module as-is won't work, will need adaptation to the execution environment (in the form of the SPI @pperalta is working on) with the knowledge that any module option `--foo=xxx` must actually be passed to the s-c-s module as `module.foo=xxx`.


spring-xd-dirt/src/main/java/org/springframework/xd/dirt/server/admin/AdminServerApplication.java
```diff
@@ -189,4 +194,18 @@ public Filter httpPutFormContentFilter() {
...
+	@Autowired
+	private ScheduledAnnotationBeanPostProcessor postProcessor;
+
+	@Autowired
+	@Qualifier("framework")
+	private TaskScheduler frameworkTaskScheduler;
+
+	@PostConstruct
+	public void forceScheduledTaskExecutor() {
+		postProcessor.setScheduler(frameworkTaskScheduler);
```
garyrussell: Interesting technique :smile: 

A couple of thoughts...

* Perhaps we should open a JIRA against SF to fall back to some well-known bean name when there's more than one scheduler in the context?
* We should probably document someplace how users can override this "framework" scheduler if they want multiple threads in the `TPTS`

markfisher: +1 to Gary's comments, and...

I'm not a fan of the word "framework" here.

Also the thread prefix should be set.

ericbottard:
Closing as s-c-d is its own project, not going through the soft transition path

PR #1749, opened 2015-07-15
[Sprint: 53] XD-631 Pluralize test classes

ghillert:
* Pluralize test classes in package org.springframework.xd.shell.command


cppwfs:
Only one issue I found, based on the Jira was that AbstractCommandTemplate should also be AbstractCommandsTemplate.  If this assumption  is correct I can fix on Merge.

cppwfs:
LGTM Merged

PR #1750, opened 2015-07-15
[Sprint: 53] XD-3235 Test looks for specific files vs *

cppwfs:
* Also set up the FileJdbc tests have unique job names instead of ec2test3


mminella:
LGTM.  Merged as 40bb9ba9815900d770137e2d4087f81ade243344

PR #1768, opened 2015-08-14
Update to SF 4.2, SI 4.2.0.RC1

garyrussell:
See XD-3394 about the SFTP source change.


garyrussell:
RC1 has been promoted to the milestone repo; holding off the announcement until Monday.

jvalkeal:
just my 2 cents, I have some concerns of mixing boot 1.2.x and core 4.2.x. I'm a bit nervous about this SO post http://stackoverflow.com/questions/31974101/spring-applicationlistener-neo4j-beforesaveevent-doesnt-work-with-spring-4-2-0 because it might be some trouble with using boot 1.2.x and core 4.2.x together.

Some combinations of boot 1.3.0 milestones didn't work very well with some of a core 4.2.0 milestones. I honestly think that update to core 4.2 should be done together with update to boot 1.3.0.M3. Thought boot 1.3.0.M3 needs a fix because of https://github.com/spring-projects/spring-boot/issues/3724 which I discussed with @ericbottard.

I don't say that we should not merge this PR but just a heads up if we see some weird issues after it.

garyrussell:
@jvalkeal For some reason, I though master was already using boot 1.3 but I see it's not.

garyrussell:
There's no rush to merge this; I just wanted to get it in the pipeline before leaving for PTO, given it needed an SFTP change and I didn't want someone else have to figure out what happened.

garyrussell:
We should change this to update to the GA versions of SI, Spring AMQP and boot M5.

@sabbyanandan We should add a JIRA to the current Sprint for this.

garyrussell:
By the way - Spring Boot 1.3 GA (provisional) is not until Oct 14 - does that align with XD 1.3?

cppwfs:
Makes sense.

trisberg:
@garyrussell based on current date for XD 1.3 (29/Oct/15) I think we will be OK. Spring Hadoop 2.3, which XD 1.3 will use, also  depends on Boot 1.3.

dturanski:
@garyrussell - I merged these Sftp config changes on my XD-3493 branch. And also upgrade to SI 4.2.0.RELEASE as part of that.

garyrussell:
Thanks; we also need Spring AMQP -> 1.5.0.RELEASE (although we'll likely have a 1.5.1 before XD GAs).

PR #1785, opened 2015-09-29
Fix Cassandra doc

ilayaperumalg:


extensions/spring-xd-extension-cassandra/src/main/java/org/springframework/xd/cassandra/mixins/CassandraConnectionMixin.java
```diff
@@ -44,7 +44,7 @@
 
 	private String initScript;
 
-	private String[] entityBasePackages = new String[0];
+	private String[] entityBasePackages;
```
artembilan: @ilayaperumalg, have you tested it before pushing?
The reason for the `new String[0]` is here (`AbstractCassandraConfiguration`):
```
/**
 * The base packages to scan for entities annotated with {@link Table} annotations. By default, returns the package
 * name of {@literal this} (<code>this.getClass().getPackage().getName()</code>). This method must never return null.
 */
public String[] getEntityBasePackages() {
	return new String[] { getClass().getPackage().getName() };
}
```
And so on up to the:
```
  ArrayList(E[] array) {
         a = Objects.requireNonNull(array);
}

```

ericbottard: @ilaya, maybe we should change the doc generation (eg to use
ArrayUtils.toCommaDelimitedString) because an array will *never* produce
something pretty anyway. So even if we workaround this one, another case
will pop up tomorrow

artembilan: +1 for @ericbottard!

From other side we should revert this change for the reason I shown above.
Even if you tested it, we just don't have the test for this case. I had it before in the `modules` project, but @markpollack remove it here because it was `@Ignored` and didn't work from test environment by the ClassLoader issue.

markpollack:
Not clear to me how we can get that type of test to pass in both the IDE and gradle...not sure where we would need to copy Book.class in each case... Any suggestions?  (For now eric said he would change the doc generation, the null value isn't an option)

artembilan:
@markpollack , we should copy `Book.class` to the `xd/lib`, like we do in normal environment: https://github.com/spring-projects/spring-xd-modules/blob/master/cassandra-sink/README.md.
Please, find the expression:
> Another more tricky INSERT

Fortunately it is less work right now, because all Cassandra stuff is on the parent CP already. Thanks to this @trisberg 's fix :smile: .

But... Do we really want to do that tricky work for the testing, if we are going to move to the `spring-cloud-streams`, where we will have only one ClassLoader already?

markpollack:
xd/lib is not available when running our tests.  Certainly this will not be an issue for scs.  Looking for a solution within XD test/build env.  We can always make it part of the formal integration tests (that work against a .zip distro)

markpollack:
Closing - created https://jira.spring.io/browse/XD-3552 to track doc fixes.

PR #1792, opened 2015-10-01
[DO NOT MERGE] Removing Hadoop eco system project versions from dependencies.properties

trisberg:


trisberg:
Reworking, seems to be more than one issue to get XD running on YARN

PR #1793, opened 2015-10-01
XD-3566 makes the twitterstream stream name unique

cppwfs:
Also updated twittersearch test so the generated stream will be unique


markpollack:
merged as 4897562616adf95934352f89b469f2a089380881

PR #1799, opened 2015-10-05
[Sprint: 59] XD-3394: SFTP Configurable Known Hosts File

garyrussell:
JIRA: https://jira.spring.io/browse/XD-3394

Previously, unknown host keys were always allowed and it wasn't configurable.
This is not best practice for SSH.

Require a `known_hosts` file by default; users can restore the previous
behavior by setting `allowUnknownKeys` to `true` in the deployment properties.

Support SpEL expressions for the known_hosts file location, such as

`--knownHostsExpression=systemProperties[\"user.home\"]+\"/.ssh/known_hosts\"`

which is the default local on Unix systems.


ericbottard:
Should it be called `knownHostsExpression`, for consistency with all other Expressions in modules (I agree, kinda ugly, but...)

garyrussell:
Yeah - that crossed my mind too; re-pushed with change, so I could fix the commit comment too.

modules/source/sftp/config/sftp.xml
```diff
@@ -21,7 +21,8 @@
 		<beans:property name="password" value="${password}" />
 		<beans:property name="privateKey" value="${privateKey}" />
 		<beans:property name="privateKeyPassphrase" value="${passPhrase}" />
-		<beans:property name="allowUnknownKeys" value="true"/>
+		<beans:property name="allowUnknownKeys" value="${allowUnknownKeys}" />
+		<beans:property name="knownHosts" value="#{'UNDEFINED'.equals('${knownHostsExpression}') ? null : ${knownHostsExpression}}" />
```
ericbottard: This feels weird. Can't we use something like `value="${knownHostsExpression:#{null}}"` and have the field default value be `null`?

spring-xd-shell/src/test/java/org/springframework/xd/shell/command/AbstractSftpSourceTests.java
```diff
@@ -64,7 +64,11 @@ public void setUp() {
 	public void testSftpSource() {
 		String streamName = generateStreamName();
 		CounterSink counter = metrics().newCounterSink(streamName);
-		stream().create(streamName, "sftp --user=%s --password=%s --port=%s --remoteDir=%s --localDir=%s | %s",
+		stream().create(streamName, "sftp --user=%s --password=%s --port=%s --remoteDir=%s --localDir=%s "
+						+ "--allowUnknownKeys=true | %s",
+						// manual test; verify the key was added to the kh file
```
ericbottard: Is this meant to stay here?

modules/source/sftp/config/sftp.xml
garyrussell: That would work too - I went this route because `${knownHostsExpression}` didn't work with a null property (unresolved property).

That seems wrong to me - it prevents explicitly setting a property to `null`, although I guess the DSL doesn't support that anyway.

I was also thinking of the generated docs - default: `UNDEFINED` seems more doc friendly than default: null.

But I can change it if you prefer.

spring-xd-shell/src/test/java/org/springframework/xd/shell/command/AbstractSftpSourceTests.java
garyrussell: I thought it would be useful to leave it as a comment.

I didn't think it was worth the extra effort to set up an automated test for this.

modules/source/sftp/config/sftp.xml
ericbottard: This is the way we do with many properties that have to deal with `null`.

spring-xd-shell/src/test/java/org/springframework/xd/shell/command/AbstractSftpSourceTests.java
ericbottard: Gotcha

modules/source/sftp/config/sftp.xml
garyrussell: ok; I'll change it

garyrussell:
Pushed - slightly different to your suggestion since the property is a String, not an expression, so it needs `#{}` for SF to evaluate it.

ericbottard:
Merged as 784e5f7 with a fix to the option description (XD requires lowercase at start)

PR #1800, opened 2015-10-07
[Sprint: 59] XD-3508: Use the SI Codec/Kryo Code

garyrussell:
JIRA: https://jira.spring.io/browse/XD-3508

The Codec/Kryo code was moved to SI in 4.2, albeit with a slightly different API.

Also fix the package for `XDRuntimeException` to avoid the same package being in 2 jars.


garyrussell:
Re-pushed after reverting `XDRuntimeException` package change.

PR #1803, opened 2015-10-07
Add a Job DSL Parser

aclement:
Related to https://jira.spring.io/browse/XD-3603

I now have code blindness from staring at these changes and it is time for some feedback from you guys. This commit adds a package alongside `...dirt.stream.dsl` called `...dirt.job.dsl` and includes a `JobParser` for the Job DSL we have been talking about.
Notes:
-  it is not an extension of the stream parser, it is a different parser. Different tokenization, differing syntax.
-  unlike the stream parser which mixed up syntactic with semantic checking, this one is pure syntax checking. There is a basic visitor pattern that could be exploited for the parsed AST where semantic validation could easily be done (i.e. is the job definition supplying the right arguments? Are the job references all references to real jobs?).
- I have (in a basic way) wired it into the ToolsController for use by Flo (similarly to how the stream parser was wired in).
- The stream parser used to return a structure back to the client which then got converted to a graph for Flo display. This parser can build the graph directly for return to the client. Are we ok with this? (See the Graph return value on the ToolsController request mapping)
- This AST that is returned has a `toXML(batchJobId)` endpoint that produces some basic XML based on a quick exchange with Michael, this will likely need tweaking but should prove a useful start.

More work to be done (in October...):
- validation
- error handling (passing errors to the client).

Syntax:

Jobs are either defined inline or reference names are used (reference names are to existing job definitions).

- single job:  `foo`
- single inline job definition, called 'foo': `filejdbc foo --aaa=bbb --ccc=ddd`
- two jobs in a sequence: `aaa || bbb`
- two jobs in parallel: `<aaa & bbb>`
- a job that transitions to a different state when it exits in a certain 'state'. Here 'aaa' will jump to 'ccc' if it exits in COMPLETED state (skipping 'ddd'):
  `<aaa | COMPLETED=ccc & bbb> || ddd || ccc`
- transition states can be quoted if include spaces or other funky chars
- multiple transition states can be specified by including multiple `| state = targetJobReference`

Any feedback welcome.

### DIFF
--- a/spring-xd-dirt/src/main/java/org/springframework/xd/dirt/job/dsl/JobParser.java
+++ b/spring-xd-dirt/src/main/java/org/springframework/xd/dirt/job/dsl/JobParser.java
@@ -34,9 +34,9 @@ public class JobParser {
 	/**
 	 * Parse a job flow definition into an abstract syntax tree (AST).
 	 *
-	 * @param the textual job specification
-	 * @return the AST for the parsed stream
-	 * @throws JobSpecificationException
+	 * @param jobSpecification the textual job specification
+	 * @return the AST for the parsed job
+	 * @throws JobSpecificationException if any problems occur during parsing
 	 */
 	public JobSpecification parse(String jobSpecification) {
 		tokens = new Tokens(jobSpecification);
@@ -123,6 +123,10 @@ public class JobParser {
 				jd = new JobReference(tokens.eat(TokenKind.IDENTIFIER));
 			}
 		}
+		if (jd == null) {
+			Token t = tokens.peek();
+			tokens.raiseException(t == null ? 0 : t.startpos, JobDSLMessage.EXPECTED_JOB_REF_OR_DEF);
+		}
 		if (tokens.peek(TokenKind.PIPE)) {
 			List<Transition> transitions = parseTransitions();
 			jd.setTransitions(transitions);
@@ -141,6 +145,11 @@ public class JobParser {
 						tokenToString(transitionName));
 			}
 			Token nextToken = tokens.peek();
+			if (nextToken == null) {
+				// no equals after the transition name
+				tokens.raiseException(transitionName.endpos - 1, JobDSLMessage.MISSING_EQUALS_AFTER_TRANSITION_NAME,
+						tokenToString(transitionName));
+			}
 			if (!nextToken.isKind(TokenKind.EQUALS)) {
 				tokens.raiseException(nextToken.startpos, JobDSLMessage.EXPECTED_EQUALS_AFTER_TRANSITION_NAME,
 						tokenToString(nextToken));
--- a/spring-xd-dirt/src/main/java/org/springframework/xd/dirt/job/dsl/JobParser.java
+++ b/spring-xd-dirt/src/main/java/org/springframework/xd/dirt/job/dsl/JobParser.java
@@ -1,0 +1,270 @@
+/*
+ * Copyright 2015 the original author or authors.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *      http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.springframework.xd.dirt.job.dsl;
+
+import java.util.ArrayList;
+import java.util.List;
+
+/**
+ * Parse a Batch DSL Job specification.
+ *
+ * @author Andy Clement
+ */
+public class JobParser {
+
+	private Tokens tokens;
+
+	public JobParser() {
+	}
+
+	/**
+	 * Parse a job flow definition into an abstract syntax tree (AST).
+	 *
+	 * @param the textual job specification
+	 * @return the AST for the parsed stream
+	 * @throws JobSpecificationException
+	 */
+	public JobSpecification parse(String jobSpecification) {
+		tokens = new Tokens(jobSpecification);
+		JobNode jobNode = parseJobNode();
+		JobSpecification js = new JobSpecification(jobSpecification, jobNode);
+		if (tokens.hasNext()) {
+			throw new JobSpecificationException(tokens.getExpression(), tokens.peek().startpos,
+					JobDSLMessage.UNEXPECTED_DATA_AFTER_JOBSPEC,
+					toString(tokens.next()));
+		}
+		return js;
+	}
+
+	private JobNode parseJobNode() {
+		// Handle (...)
+		if (tokens.maybeEat(TokenKind.OPEN_PAREN)) {
+			JobNode jn = parseJobNode();
+			tokens.eat(TokenKind.CLOSE_PAREN);
+			return jn;
+		}
+		// Handle a split < ... >
+		if (tokens.peek(TokenKind.SPLIT_OPEN)) {
+			JobNode jn = parseSplit();
+			// is the split part of a flow? "<..> || b"
+			return maybeParseFlow(jn);
+		}
+		JobDescriptor jd = parseJobDescriptor();
+		// Handle a potential flow "a || b"
+		return maybeParseFlow(jd);
+	}
+
+	private JobNode maybeParseFlow(JobNode firstJobNodeInFlow) {
+		if (tokens.peek(TokenKind.DOUBLE_PIPE)) {
+			List<JobNode> jobNodes = new ArrayList<>();
+			jobNodes.add(firstJobNodeInFlow);
+			while (tokens.maybeEat(TokenKind.DOUBLE_PIPE)) {
+				JobNode nextNode = parseJobNode();
+				// If nextNode is a Flow node, merge it with this one
+				if (nextNode instanceof Flow) {
+					jobNodes.addAll(nextNode.getSeries());
+				}
+				else {
+					jobNodes.add(nextNode);
+				}
+			}
+			return new Flow(jobNodes);
+		}
+		else {
+			return firstJobNodeInFlow;
+		}
+	}
+
+	// '<' jobs ['&' jobs]+ '>'
+	private JobNode parseSplit() {
+		List<JobNode> flows = new ArrayList<>();
+		tokens.eat(TokenKind.SPLIT_OPEN);
+		// '<' jobSequence  [ '&' jobSequence]* '>'
+		flows.add(parseJobNode());
+		while (tokens.maybeEat(TokenKind.AMPERSAND)) {
+			flows.add(parseJobNode());
+		}
+		tokens.eat(TokenKind.SPLIT_CLOSE);
+		return new Split(flows);
+	}
+
+	// JobDescriptor can be a simple JobReference (reference to a pre-existing definition) or
+	// an inlined JobDefinition.
+	//
+	// JobReference := identifier:jobname [stateTransitions]*
+	// JobDefinition := identifier:jobmodulename identifier:jobname ['--'optionname'='optionvalue]* [stateTransitions]*
+	private JobDescriptor parseJobDescriptor() {
+		JobDescriptor jd = null;
+		// Double identifiers indicates job definition
+		if (tokens.peek(0, TokenKind.IDENTIFIER)) {
+			if (tokens.peek(1, TokenKind.IDENTIFIER)) {
+				jd = parseJobDefinition();
+			}
+			else if (tokens.peek(1, TokenKind.DOUBLE_MINUS)) {
+				// Error: something like 'foo --bar=xxx' - an inline job definition needs
+				// the job name to follow the job module
+				tokens.raiseException(tokens.peek(1).startpos, JobDSLMessage.MISSING_JOB_NAME_IN_INLINEJOBDEF);
+			}
+			else {
+				jd = new JobReference(tokens.eat(TokenKind.IDENTIFIER));
+			}
+		}
+		if (tokens.peek(TokenKind.PIPE)) {
+			List<Transition> transitions = parseTransitions();
+			jd.setTransitions(transitions);
+		}
+		return jd;
+	}
+
+	// [| state = target]*
+	// where state is an identifier/string literal whilst target is an identifier (target job).
+	private List<Transition> parseTransitions() {
+		List<Transition> transitions = new ArrayList<>();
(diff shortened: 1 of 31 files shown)


ericbottard:
I'm sure this has been thought through a lot, but I don't quite get the syntax for parallel and sequence.

There is meaning for `||` and `&&` in a shell (to which we keep referring to for streams, as "like a unix shell"). 
So `A && B` would mean "run A *and then* B, with requirement for A to complete successfully.

I understand we don't want to map to those meanings, but I think they may bring some knowledge baggage.


Also, regarding inline naming: Do we want to have something similar to streams module labelling? _i.e._ `foo: filejdbc` instead of `filejdbc foo`.

Again, this just pops out of the top of my head. Maybe a GoogleDoc would be more appropriate for further discussion

mminella:
> There is meaning for || and && in a shell (to which we keep referring to for streams, as "like a unix shell"). 
So A && B would mean "run A and then B, with requirement for A to complete successfully.

I think you mean `A || B` would mean "run A and then B,  with the requirement for A to complete successfully".  That's actually what we're looking for which is why we went with that syntax.  Along with that, when using & in a shell, it means to background so you can run other things in the foreground which we felt was similar to the concept of a split (executing multiple flows in parallel.

spring-xd-dirt/src/main/java/org/springframework/xd/dirt/job/dsl/Link.java
```diff
@@ -0,0 +1,51 @@
...
+ */
+
+package org.springframework.xd.dirt.job.dsl;
+
+import java.util.HashMap;
+import java.util.Map;
+
+
+/**
+ * Represents a node in a {@link Graph} object that Flo will display as a block.
```
cppwfs: Copy Paste hiccup

spring-xd-dirt/src/main/java/org/springframework/xd/dirt/job/dsl/Tokenizer.java
```diff
@@ -0,0 +1,388 @@
...
+						justProcessedEquals = true;
+						pushCharToken(TokenKind.EQUALS);
+						break;
+					case ' ':
+					case '\t':
+					case '\r':
+						// drift over white space
+						pos++;
+						break;
+					case '\n':
```
cppwfs: We could move the \n with other cases: i.e. ' ', '\t', '\r'  a

spring-xd-dirt/src/main/java/org/springframework/xd/dirt/job/dsl/Tokenizer.java
```diff
@@ -0,0 +1,388 @@
...
+						pos++;
+						//						pushCharToken(TokenKind.NEWLINE);
+						break;
+					//					case '.':
```
cppwfs: We can remove the commented out code, until we need those tokens.

spring-xd-dirt/src/main/java/org/springframework/xd/dirt/job/dsl/Tokenizer.java
```diff
@@ -0,0 +1,388 @@
...
+						break;
+					//					case '@':
+					//						pushCharToken(TokenKind.REFERENCE);
+					//						break;
+					case 0:
+						// hit sentinel at end of char data
+						pos++; // will take us to the end
+						break;
+					case '\\':
+						throw new StreamDefinitionException(
```
cppwfs: I'm probably being doing nitpicky here... but would we want to have a JobDefinitionException.  Feel free to argue with me. ;-)

spring-xd-dirt/src/main/java/org/springframework/xd/dirt/job/dsl/Tokenizer.java
```diff
@@ -0,0 +1,388 @@
...
+		tokens.add(new Token(TokenKind.IDENTIFIER, subarray, start, pos));
+	}
+
+	/**
+	 * For the variant tokenizer (used following an '=' to parse an argument value) we only terminate that identifier if
+	 * encountering a small set of characters. If the argument has included a ' to put something in quotes, we remember
+	 * that and don't allow ' ' (space) and '\t' (tab) to terminate the value.
+	 */
+	private boolean isArgValueIdentifierTerminator(char ch, boolean quoteOpen) {
+		return (ch == '|' && !quoteOpen) || (ch == ';' && !quoteOpen) || ch == '\0' || (ch == ' ' && !quoteOpen)
```
cppwfs: do we need to add a (ch == '&' && !quoteOpen) ?

spring-xd-dirt/src/main/java/org/springframework/xd/dirt/job/dsl/Link.java
aclement: good catch, thanks for all the feedback Glenn. As I say, I had code blindness and you've picked up on a few things i need to sort out. I've also got some new features already that I might fold in with the next version of the PR (Flo needs a graph-to-DSL conversion function and I have that working now).

spring-xd-dirt/src/main/java/org/springframework/xd/dirt/job/dsl/Tokenizer.java
aclement: heh, no argument. Another copy/paste problem, damn the exposed visibility of that exception class from the other package.

spring-xd-dirt/src/main/java/org/springframework/xd/dirt/job/dsl/Tokenizer.java
aclement: yep, I think you are right.

spring-xd-dirt/src/main/java/org/springframework/xd/dirt/job/dsl/JobParser.java
```diff
@@ -0,0 +1,270 @@
...
+	// JobDescriptor can be a simple JobReference (reference to a pre-existing definition) or
+	// an inlined JobDefinition.
+	//
+	// JobReference := identifier:jobname [stateTransitions]*
+	// JobDefinition := identifier:jobmodulename identifier:jobname ['--'optionname'='optionvalue]* [stateTransitions]*
+	private JobDescriptor parseJobDescriptor() {
+		JobDescriptor jd = null;
+		// Double identifiers indicates job definition
+		if (tokens.peek(0, TokenKind.IDENTIFIER)) {
+			if (tokens.peek(1, TokenKind.IDENTIFIER)) {
```
cppwfs: Just curious on this one.  Is this case where we would have 2 identifiers back to back.  Would this be a case as enumerated above as identifier:jobname.  If so we need to update the tokenizer to support the ':'.

aclement: We don't yet support labelling (with ':'), the colon mentioned in the comments above is merely my crude grammar description.  'identifier:jobname' is reminding me that we expect an identifier and that identifier should be treated as a jobname.  More work will need doing if we decide to support labelling.  I included those suffixes in the comments so that when I see 'identifier identifier' I know what I'm supposed to be treating those identifiers as.

spring-xd-dirt/src/main/java/org/springframework/xd/dirt/job/dsl/Tokens.java
```diff
@@ -0,0 +1,307 @@
...
+	protected Token next() {
+		if (!hasNext()) {
+			raiseException(expression.length(), JobDSLMessage.OOD);
+		}
+		return tokenStream.get(position++);
+	}
+
+	/**
+	 * Consume the next token if it matches the indicated token kind;
+	 * otherwise throw {@link CheckPointedParseException}.
```
cppwfs: CheckPointedParseException  needs to be CheckpointedJobDefinitionException

spring-xd-dirt/src/main/java/org/springframework/xd/dirt/job/dsl/Tokens.java
```diff
@@ -0,0 +1,307 @@
...
+	 *
+	 * @param expectedKind the expected token kind
+	 * @return the next token
+	 * @throws CheckPointedParseException if the next token does not match
+	 * the expected token kind
+	 */
+	protected Token eat(TokenKind expectedKind) {
+		Token t = next();
+		if (t == null) {
+			raiseException(expression.length(), JobDSLMessage.OOD);
```
cppwfs: CheckPointedParseException needs to be CheckpointedJobDefinitionException

spring-xd-dirt/src/main/java/org/springframework/xd/dirt/job/dsl/Tokens.java
```diff
@@ -0,0 +1,307 @@
...
+	 * Indicate that the current token has been successfully processed.
+	 *
+	 * @see #lastGoodPosition
+	 */
+	protected void checkpoint() {
+		lastGoodPosition = position;
+	}
+
+	/**
+	 * Throw a new {@link CheckPointedParseException} based on the current and
```
cppwfs: CheckPointedParseException needs to be CheckpointedJobDefinitionException

spring-xd-dirt/src/main/java/org/springframework/xd/dirt/job/dsl/JobParser.java
```diff
@@ -0,0 +1,270 @@
...
+
+	private Tokens tokens;
+
+	public JobParser() {
+	}
+
+	/**
+	 * Parse a job flow definition into an abstract syntax tree (AST).
+	 *
+	 * @param the textual job specification
```
cppwfs: need to enumerate "jobSpecification" in the @param

spring-xd-dirt/src/main/java/org/springframework/xd/dirt/job/dsl/Tokens.java
```diff
@@ -0,0 +1,307 @@
...
+			return false;
+		}
+
+		Token last = tokenStream.get(position - 1);
+		Token next = tokenStream.get(position);
+		return next.startpos == last.endpos;
+	}
+
+	/**
+	 * Indicate that the current token has been successfully processed.
```
ericbottard: The whole "checkpoint" thing in StreamParser is about recording the last good, parseable position at the semantic level (i.e. not just token boundaries).
So for example, we should not checkpoint after just `<a & b` (missing `>`)

spring-xd-dirt/src/main/java/org/springframework/xd/dirt/job/dsl/JobSpecification.java
```diff
@@ -0,0 +1,557 @@
...
+				//		  xsi:schemaLocation="http://www.springframework.org/schema/beans http://www.springframework.org/schema/beans/spring-beans.xsd
+				//		    http://www.springframework.org/schema/batch http://www.springframework.org/schema/batch/spring-batch.xsd">
+				this.doc = domImplementation.createDocument("http://www.springframework.org/schema/beans", "beans",
+						null);
+				doc.createElementNS("http://www.springframework.org/schema/batch", "batch");
+				doc.getDocumentElement().setAttributeNS("http://www.w3.org/2000/xmlns/", "xmlns:batch",
+						"http://www.springframework.org/schema/batch");
+				doc.getDocumentElement().setAttributeNS("http://www.w3.org/2000/xmlns/", "xmlns:xsi",
+						"http://www.w3.org/2001/XMLSchema-instance");
+				doc.getDocumentElement().setAttributeNS("http://www.w3.org/2000/xsi/", "xsi:schemaLocation",
```
cppwfs: Need a schema location for batch as well as beans.

cppwfs:
I did a quick test of a composed job of "aaa || bbb"  and it created 
 ```
 <flow >
  <step id="aaa">
  <tasklet ref="jobRunner-aaa"/>
  <next on="*" to="bbb"/>
  </step>
  <step id="bbb">
  <tasklet ref="jobRunner-bbb"/>
  </step>
  </flow>
  </batch:job>
```
In this case we don't need to add the `<flow></flow>` for this case.

File: spring-xd-dirt/src/main/java/org/springframework/xd/dirt/job/dsl/JobSpecification.java [hunk 13]
```diff
@@ -0,0 +1,557 @@
...
+
+				// Generate:
+				// <beans xmlns="http://www.springframework.org/schema/beans"
+				//		  xmlns:batch="http://www.springframework.org/schema/batch"
+				//		  xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+				//		  xsi:schemaLocation="http://www.springframework.org/schema/beans http://www.springframework.org/schema/beans/spring-beans.xsd
+				//		    http://www.springframework.org/schema/batch http://www.springframework.org/schema/batch/spring-batch.xsd">
+				this.doc = domImplementation.createDocument("http://www.springframework.org/schema/beans", "beans",
+						null);
+				doc.createElementNS("http://www.springframework.org/schema/batch", "batch");
```
cppwfs: also had to use xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" else I received SAX exception

spring-xd-dirt/src/main/java/org/springframework/xd/dirt/job/dsl/JobSpecification.java
```diff
@@ -0,0 +1,557 @@
...
+		};
+
+		@Override
+		public void postJobSpecWalk() {
+			for (String jobRunnerId : jobRunnerIds) {
+				// Producing:
+				// <bean id="jobRunner-a6e0" class="JobRunningTasklet" scope="step"/>
+				Element bean = doc.createElement("bean");
+				bean.setAttribute("scope", "step");
+				bean.setAttribute("class", "JobRunningTasklet");
```
cppwfs: Just a quick update.  The name now is: org.springframework.xd.dirt.batch.tasklet.JobLaunchingTasklet

cppwfs:
Ever closer :-)  we need to add constructor args to the JobLaunchingTasklet in the XML.   For example:
```
<bean class="org.springframework.xd.dirt.batch.tasklet.JobLaunchingTasklet" id="jobRunner-aaa" scope="step">
        <constructor-arg ref="messageBus"/>
        <constructor-arg ref="jobDefinitionRepository"/>
        <constructor-arg ref="xdJobRepository"/>
        <constructor-arg value="aaa"/>
</bean>
<bean class="org.springframework.xd.dirt.batch.tasklet.JobLaunchingTasklet" id="jobRunner-bbb" scope="step">
        <constructor-arg ref="messageBus"/>
        <constructor-arg ref="jobDefinitionRepository"/>
        <constructor-arg ref="xdJobRepository"/>
        <constructor-arg value="bbb"/>
</bean>
```

cppwfs:
LGTM

cppwfs:
Rebased Merged

PR #1813, opened 2015-10-21
[Sprint: 60] XD-2882 Adding options to provide 'enableSync' and 'syncTimeout' values

trisberg:
- using '--enableSync=true' and '--syncTimeout=10000' will flush the data written every 10 seconds and make it available for reads from the datanode. Using these options will have an impact on throughput.


jvalkeal:
looks good to me, I'll test and merge.

jvalkeal:
merged, closing.

PR #1815, opened 2015-10-22
[Sprint: 60] JobDSL changes, including support for XD-3636

aclement:
Changes covered here:
- test cases added to cover some cases Glenn was uncovering in his investigations
- Turned *off* support for inline job definitions in the DSL. It is actually configurable in the parser but turned off for now. Adjusted all the tests to allow for this.  Also ensure appropriate error message for Glenns case "<aaa & bbb & ccc> || <ddd & eee fff>" will get an error on the fff now.
- Parser allows options on job references: "aaa --timeout=1000 || bar --timeout=100" (XD-3636)
- Parser ensures job reference options manifest in XML, multiple new test cases for this, hope the syntax is correct (they are passed as constructor-args)


spring-xd-dirt/src/main/java/org/springframework/xd/dirt/job/dsl/JobDSLMessage.java
```diff
@@ -71,8 +71,13 @@
 															"Only a single '&' is required between jobs in a split"), //
 															EXPECTED_JOB_REF_OR_DEF(ERROR,
 																	219,
-																	"Expected job reference or definition");
-	;
+																	"Expected job reference or definition"), //
```
cppwfs: More Eclipse Formatting fun ;-)

aclement:
Updates based on this mornings discussion:
- now only allows options at the end of the DSL string. These are attached to the parsed JobSpecification object rather than any individual step
- for DSL where the same job is referenced multiple times, only one task let should be created in the XML and it is reused by the different steps that need it. (The ids for the steps are different)

cppwfs:
rebased squashed and merged

PR #1818, opened 2015-10-23
Added support for the timeout module option

cppwfs:
Fixed Parser bug where a||b would not be recognized as a orchestrated job.

No longer supports the pollInterval


spring-xd-dirt/src/main/java/org/springframework/xd/dirt/module/support/ModuleDefinitionService.java
```diff
@@ -143,7 +145,17 @@ private void writeXML(JarOutputStream target, String xml) {
 			throw new IllegalStateException(ioe.getMessage(), ioe);
 		}
 	}
-
+	private void writeParameters(JarOutputStream target, String parameters) {
+		try (InputStream in = new ByteArrayInputStream(parameters.getBytes(Charset.forName("UTF-8")))) {
```
mminella: I think you can use 'StandardCharsets.UTF_8` instead of a string literal here.

mminella:
Merged as 3be80fd9f75d9a25c77bb026be965a8ae20ea84b.

PR #1821, opened 2015-10-25
[Sprint: 60] Updated JobLaunchingTasklet to support restarts of child jobs

mminella:


cppwfs:
This change did not give me the expected behavior: 
1. create the following stream: `aaa || || <bbb & ccc> ||failmenow ||ddd
   * where `aaa-ddd` were created using the batch simple job that returned COMPLETED
   * where `failmenow` was created using a batch simple job that returned FAILED
2. the state of the job after execution showed that `failemenow` FAILED
   * however the state of the orchestrated job was COMPLETED
   * Expected the orchestrated job to be marked 'FAILED'

mminella:
Does your child job fail or just return "FAILED"?  Just returning "FAILED"
is different than actually failing.

cppwfs:
Rebased.  Fixed some imports.  Merged

PR #1826, opened 2015-10-26
[Sprint: 60] Modifying parsing, graph and XML generation to reflect current thinking

aclement:
Includes the changes to reflect the 'current thinking' around DSL > XML.  Also updated graph generation to match and graph to DSL text conversion. Includes support for $END and $FAIL transition targets.

### DIFF
--- a/spring-xd-dirt/src/main/java/org/springframework/xd/dirt/job/dsl/Graph.java
+++ b/spring-xd-dirt/src/main/java/org/springframework/xd/dirt/job/dsl/Graph.java
@@ -150,22 +150,6 @@ public class Graph {
 			}
 		}
 		return candidate;
-		//		if (unvisitedLinks.size() == 0) {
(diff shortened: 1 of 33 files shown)


spring-xd-dirt/src/main/java/org/springframework/xd/dirt/job/dsl/Graph.java
```diff
@@ -263,7 +321,8 @@ private void printNode(StringBuilder graphText, Node node) {
 			}
 		}
 		else {
-			graphText.append(node.name);
+			String nameInDSL = node.name;
+			graphText.append(nameInDSL);
```
cppwfs: can use node.name in the append.  Will fix on merge.

cppwfs:
LGTM

cppwfs:
Rebased Squashed Merged.

PR #1829, opened 2015-10-27
[Sprint: 60] XD-3648 Job Executions without Deployed Job (deleted) shall not be restartable

ghillert:


cppwfs:
At this time I can't reproduce this bug.  
I have a job that throws an exception and is restartable.  The restart button is enabled.
However once I delete it and refresh the job execution page the restart button is disabled

PR #1831, opened 2015-10-28
[Sprint: 60] XD-3234: Remove XML REST support

ericbottard:
As discussed with @ilayaperumalg this does not really get rid of XML support at 100%.
Someone requesting rendering in xml will still trigger it.
The problem is, getting rid of `HttpMessageConverters` is pretty tricky, mostly because of boot auto-configurations adding them.


markpollack:
rebased, merged as b15426c3985daed64b7f104d6675ab918dca176e

PR #1832, opened 2015-10-28
[Sprint: 60] Add overwrite flag to HE module

dturanski:
See http://stackoverflow.com/questions/33387778/springxd-receive-mqtt-and-publish-to-mqtt-on-different-topic/33391384#33391384.  Note the flag applies to all headers. Native SI allows this to be set at the individual header level.

### DIFF
--- a/extensions/spring-xd-extension-header-enricher/src/test/java/org/springframework/xd/header/enricher/HeaderEnricherIntegrationTests.java
+++ b/extensions/spring-xd-extension-header-enricher/src/test/java/org/springframework/xd/header/enricher/HeaderEnricherIntegrationTests.java
@@ -18,11 +18,16 @@ package org.springframework.xd.header.enricher;
 import static org.junit.Assert.assertEquals;
 import static org.springframework.xd.dirt.test.process.SingleNodeProcessingChainSupport.chain;
 
+import java.util.Collections;
+import java.util.Map;
+
 import org.junit.AfterClass;
 import org.junit.BeforeClass;
 import org.junit.Test;
 
 import org.springframework.messaging.Message;
+import org.springframework.messaging.MessageHeaders;
+import org.springframework.messaging.support.MessageBuilder;
 import org.springframework.xd.dirt.server.singlenode.SingleNodeApplication;
 import org.springframework.xd.dirt.test.process.SingleNodeProcessingChain;
 import org.springframework.xd.test.RandomConfigurationSupport;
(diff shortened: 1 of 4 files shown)


modules/processor/header-enricher/config/header-enricher.groovy
```diff
@@ -26,7 +27,7 @@ beans {
 	
 	si.'header-enricher'('input-channel':'input','output-channel':'output') {
 		result.each {k,v->
-			si.'header'(name:k,expression:v)
+			si.'header'(name:k,expression:v,overwrite:overwrite)
```
artembilan: Having `overwrite` as a common option, I'd suggests to do like this:
```
si.'header-enricher'('input-channel':'input','output-channel':'output', 'default-overwrite':overwrite) 
```
Instead of overheading on each `header`

garyrussell: Of course, the next request we'll get is how to overwrite some but not others - but this is an important good start; they can chain two HEs as a work-around.

extensions/spring-xd-extension-header-enricher/src/test/java/org/springframework/xd/header/enricher/HeaderEnricherIntegrationTests.java
```diff
@@ -78,6 +83,25 @@ public void testSimpleExpression() {
 	}
 
 	@Test
+	public void testOverwriteHeaders() {
+		String streamName = "testOverwrite";
+
+		String headers = "{\"bar\":\"payload.substring(1)\"}";
+
+		String processingChainUnderTest = String.format("%s --headers=%s --overwrite=true", moduleName, headers);
```
garyrussell: Not sure what's going on but if I change this to `false`, this test fails as expected, but so does `testSimpleExpression` with an NPE at line 77.

garyrussell: Ah - crosstalk from the failed test.

garyrussell:
Polishing here: https://github.com/garyrussell/spring-xd/commit/2d608f84517234f5bbb853182e57f73bffad66f7

@artembilan please review and I'll merge

artembilan:
Yes, I'm fine with the polishing.
Please, go ahead and merge.

@dturanski , thank you for the contribution, BTW ! :smile:

garyrussell:
Merged as 374c4ef4d40095bb123cf96f61ad9ea9a97f24b0

PR #1838, opened 2015-10-30
XD-3655 Document admin-ui improvements

ghillert:
Jira: https://jira.spring.io/browse/XD-3655


cppwfs:
The primary change will be that we now use the term Orchestrated job vs. Composed Job.  For the docs and examples lets change the wording to Orchestrated jobs.  We can update the screen captures when we go GA.

cppwfs:
Based on a previous discussion, it LGTM

PR #1841, opened 2015-11-03
[Sprint: 61] XD-3683 Supports single job composition

cppwfs:
If a user decides to create a single job composition they can do so but it requires them to use transitions
i.e. job create --definition "aaa |FOO=ccc |'*'=$END"

If a user creates a job composition with job instance without transitions the following exception will be thrown.
Command failed org.springframework.xd.rest.client.impl.SpringXDException: Job Instance specified  aaa must use transitions.  Or just launch the job instance directly

If a user selects a job that is neither an instance nor a module, they will get the traditional message
Could not find module with name 'aaa' and type 'job

### DIFF
--- a/spring-xd-dirt/src/main/java/org/springframework/xd/dirt/job/dsl/ComposedJobUtil.java
+++ b/spring-xd-dirt/src/main/java/org/springframework/xd/dirt/job/dsl/ComposedJobUtil.java
@@ -59,11 +59,11 @@ public class ComposedJobUtil {
 			//Check for single instanceof job with a transition
 			if (jobSpecification.getJobReferences().size() == 1){
 				JobReference reference = jobSpecification.getJobReferences().get(0);
-				if( reference.transitions.size() > 0){
+				if( reference.transitions != null && reference.transitions.size() > 0){
 					result = true;
 				}
 			}
-		} catch (Exception e) {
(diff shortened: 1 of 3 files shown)


spring-xd-dirt/src/main/java/org/springframework/xd/dirt/job/dsl/ComposedJobUtil.java
```diff
@@ -28,25 +29,67 @@
...
+	 * @param definition the definition to be validated
+	 * @param jobInstances collection of available job instances
+	 */
+	public static void validateNotSingleJobInstance(String definition, Collection<String> jobInstances) {
+		JobSpecification jobSpecification = null;
+		try {
+			JobParser parser = new JobParser();
+			jobSpecification = parser.parse(definition);
+		} catch (Exception e) {
+			return; //failed to parse.
```
mminella: We're just swallowing this?

cppwfs: Gotcha:  I'll catch the following exceptions instead:
JobSpecificationException|CheckpointedJobDefinitionException

mminella:
Merged

PR #1843, opened 2015-11-04
XD-3687 Config docs added for Composed Jobs

cppwfs:
Added configuration section to the composed jobs docs.  These are changes required if the user wants to run jobs in parallel.

### DIFF
--- a/src/docs/asciidoc/Composed-Jobs.asciidoc
+++ b/src/docs/asciidoc/Composed-Jobs.asciidoc
@@ -121,3 +121,41 @@ job create testjob --definition "aaa || bbb --timeout=5000"
 ----
 The timeout parameter affects all the jobs within the job composition.  All parameters
 for aaa and bbb must have been previously specified in their own job definitions.
+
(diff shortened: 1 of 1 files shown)


src/docs/asciidoc/Composed-Jobs.asciidoc
```diff
@@ -121,3 +121,41 @@ job create testjob --definition "aaa || bbb --timeout=5000"
...
+----
+spring:
+  batch:
+# Configure other Spring Batch repository values.  Most are typically not needed
+    isolationLevel: ISOLATION_READ_COMMITTED
+----
+
+==== Updating the default hsqldb url
+If you are using hsqldb for development purposes it is recommended that you update the
+url from the current default by adding `;sql.enforce_strict_size=true;hsqldb.tx=mvcc` to
```
mminella: Why are we not just making those parameters the default?  We do in batch...

cppwfs:
Will change the defaults instead of adding docs.

PR #1854, opened 2015-11-12
XD-3703 : Add SSL and attachments to mail sink

fmarchand:

This PR takes in consideration XD-2076, XD-2498 & XD-3076 but needs discussion (test coverage is not perfect and I have some doubts on where to put packages and classes).


garyrussell:
@fmarchand 

First of all; thanks for the contribution - our sincere apologies for it taking so long to look at it; I have some initial observations:

1. The spring standard is to use leading tabs, not spaces
2. Curly braces are always required:

        if(attachmentFilename.contains(";"))
            attachmentFilenames = StringUtils.split(attachmentFilename, ";");
        else attachmentFilenames = new String[] { attachmentFilename };

 should be

        if(attachmentFilename.contains(";")) {
            attachmentFilenames = StringUtils.split(attachmentFilename, ";");
        }
        else {
    	    attachmentFilenames = new String[] { attachmentFilename };
        }

3. With `try {...} catch {...}` the `catch` should start on a new line.

4. MailTransformer: Using a `<transformer/>` in a sink that routes to `nullChannel` is not appropriate. In fact, this "transformer" transforms nothing; it returns the message unchanged. If we were to use it this way, the class should be called something like `MailSender`; the method should return `void` and be invoked using a `<service-activator/>`. However, I think it __should__ be a `Transformer` but return the assembled `MimeMessage` which we then route to the standard Spring Integration `MailSendingMesageHandler` - `<int-mail:outbound-channel-adapter/>` (which you removed).

5. I like the idea of supporting multiple attachments (the standard Spring Integration `MailSendingMessageHandler` only supports one). However, I don't know how useful it is to only support getting the attachments from the file system in an XD environment. We should support other mechanisms for the attachments (such as a `Map` payload).

A fully featured transformer like this would certainly be a candidate to be promoted to Spring Integration itself (or at least add the logic directly into the conversion logic within the adapter).

That's all for this pass - thanks again for the contribution.

fmarchand:
@garyrussell 

That's fine :) I guess you have a lot of work with spring-cloud-data-flow.

First : thanks for your review !

I'm gonna fix what you mentioned : remark 1 to 3. I changed my IDE and I guess I forgot to use the right formater and code cleaner.

For the point 4. You're right : a transformer that routes to nullChannel is a bit weird. I'll fix it too if you want, with a transformer that will return a MimeMessage to a `<int-mail:outbound-channel-adapter/>`

For the point 5. I need to think a bit more about it :)

For the last thing you said, when this PR will be finished, I can propose a pull request to Spring Integration base on this contribution.

Thx again Gary for your review.

PR #1857, opened 2015-11-12
[Sprint: 61] XD-3673 Propagate sequence and count to modules with named channel input

mbogoevici:
* Allow modules with a named channel input to receive index and count, so that zero-indexed
modules that receive data off a tap or named channel can act as competing consumers in Kafka
* Propagate group name to consumers for establishing consumer groups for offset management with Kafka
* Ensure that there is no automatic repartitioning on shared destinations (topics and named channels) so
that modules do not interfere with each other

### DIFF
--- a/config/servers.yml
+++ b/config/servers.yml
@@ -95,6 +95,7 @@
 #        maxWait:                   100
 #        fetchSize:                 1048576
 #        minPartitionCount:         1
+#        durableSubscription:       false
 
 #Disable batch database initialization
 #spring:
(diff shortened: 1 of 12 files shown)


mbogoevici:
In retrospect, this could use a couple of tests.

garyrussell:
We should get rid of `MessageBus.Capability` and change the rabbit bus to use the same technique to access the group.

I can do that and you can add it to this PR if you wish.

mbogoevici:
Oh, sure, that would be sweet.

garyrussell:
Just to close out the comments; as discussed with @mbogoevici on skype - we've abandoned this unification attempt - too much to unravel at this late stage. It is addressed in the s-c-s Binders.

mbogoevici:
Pushed update to Kafka implementation, so that it is consistent with Rabbit.

garyrussell:
Can you rebase and fix up the conflicts?

mbogoevici:
Yes, I was in the process of doing that :)

garyrussell:
Rebased, merged as e99f21c33fe7770c33346846c192d20c726b6ccb after minor polish (mainly imports).

PR #1866, opened 2015-11-30
XD-3709: Unique Domain Name for Module MBeans

garyrussell:
JIRA: https://jira.spring.io/browse/XD-3709

Spring Integration has a (regression) bug that causes router MBeans to be
registered by the context `MBeanExporter`. Usually, Spring Integration components
are registered by the `IntegrationMBeanExporter`, if present.

This problem means that if more than one `router` sink is deployed with JMX enabled, deployment
fails with `InstanceAlreadyExistsException`.

This has been corrected in Spring Integration and will be available in 4.2.3.

However, the same issue would apply to user MBeans in custom modules.

The `mbean-exporters` XML currently adds a unique domain name to the integration exporter, but
leaves the context exporter with the default domain.

Tested with `time | router --expression='''queue:foo'''` (two streams) before and after.


dependencies.properties
```diff
@@ -282,27 +282,27 @@ org.slf4j\:slf4j-api=1.7.12
...
-org.springframework\:spring-messaging=4.2.2.RELEASE
-org.springframework\:spring-orm=4.2.2.RELEASE
-org.springframework\:spring-oxm=4.2.2.RELEASE
-org.springframework\:spring-test=4.2.2.RELEASE
-org.springframework\:spring-tx=4.2.2.RELEASE
-org.springframework\:spring-web=4.2.2.RELEASE
-org.springframework\:spring-webmvc=4.2.2.RELEASE
-org.springframework\:spring-webmvc-portlet=4.2.2.RELEASE
-org.springframework\:spring-websocket=4.2.2.RELEASE
+org.springframework\:spring-messaging=4.2.3.RELEASE
```
ilayaperumalg: Is there a requirement to upgrade spring version to `4.2.3` or are we just updating to the latest release?

ilayaperumalg: Nevermind, I see the reason as mentioned in the commit description.

ilayaperumalg:
LGTM.

PR #1876, opened 2016-01-15
XD-3734: Rabbit: Fix Partitions with AutoDLQ

garyrussell:
JIRA: https://jira.spring.io/browse/XD-3734

Partitioned streams have a single DLQ for each pipe, the originating partition
is available in the `partition` message header.

There were two issues with the implementation:

1. The DLQ was bound to the DLX with the base queue name (no partitions) so rejected
messages were not routed to the DLQ.
2. On the producer side, the queue was bound with different (no) dlq arguments causing
the declaration to fail; the bus should call `declareQueueIfNotPresent`.

Add a test to reproduce the issues and fix the bus.


mbogoevici:
Merged w/https://github.com/spring-projects/spring-xd/commit/52a7c23cbd9cb75962c3f4df3bf1ca3f7af7f626

PR #1880, opened 2016-02-01
XD-3716: Rabbit Bus - Configurable longStringLimit

garyrussell:
JIRA: https://jira.spring.io/browse/XD-3716

The `longStringLimit` property determines when long
headers are converted to `String` Vs. `DataInputStream`.

The latter is not (currently) property converted on output.

https://jira.spring.io/browse/AMQP-568


markpollack:
merged f6be8387e31783d4b619240978fc6fdaeec64139

PR #1882, opened 2016-02-01
XD-3736: Rabbit Bus - Support Pub/Sub Concurrency

garyrussell:
JIRA: https://jira.spring.io/browse/XD-3736

There is no reason to not support concurrency since the consumer threads
compete for messages from the bound queue.


markpollack:
merged 0c4297b17cd94806d2ae167332c26093edae17fe

PR #1886, opened 2016-02-05
XD-3739: Fix and Configure Script Refresh

garyrussell:
JIRA: https://jira.spring.io/browse/XD-3739

Scripts were incorrectly checked for changes with a hard-wired 60ms interval
instead of 60 seconds as stated in the reference manual.

Change the default to 60 seconds and make it configurable; `-1` disables refresh.


markpollack:
merged as a42ccbcc3c233bc5c784262958e5e0ed001f247f

PR #1887, opened 2016-02-10
Make the `MessageConverterTests` Win compatible

artembilan:
https://build.spring.io/browse/XD-WIN-JOB1-353

We should change constant in the the code
`ØØØØØØØØ`

into its Unicode representation:

`\u00d8\u00d8\u00d8\u00d8\u00d8\u00d8\u00d8\u00d8`

With that we have the same `byte[]`.

The thought is Java uses `file.encoding` to read similar compiled constant from its classes in the classpath.


artembilan:
Merged as https://github.com/spring-projects/spring-xd/commit/e4b163e942911b54947bd82d32463fc08bf0bc7e by @markpollack

PR #1889, opened 2016-02-15
XD-3470 Add maxWait property to set in kafka message bus

ilayaperumalg:
- Add `maxWait` property in kafka message bus config setter


garyrussell:
Rebased, merged as d48dc793d99195508e6a63dbc4ae4abaeea52354

PR #1893, opened 2016-02-17
Xd 3742 - Enable in line SSL properties as an alternative to external properties files.

dturanski:
This adds explicit in line properties for `keyStore`, `keyStorePassphrase`, `trustStore`, and `trustStorePassphrase`. These can be set enable storing properties in encrypted form enabled in https://jira.spring.io/browse/XD-3738.

This affects SSL settings for RabbitMessageBus,  Rabbit source, Rabbit sink, and Http source. 

Manual testing performed:

 - Regression test each of these using the current mechanisms: Enable SSL, use rabbit SSL address, external properties files with valid and invalid values in servers.yml, modules.yml, and module config properties.
- Test Rabbit MB with in line configuration in servers.yml
- Test Rabbit modules inherit the default MB configuration
- Test Streams with rabbit sink and source modules, configured with inline properties
- Regression test http source with SSL configured from new `config/modules/source/http/http.properties` 
`http --https=true | log` and `curl -X post --cacert path/to/cecert https://localhost:9000 -d hello`

### DIFF
--- a/modules/source/rabbit/config/rabbit.xml
+++ b/modules/source/rabbit/config/rabbit.xml
@@ -44,13 +44,19 @@
 		virtual-host="${vhost}"
 		username="${username}" password="${password}" />
 
+	<rabbit:connection-factory id="rabbitConnectionFactory"
+							   connection-factory="connectionFactory"
+							   addresses="${addresses}"
+							   virtual-host="${vhost}"
+							   username="${username}" password="${password}" />
+
 	<bean id="connectionFactory" class="org.springframework.amqp.rabbit.connection.RabbitConnectionFactoryBean">
 		<property name="useSSL" value="${useSSL}" />
-		<property name="sslPropertiesLocation" value="${sslPropertiesLocation:}" />
-		<property name="keyStore" value="${keyStore:}" />
-		<property name="keyStorePassphrase" value="${keyStorePassphrase:}" />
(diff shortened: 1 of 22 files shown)


modules/source/rabbit/config/rabbit.xml
```diff
@@ -44,9 +44,19 @@
 		virtual-host="${vhost}"
 		username="${username}" password="${password}" />
 
+	<rabbit:connection-factory id="rabbitConnectionFactory"
+							   connection-factory="connectionFactory"
+							   addresses="${addresses}"
+							   virtual-host="${vhost}"
+							   username="${username}" password="${password}" />
+
```
garyrussell: Duplicate bean ?

garyrussell:
Otherwise, LGTM; I can remove that on merge if you agree.

dturanski:
good catch

sabbyanandan:
@dturanski: Thoughts on Gary's comment?

dturanski:
"good catch"  ^^^ wasn't clear enough I guess. Yes we should fix it on the merge.

garyrussell:
Yup - merging now

garyrussell:
Rebased, squashed, merged as e63e6c545f188cc388dfe2433003596be7c53a19 with a bit o' spit and polish.

PR #1898, opened 2016-02-23
Adding a close for the JarOutputStream

cppwfs:
Was removed during a refactoring effort.  Caused the Composed jar to not be properly opened during deployment.
Also removed the JarOutputStream from the try parameters.


mminella:
Merged

PR #1900, opened 2016-02-29
XD-3748: Fix ModuleObjectNamingStrategy

garyrussell:
JIRA: https://jira.spring.io/browse/XD-3748

Previously, the `ModuleObjectNamingStrategy` didn't handle beans with
a `beanKey` that's a simple bean name, such as the `MessageHistoryConfigurer`.

Subclass `MetadataNamingStrategy` which handles simple bean names.

Add tests.

### DIFF
--- a/spring-xd-dirt/src/main/resources/META-INF/spring-xd/plugins/jmx/mbean-exporters.xml
+++ b/spring-xd-dirt/src/main/resources/META-INF/spring-xd/plugins/jmx/mbean-exporters.xml
@@ -22,8 +22,11 @@
 
 	<bean id="moduleObjectNamingStrategy"
 		class="org.springframework.xd.dirt.module.jmx.ModuleObjectNamingStrategy">
+		<constructor-arg ref="attributeSource" />
 		<constructor-arg name="domain" value="xd.${xd.stream.name:${xd.job.name:}}" />
 		<constructor-arg name="objectNameProperties" ref="objectNameProperties" />
 	</bean>
+
+	<bean id="attributeSource" class="org.springframework.jmx.export.annotation.AnnotationJmxAttributeSource" />
 
 </beans>
--- a/spring-xd-dirt/src/main/java/org/springframework/xd/dirt/module/jmx/ModuleObjectNamingStrategy.java
+++ b/spring-xd-dirt/src/main/java/org/springframework/xd/dirt/module/jmx/ModuleObjectNamingStrategy.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2013-2015 the original author or authors.
+ * Copyright 2013-2016 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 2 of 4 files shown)


spring-xd-dirt/src/main/resources/META-INF/spring-xd/plugins/jmx/mbean-exporters.xml
```diff
@@ -22,8 +22,11 @@
 
 	<bean id="moduleObjectNamingStrategy"
 		class="org.springframework.xd.dirt.module.jmx.ModuleObjectNamingStrategy">
+		<constructor-arg ref="attributeSource" />
 		<constructor-arg name="domain" value="xd.${xd.stream.name:${xd.job.name:}}" />
 		<constructor-arg name="objectNameProperties" ref="objectNameProperties" />
 	</bean>
 
+	<bean id="attributeSource" class="org.springframework.jmx.export.annotation.AnnotationJmxAttributeSource" />
```
artembilan: Looks like you use the same object in the other constructor.
What is the point of this bean?
To be able to utilize its `embeddedValueResolver` through the `BeanFactoryAware` ?
Any sample on the matter then?

Thanks

garyrussell: In the other constructor you get a basic one with no bean factory; with this one we get a bean factory so we can resolve pph in the IMR.

I added this here so we get a value resolver in case (in future) some Mbean might have a PPH in an annotated object name but, I agree, I don't think it matters now - but I didn't think it would hurt to initialize the bean properly - in case something else changes in future.

artembilan: :+1: 
Perfect answer!
Nothing to complain about :smile: 

I can merge if there is no objections from other guys.

garyrussell: BTW; I think we should make `IMBE.IntegrationJmxAttributeSource` public so it can be reused for situations like this.

artembilan: No reason. See its source code and usage from the outer `IMBE` class.
We don't delegate to the super `AnnotationJmxAttributeSource` because we fully override `getManagedResource()`, so that extra `BeanFactoryAware` is redundant.

garyrussell: I mean if someone creates a custom component, they might want to mark it with IntegrationManagedResource and export it with the IMBE instead of the context MBE - so it's grouped with related MBeans.

Right now, IMR is only good for "internal" components.

artembilan: I think I got it.
Thanks!

https://jira.spring.io/browse/INT-3959

Merging BTW...

artembilan:
Sorry, I can't do that:
```
remote: Permission to spring-projects/spring-xd.git denied to artembilan.
fatal: unable to access 'https://github.com/spring-projects/spring-xd.git/': The requested URL returned error: 403
```
:smile:

artembilan:
Merged as https://github.com/spring-projects/spring-xd/commit/115487070e9196a4bb8bfd9b6fe8b85494ac6cca after rebase.

PR #1901, opened 2016-03-03
XD-3751 Fix gpfdist processor shutdown

jvalkeal:
- Adding a workaround for a problem in reactor 2.0.x where
  onNext and onComplete will result deadlock if ringbuffer
  is full.
- We now try to let gpdb load session to drain stream and detect
  if that succeed by checking buffer size and what's a remaining
  capacity. If we can't drain, last possible option is to
  force processor shutdown.

### DIFF
--- a/extensions/spring-xd-extension-gpfdist/src/main/java/org/springframework/xd/greenplum/gpfdist/GPFDistMessageHandler.java
+++ b/extensions/spring-xd-extension-gpfdist/src/main/java/org/springframework/xd/greenplum/gpfdist/GPFDistMessageHandler.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2015 the original author or authors.
+ * Copyright 2015-2016 the original author or authors.
  *
  * Licensed under the Apache License, Version 2.0 (the "License");
  * you may not use this file except in compliance with the License.
@@ -13,6 +13,7 @@
  * See the License for the specific language governing permissions and
  * limitations under the License.
  */
+
 package org.springframework.xd.greenplum.gpfdist;
 
 import java.util.Date;
@@ -21,7 +22,7 @@ import java.util.concurrent.TimeUnit;
 
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
-import org.reactivestreams.Processor;
+
 import org.springframework.messaging.Message;
 import org.springframework.messaging.MessageHandlingException;
 import org.springframework.scheduling.TaskScheduler;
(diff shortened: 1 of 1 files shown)


extensions/spring-xd-extension-gpfdist/src/main/java/org/springframework/xd/greenplum/gpfdist/GPFDistMessageHandler.java
```diff
@@ -127,49 +134,93 @@ protected void doStart() {
...
 			log.info("Skipping gpload tasks because greenplumLoad is not set");
 		}
 	}
 
 	@Override
 	protected void doStop() {
+		boolean drained = false;
 		if (greenplumLoad != null) {
+
+			// xd waits 30s to shutdown module, so lets wait 25 to drain
```
markpollack: Can you explain this a bit more - sounds like some sort of race condition, shouldn't we let the entire buffer drain since the messages in the buffer have been ack'd (say in rabbit) ?

extensions/spring-xd-extension-gpfdist/src/main/java/org/springframework/xd/greenplum/gpfdist/GPFDistMessageHandler.java
```diff
@@ -127,49 +134,93 @@ protected void doStart() {
...
 				long now = System.currentTimeMillis();
 				// wait a bit more than batch period
+				log.info("Cancelling loading task");
 				Boolean value = taskFuture.get(batchTimeout + batchPeriod + 2, TimeUnit.SECONDS);
 				log.info("Stopping, got future value " + value + " from task which took "
 						+ (System.currentTimeMillis() - now) + "ms");
-			} catch (Exception e) {
+			}
+			catch (Exception e) {
 				log.warn("Got error from task wait value which may indicate trouble", e);
```
markpollack: just seeing this now, but I'd probably rip my hair out if I had to read this error message in production.  What is the trouble?

extensions/spring-xd-extension-gpfdist/src/main/java/org/springframework/xd/greenplum/gpfdist/GPFDistMessageHandler.java
jvalkeal: This deadlock within a reactor is something which exists in 2.0.x, although it fixed in 2.5. Effectively when trying to shutdown a processor, signal is sent into a downstream indicating its complete but if there's existing messages in a ringbuffer, that terminate signal never reach a correct component in a reactor because we already stopped draining. Module shutdown timeout is afaik hardcoded to 30 secs in XD and after that things go a bit haywire if module is not actually properly closed.

This was a workaround I came out with discussion with stephane. It rely on a fact that we try to keep the load operations running little less time when XD would throw errors that it's unable to shutdown a module. We're hoping that these load operations will eventually drain the buffers and allows terminate signal to go down stream, thus allowing processor to shutdown and thus allow clean shutdown of a module.

extensions/spring-xd-extension-gpfdist/src/main/java/org/springframework/xd/greenplum/gpfdist/GPFDistMessageHandler.java
jvalkeal: I just wanted to add a logging here instead of just ignoring if we get an exception from a `TaskScheduler` when cancelling `Future`. I'm not entirely sure when that exception would be thrown but wanted to add it to logs in case shutdown somehow still fails and we're getting exception here.

PR #1907, opened 2016-04-05
Gpdb header support

caxqueiroz:
This PR adds support to the HEADER tag for GPDB so that it can skip the first row when data is being loaded.


jvalkeal:
Right at least you should convert app leading spaces to tabs.

Did you ever figure out if we can even use this header support as it looks like header needs to be skipped on a sender side in our component? Which makes its use quite awkward in a stream.

PR #1911, opened 2016-05-17
Add MongoDbSourceModuleTests

artembilan:


artembilan:
Out of topic: I spent some time (at about an hour?) to figure out that I just should add this `hadoop.home.dir=d:\Hadoop` into the `gradle.properties` to make it working on Windows...
Even just from the project import into the IDEA perspective.
It would be better to make:
```
throw new RuntimeException("ERROR-'hadoop.home.dir' is not set. Use -Phadoop.home.dir=[].\n see https://github.com/spring-projects/spring-hadoop/wiki/Using-a-Windows-client-together-with-a-Linux-cluster")
```
more useful in context of XD. All system and ENV options aren't relevant for XD.

Thank you for attention :smile:

